PR #123, opened 2018-04-04
ADEN-6957 Implement slot viewport conflicts logic

fraszczakszymon:


src/services/btf-blocker-service.js
```diff
@@ -4,19 +4,23 @@ import { slotService } from './slot-service';
 
 const logGroup = 'btf-blocker';
 
+function disableBtf() {
+	const slots = context.get('slots');
+
+	Object.keys(slots).forEach((adSlotKey) => {
+		const adSlot = slots[adSlotKey];
+
+		if (!adSlot.aboveTheFold && this.unblockedSlots.indexOf(adSlot.getSlotName()) === -1) {
```
v3rth: Did you mean `!this.unblockedSlots.includes(adSlot.getSlotName())`? :D

src/services/slot-service.js
```diff
@@ -37,22 +40,48 @@ class SlotService {
...
+			if (distance < viewportHeight) {
+				hasConflict = true;
+			}
+
+			logger(groupName, 'hasViewportConflict', hasConflict, adSlot.getSlotName(), elementId);
+
+			return !hasConflict;
+		});
+
+		return hasConflict;
```
v3rth: I think [`every`](https://github.com/Wikia/ad-engine/pull/123/files#diff-7b5430adb62b98635a56208161791dc5R53) is able to return the result so you can remove this line and maybe consider replacing `every` to `some`

src/services/slot-service.js
```diff
@@ -37,22 +40,48 @@ class SlotService {
...
+		setState(slotName, false, status);
+	}
+
+	hasViewportConflict(adSlot) {
+		const slotHeight = adSlot.getElement().offsetHeight,
+			slotOffset = getTopOffset(adSlot.getElement()),
+			viewportHeight = window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
+
+		let hasConflict = false;
+		adSlot.getViewportConflicts().every((elementId) => {
```
v3rth: I think this function is big and separated enough to be extracted

src/services/btf-blocker-service.js
fraszczakszymon: This code was extracted only. You're right but it isn't supported in IE11 :(

src/services/slot-service.js
fraszczakszymon: Thanks! Addressed

src/services/slot-service.js
fraszczakszymon: Addressed 👍

PR #129, opened 2018-04-12
ADEN-6951 Adds global target

leshniak:


fraszczakszymon:
Could you add uglify?

PR #133, opened 2018-05-02
ADEN-7078 Adds events service

leshniak:


package.json
```diff
@@ -33,6 +32,7 @@
     "eslint-plugin-import": "2.11.0",
     "eslint-plugin-jsx-a11y": "6.0.3",
     "eslint-plugin-react": "7.7.0",
+    "eventemitter3": "3.0.1",
```
dmnsobczak: Shouldn't it be listed under dependencies, not devDependencies?

leshniak: ofc you're right

PR #135, opened 2018-05-09
ADEN-7118 Expose method to finish ATF queue

fraszczakszymon:


src/services/btf-blocker-service.js
```diff
@@ -16,39 +16,47 @@ function disableBtf() {
...
 				}
 			}
 		});
 	}
 
+	finishAboveTheFold() {
+		this.atfEnded = true;
+
+		if (window.ads.runtime.disableBtf) {
+			disableBtf.bind(this)();
```
leshniak: since `.bind` is a bit slower than `.call`, it's better to use the second in case of calling a function immediately.

PR #155, opened 2018-08-07
ADEN-7362 Prepare AE for outstream template

fraszczakszymon:


src/listeners/slot-listener.js
```diff
@@ -88,6 +88,8 @@ class SlotListener {
 			case 'collapse':
 				adSlot.collapse();
 				break;
+			case 'manual':
```
jacekbj: Do we have a list of "ad types"?

src/utils/dimensions.js
```diff
@@ -30,7 +30,7 @@ export function getViewportHeight() {
 	return Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
 }
 
-export function isInViewport(element, topOffset = 0, bottomOffset = 0) {
+export function isInViewport(element, topOffset = 0, bottomOffset = 0, areaThreshold = 0.5) {
```
jacekbj: Please add jsdoc.

src/video/player/porvata/porvata.js
```diff
@@ -344,4 +346,16 @@ export class Porvata {
...
 			});
 	}
+
+	static isVpaid(contentType) {
+		return contentType === 'application/javascript';
+	}
+
+	static isVideoAutoplaySupported() {
+		const isAndroid = client.getOperatingSystem() === 'Android';
+		const browserInfo = client.getBrowser().split(' ');
```
jacekbj: It's funny that 2 last lines of `getBrowser()` method are:
```
browser = matches.join(' ');

return browser;
```

src/video/player/porvata/porvata.js
```diff
@@ -344,4 +346,16 @@ export class Porvata {
...
+		const isCompatibleChrome = browserInfo[0].indexOf('Chrome') !== -1 && parseInt(browserInfo[1], 10) >= 54;
```
jacekbj: https://github.com/video-dev/can-autoplay funny how it tries to insert video/audio element into website and play it :D I couldn't find "source of truth" for supported browsers so I think this is fine.

src/listeners/slot-listener.js
fraszczakszymon: There are a couple of them in the running DFP creatives. But ad-engine isn't going to handle all of them. Here is the main list.

src/video/player/porvata/porvata.js
fraszczakszymon: You're right. It doesn't makes sense. Fixed :)

src/utils/dimensions.js
fraszczakszymon: Added

src/video/player/porvata/porvata.js
fraszczakszymon: We can consider that. But I would run it in separate story/epic. We would need to check what's the overall impact

PR #157, opened 2018-08-08
ADEN-7362 Add flag to prevent starting out of viewport

fraszczakszymon:


src/video/player/porvata/porvata.js
```diff
@@ -285,7 +285,7 @@ export class Porvata {
 				}
 
 				function setupAutoPlayMethod() {
-					if (params.blockOutOfViewportPausing) {
+					if (params.blockOutOfViewportPausing && !params.startInViewportOnly) {
```
fraszczakszymon: This is a new flag, not defined anywhere yet. Feel free to merge after CR

nandy-andy: It will be needed not to make Chrome angry ;) They don't allow autoplay videos which aren't in the viewport.

PR #162, opened 2018-08-20
ADEN-7568 Disable SRA when context "options.isSraDisabled" is truthy. 

jacekbj:


src/providers/gpt-provider.js
```diff
@@ -40,16 +40,21 @@ function configure() {
...
+	}
+
+	isInitialized() {
+		return initialized;
 	}
 
 	@decorate(gptLazyMethod)
 	init() {
-		if (initialized) {
+		console.log(this.isInitialized());
```
nandy-andy: 🙈

PR #166, opened 2018-08-31
ADEN-7613 Expose getState method in slotService

fraszczakszymon:


spec/services/slot-service.spec.js
```diff
@@ -79,4 +79,20 @@ describe('slot-service', () => {
 
 		expect(slotService.hasViewportConflict(adSlot)).to.equals(false);
 	});
+
+	it('slot state is truthy when it was not defined before', () => {
+		expect(slotService.getState('foo')).to.equals(true);
+	});
+
+	it('slot state is truthy when it was enabled defined before', () => {
```
nandy-andy: Does it mean the same as: `slot state is truthy when it was defined before and enabled` or the order doesn't matter?

spec/services/slot-service.spec.js
```diff
@@ -79,4 +79,20 @@ describe('slot-service', () => {
...
+		slotService.enable('foo');
+
+		expect(slotService.getState('foo')).to.equals(true);
+	});
+
+	it('slot state is falsy when it was enabled defined before', () => {
```
nandy-andy: What do you think of changing the description? I believe it's supposed to go similarly to this: `slot state is falsy when it was defined before as disabled`.

src/services/slot-service.js
```diff
@@ -25,6 +25,20 @@ function isSlotInTheSameViewport(slotHeight, slotOffset, viewportHeight, element
 	return distance < viewportHeight;
 }
 
+function setState(slotName, state, status = null) {
+	const slot = slotService.get(slotName);
+	slotStates[slotName] = state;
+	slotStatuses[slotName] = status;
+
+	if (slot) {
```
nandy-andy: What do you think about the same thing but written differently:
```js
if (slot && state) {
	slot.enable();
} else if (slot && !state) {
	slot.disable(status);
} else {
	throw new Error('There is no slot nor state - nothing setState() can do here');
}
```

spec/services/slot-service.spec.js
fraszczakszymon: Fixed description. Copy&paste didn't handle it for me 🙈

spec/services/slot-service.spec.js
fraszczakszymon: Fixed description. Copy&paste didn't handle it for me 🙈

src/services/slot-service.js
fraszczakszymon: It can't work in that way. We have to be able to disable slots before they are really configured (exampe: incontent_player on FV page)

fraszczakszymon: In other words: your code doesn't do the same thing. Slot state can be true or false. But it's fine that slot doesn't exist at the moment of setting state

PR #170, opened 2018-09-07
ADEN-7633 UAP:Twitch

pnather:


src/ad-engine/video/player/twitch/embed/twitch-embed.js
```diff
@@ -0,0 +1,27 @@
+import { scriptLoader } from '../../../../utils';
+
+const twitchLibraryUrl = '//player.twitch.tv/js/embed/v1.js';
+
+function load() {
+	if (window.Twitch) {
+		return new Promise((resolve) => {
```
fraszczakszymon: `return Promise.resolve();`

src/ad-products/templates/uap/big-fancy-ad-twitch.js
```diff
@@ -0,0 +1,130 @@
...
+						offset ? '' : `top ${time}ms ${universalAdPackage.CSS_TIMING_EASE_IN_CUBIC}`
+					);
+					navbarElement.style.top = (offset ? `${offset}px` : '');
+				}
+			}
+		};
+	}
+
+	/**
+		 * Constructor
```
fraszczakszymon: Something is wrong with indents :)

src/ad-products/templates/uap/big-fancy-ad-twitch.js
```diff
@@ -0,0 +1,130 @@
...
+	constructor(adSlot) {
+		this.adSlot = adSlot;
+		this.config = context.get('templates.bfat');
+		this.container = document.getElementById(this.adSlot.getSlotName());
+		this.videoSettings = null;
+		this.theme = null;
+	}
+
+	/**
+		 * Initializes the BFAT unit
```
fraszczakszymon: Something is wrong with indents :)

src/ad-products/templates/uap/big-fancy-ad-twitch.js
```diff
@@ -0,0 +1,130 @@
+import { context, utils } from '@wikia/ad-engine';
+import { universalAdPackage } from './universal-ad-package';
+import { VideoSettings } from './video-settings';
+import { SLIDE_OUT_TIME } from './constants';
+import * as classicTheme from './themes/classic';
+import * as hiviTheme from './themes/hivi';
+
+export class BigFancyAdTwitch {
```
fraszczakszymon: Do we really need this template? It looks like 1:1 copy of BigFancyAdAbove (apart from one line L128: `universalAdPackage.loadTwitchAd(iframe, this.params);`). What do you think about passing new parameter (e.g. `player: 'twitch'` in the creative template, in loadCustomAd call) so it loads the same template but different player.

fraszczakszymon: It doesn't need to be `player` parameter, it may base on `channelName `

fraszczakszymon: After seeing separate creative template for that https://github.com/Wikia/dfp-creative-templates/pull/53/files it makes sense for me to have separate template in code. But what about extending existing BigFancyAdAbove?

src/ad-products/templates/uap/universal-ad-package.js
```diff
@@ -55,6 +55,37 @@ async function loadPorvata(videoSettings, slotContainer, imageContainer) {
 	return video;
 }
 
+async function loadTwitchPlayer(iframe, params) {
+	const { playerPosition } = params,
+		twitchContainer = iframe.contentWindow.document.getElementById(playerPosition),
+		twitchContainerHeight = twitchContainer.clientHeight,
+		adContainer = iframe.contentWindow.document.getElementById('adContainer'),
+		clickMacroContainer = playerPosition === 'left' ?
```
fraszczakszymon: What about renaming `left`/`right` to `player` / `clickArea`. Then we won't need checking if playerPosition is left in multiple places. We can add `.left` / `.right` class (using macro) to `#player` container so it will be responsible for styling

examples/templates/hivi-uap-twitch/script.js
```diff
@@ -1,11 +1,11 @@
...
 
 import customContext from '../../context';
 import '../../styles.scss';
 
 customContext.targeting.artid = '415';
 
-templateService.register(BigFancyAdTwitch);
+templateService.register(BigFancyAdAbove);
 
 new AdEngine(customContext).init();
```
pnather: Switched to BFAA template

src/ad-products/templates/uap/big-fancy-ad-above.js
```diff
@@ -133,6 +133,8 @@ export class BigFancyAdAbove {
 			); // defers for proper rendering
 
 			this.theme.onVideoReady(video);
+		} else {
+			universalAdPackage.loadTwitchAd(iframe, this.params);
 		}
 	}
 }
```
pnather: If params don't have videoTrigger load twitch

src/ad-products/templates/uap/universal-ad-package.js
```diff
@@ -70,7 +68,7 @@ async function loadTwitchPlayer(iframe, params) {
 
 	const player = new TwitchPlayer(twitchContainer, options);
 	twitchContainer.style.width = `${twitchContainerHeight * params.twitchAspectRatio}px`;
-	clickMacroContainer.style.width = `${adContainer.clientWidth - twitchContainer.clientWidth}px`;
+	clickMacroContainer.style.width = `${adContainer - twitchContainer.clientWidth}px`;
 	return player;
 }
```
pnather: Changed left/right to player/clickArea

src/ad-products/templates/uap/big-fancy-ad-above.js
```diff
@@ -133,6 +133,8 @@ export class BigFancyAdAbove {
 			); // defers for proper rendering
 
 			this.theme.onVideoReady(video);
+		} else {
```
fraszczakszymon: It would be safer with `} else if (!!params.channelName) {`

fraszczakszymon: Otherwise it will break UAP static (without video)

PR #175, opened 2018-09-17
ADEN-7664

ThisIsJesus:


tests/common/helpers.js
```diff
@@ -1,19 +1,20 @@
 const newUrlTimeout = 10000;
-const interval = 500;
 const valueToDivideBy = 10;
 const pauseBetweenScrolls = 250;
 
 class Helpers {
 	constructor() {
 		this.newsAndStories = 'http://www.wikia.com/fandom';
+		this.interval = 500;
+		this.isHidden = '.hide';
```
emadry: change to e.g. classHidden, otherwise it sounds like method which should return bool

tests/specs/delay-ad.test.js
```diff
@@ -0,0 +1,110 @@
+import delayAd from '../pages/delay-ad.page';
+import { timeouts } from '../common/timeouts';
+import helpers from '../common/helpers';
+
+const { expect } = require('chai');
+
+describe('It will test delay ad page', () => {
+	beforeEach(() => {
+		browser.url(delayAd.pageLink);
```
emadry: Wait for something to be sure, that page is loaded

tests/specs/empty-response.test.js
```diff
@@ -0,0 +1,34 @@
+import emptyResponse from '../pages/empty-response.page';
+import { timeouts } from '../common/timeouts';
+import helpers from '../common/helpers';
+
+const { expect } = require('chai');
+
+describe('It will test empty response page', () => {
+	beforeEach(() => {
+		browser.url(emptyResponse.pageLink, timeouts.standard);
```
emadry: browser.url doesn't take 2nd parameter.
Add separate wait here

tests/common/helpers.js
ThisIsJesus: Changed to classHidden, as you suggested.

tests/specs/delay-ad.test.js
ThisIsJesus: Done.

tests/specs/empty-response.test.js
ThisIsJesus: Must've accidentally added that timeout here, thanks for pointing this out!

tests/common/timeouts.js
```diff
@@ -1,6 +1,7 @@
 class Timeouts {
 	constructor() {
 		this.standard = 5000;
+		this.extended = 10000; // currently added only for animations ad, as top leaderboard hides after 6 seconds
```
emadry: So it can be in animations page and described as wait for animations

tests/pages/animations-ad.page.js
```diff
@@ -0,0 +1,23 @@
...
+	constructor() {
+		this.pageLink = 'slots/animations/';
+		this.topLeaderboard = '#top_leaderboard';
+		this.topBoxad = '#top_boxad';
+		this.topLeaderboardWidth = 728;
+		this.topLeaderboardHeight = 90;
+		this.topLeaderboardHeightHidden = 0;
+		this.topBoxadWidth = 300;
+		this.topBoxadHeight = 250;
+		this.style = 'style';
```
emadry: Name unclear

tests/pages/delay-ad.page.js
```diff
@@ -0,0 +1,31 @@
+import helpers from '../common/helpers';
+import { timeouts } from '../common/timeouts';
+
+const loadAdTime = 10000;
+
+class DelayAd {
+	constructor() {
+		this.pageLink = 'slots/delay/';
+		this.topLeaderboard = '#top_leaderboard[data-gpt-line-item-id="271491732"]';
```
emadry: Id should be unique so no need for additional attribute in form of data-gpt-line-item-id
Line item can be stored as separate value and checked in test

tests/pages/block-btf-ad.page.js
```diff
@@ -0,0 +1,11 @@
+class BlockBtfAd {
+	constructor() {
+		this.pageLink = 'slots/block-btf/';
+		this.incontentBoxad = '#incontent_boxad';
```
emadry: Create AdSlot page for collection of slot id's to share between pages - applies to every page

tests/common/timeouts.js
ThisIsJesus: Moved to animations page.

tests/pages/animations-ad.page.js
ThisIsJesus: Changed to topLeaderboardStyle to make it clearer

tests/pages/delay-ad.page.js
ThisIsJesus: There is a possibility that they were some leftovers when there was some different logic in some test. Cleaned this up now.

tests/pages/empty-response.page.js
```diff
@@ -0,0 +1,8 @@
+class EmptyResponse {
+	constructor() {
+		this.pageLink = 'slots/empty-response/';
+		this.articleClass = '.main';
```
emadry: You might want to change it to `article.main` so selector will be clear

tests/pages/viewport-conflict-ad.page.js
```diff
@@ -0,0 +1,24 @@
+class ViewportConflictAd {
+	constructor() {
+		this.pageLink = 'slots/viewport-conflicts/';
+		this.hideBoxadButton = '#hideTopBoxad';
+		this.addParagraphButton = '#addParagraph';
+		this.dataSlotResult = 'data-slot-result';
```
emadry: change to `[data-slot-result]`

tests/pages/viewport-conflict-ad.page.js
```diff
@@ -0,0 +1,24 @@
...
+		this.maxNewParagraphs = 10;
+	}
+
+	/**
+	 * Provides the parameter of paragraphs to add to the page.
+	 * @param {number} paragraphs - number of paragraphs to load
```
emadry: Add info, that if it's above 10, then it will be changed to 10

tests/specs/animations-ad.test.js
```diff
@@ -0,0 +1,77 @@
...
+			.to
+			.equal(adSlots.leaderboardWidth, 'Top leaderboard ad width incorrect');
+		expect(topLeaderboardSize.height)
+			.to
+			.equal(adSlots.leaderboardHeight, 'Top leaderboard ad height incorrect');
+		expect(browser.isVisibleWithinViewport(adSlots.topLeaderboard))
+			.to
+			.be
+			.true;
+		expect(topBoxadSize.width)
```
emadry: Make test per slot so it won't depend on success of other slots

tests/specs/animations-ad.test.js
```diff
@@ -0,0 +1,77 @@
...
+
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.newsAndStories);
+		expect(browser.getUrl())
+			.to
+			.equal(helpers.newsAndStories);
+		helpers.closeNewTabs();
+	});
+
+	it('will test top leaderboard disappearing after some time', () => {
```
emadry: Change to number instead of 'some'

tests/specs/block-btf-ad.test.js
```diff
@@ -0,0 +1,49 @@
...
+describe('It will test block btf ad page', () => {
+	beforeEach(() => {
+		browser.url(blockBtfAd.pageLink, timeouts.standard);
+		browser.waitForVisible(adSlots.topLeaderboard, timeouts.standard);
+	});
+
+	it('will test top leaderboard visibility and dimensions', () => {
+		const size = browser.getElementSize(adSlots.topLeaderboard);
+
+		expect(size.width)
```
emadry: Try to use soft assertions so it won't exit after first fail

tests/specs/btf-only-ad.test.js
```diff
@@ -0,0 +1,47 @@
...
+	beforeEach(() => {
+		browser.url(btfOnlyAd.pageLink);
+		browser.waitForVisible(btfOnlyAd.finishQueueButton, timeouts.standard);
+		browser.click(btfOnlyAd.finishQueueButton);
+		helpers.slowScroll(2500);
+		browser.waitForVisible(adSlots.incontentBoxad, timeouts.standard);
+	});
+
+	it('will test the visibility of btf ad after manually finishing the queue', () => {
+		reporter.severity('Critical');
```
emadry: Remove from here and add to readme in chapter Severity and Categories

tests/specs/delay-ad.test.js
```diff
@@ -0,0 +1,110 @@
...
+		browser.waitForVisible(delayAd.loadAdsButton, timeouts.standard);
+	});
+
+	it('will test if ads are not immediately visible', () => {
+		expect(browser.isExisting(delayAd.resultAttribute))
+			.to
+			.be
+			.false;
+		expect(browser.isExisting(delayAd.resultAttribute))
```
emadry: Both expects are checking the same thing. Either remove one or use other element

tests/pages/viewport-conflict-ad.page.js
ThisIsJesus: This variable is used in `getAttribute` - putting it in curly brackets results in `waitUntil` failing due to timeout.

tests/common/adSlots.js
```diff
@@ -0,0 +1,15 @@
+class AdSlots {
+	constructor() {
+		this.topLeaderboard = '#top_leaderboard';
+		this.bottomLeaderboard = '#bottom_leaderboard';
+		this.leaderboardWidth = 728; // shared between leaderboards
```
nandy-andy: Just a reminder of our offline discussion: many of our ad slots in real life accepts multiple sizes, so in future we may need here a list of widths.

tests/pages/animations-ad.page.js
```diff
@@ -0,0 +1,20 @@
+const scrollWaitTime = 500;
+
+class AnimationsAd {
+	constructor() {
+		this.pageLink = 'slots/animations/';
+		this.topLeaderboardHeightHidden = 0;
```
nandy-andy: How about `s/topLeaderboardHeightHidden/topLeaderboardHeightWhenHidden/`?

tests/pages/animations-ad.page.js
```diff
@@ -0,0 +1,20 @@
...
+		this.topLeaderboardStyle = 'style';
```
nandy-andy: Why do we need it as a property?

tests/pages/animations-ad.page.js
```diff
@@ -0,0 +1,20 @@
...
+		this.collapsedAdValue = 'max-height: 0px;';
```
nandy-andy: It's hard to connect name of this property with its value. Consider renaming it, please :)

tests/pages/delay-ad.page.js
```diff
@@ -0,0 +1,25 @@
...
+const loadAdTime = 10000;
+
+class DelayAd {
+	constructor() {
+		this.pageLink = 'slots/delay/';
+		this.loadAdsButton = '#clickDelay';
+		this.resultAttribute = 'data-slot-result';
+		this.viewedAttribute = 'data-slot-viewed';
+		this.adLoaded = 'success';
+		this.adViewed = 'true';
```
nandy-andy: Do we really need all of those properties? I'm sceptical about `resultAttribute`, `viewedAttribute` and `adViewed` but I'm afraid I don't see the big picture ;)

tests/pages/animations-ad.page.js
nandy-andy: Oh, I found the usage: https://github.com/Wikia/ad-engine/pull/175/files#diff-fc98113acafe1bac32df37b60cbfcdbcR106 What do you think of removing the property and using in the test `'style'` string instead?

tests/common/adSlots.js
ThisIsJesus: I remember about it. Currently all of them share the same value (hence the comment added here), but as soon as we need more widths/heights (read: when I start writing tests for adproducts), I will rename them to make everything clear.

nandy-andy: Sounds like a plan 👍 Perfect, thanks! :)

tests/pages/animations-ad.page.js
ThisIsJesus: Renamed to `topLeaderboardHeightWhenHidden`. You're right, this `Hidden` word should make it slightly clearer.

tests/pages/delay-ad.page.js
emadry: yes, since they can be reused and tests are more readable in this way than with 'random' strings

tests/specs/top-boxad-ad.test.js
```diff
@@ -12,12 +12,25 @@ describe('It will test top boxad page', () => {
 		browser.waitForVisible(topBoxad.topBoxadRail, timeouts.standard);
 
 		const size = browser.getElementSize(topBoxad.topBoxadRail);
+		const tableOfErrors = [];
 
-		expect(size.width)
-			.to
-			.equal(topBoxad.topBoxadWidth, 'Top boxad width incorrect');
-		expect(size.height)
+		try {
```
emadry: If you're not checking here visibility in viewport, then you can skip try catch block (since if one dimension is wrong, then it has to be fixed and you will check both values anyway - they aren't independent like visibility and size

tests/specs/top-leaderboard-ad.test.js
```diff
@@ -7,23 +8,35 @@ const { expect } = require('chai');
...
 	it('will test visibility of top leaderboard', () => {
-		const size = browser.getElementSize(topLeaderboard.topLeaderboard);
+		const size = browser.getElementSize(adSlots.topLeaderboard);
+		const tableOfErrors = [];
 
-		expect(size.width)
-			.to
-			.equal(topLeaderboard.topLeaderboardWidth, 'Top leaderboard width incorrect');
-		expect(size.height)
+		try {
```
emadry: same as above

tests/specs/animations-ad.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+	it('will test top leaderboard disappearing after 6 seconds', () => {
+		browser.waitUntil(() => browser.element(adSlots.topLeaderboard).getAttribute(animationsAd.topLeaderboardStyle) === animationsAd.collapsedAdMaxHeight, animationsAd.waitForAnimationsTime, 'Top leaderboard ad did not collapse', helpers.interval);
+		animationsAd.waitToScroll();
+
+		const topLeaderboardSize = browser.getElementSize(adSlots.topLeaderboard);
+		const tableOfErrors = [];
+
+		try {
+			expect(topLeaderboardSize.height)
```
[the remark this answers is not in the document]
ThisIsJesus: Removed.

tests/specs/btf-only-ad.test.js
```diff
@@ -0,0 +1,62 @@
...
+		browser.click(btfOnlyAd.finishQueueButton);
+		helpers.slowScroll(2500);
+		browser.waitForVisible(adSlots.incontentBoxad, timeouts.standard);
+	});
+
+	it('will test the visibility of btf ad after manually finishing the queue', () => {
+		const size = browser.getElementSize(adSlots.incontentBoxad);
+		const tableOfErrors = [];
+
+		try {
```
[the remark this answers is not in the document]
ThisIsJesus: Removed unnecessary try catch blocks from these.

PR #176, opened 2018-09-17
ADEN-7081 Add Krux service handler

fraszczakszymon:


README.md
```diff
@@ -44,6 +44,8 @@ npm install github:Wikia/ad-engine#v14.0.0
 |`options.video.moatTracking.partnerCode`|MOAT identifier|string|✔|
 |`options.video.moatTracking.sampling`|Sampling for MOAT tracking|string|✔|
 |`options.video.porvata.audio.exposeToSlot`|Decides whether Porvata stores `audio` flag in slot object|boolean|✔|
+|`services.krux.enabled`|Decides whether Krux is loaded (after calling `krux.call()`)|boolean|✘|
+|`services.krux.id`|Krux ID|string|✔|
```
nandy-andy: ❤️ for updating docs :)

src/ad-services/krux/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Gets Krux data from localStorage
+ * @param {string} key
+ * @returns {string}
+ */
+function getKruxData(key) {
+	if (window.localStorage) {
+		return window.localStorage[key];
+	} else if (window.navigator.cookieEnabled) {
```
nandy-andy: Not directly connected with the scope of this task and feel free to create a follow-up ticket. Can we check how many Krux data comes from cookies? Or can we easily test if the Krux cookie is ported in a domain changes?

nandy-andy: Probably it's quite small since we have many different domains for one community but the thing we don't know the number is bothering ;)

PR #178, opened 2018-09-19
ADEN-7530 Adding Chesire Cat to ad-services

dmnsobczak:


examples/context.js
```diff
@@ -249,19 +249,33 @@ export default {
 						dfp_targeting: true,
 						on_1: ['logResult']
 					}
+				],
+				cheshire_cat: [
```
fraszczakszymon: Keep it consistent please :) `cheshirecat`

examples/bidders/reusable-prebid/index.html
```diff
@@ -12,8 +12,8 @@
 
 			(function() {
 				var gads = document.createElement('script'),
-						useSSL = 'https:' == document.location.protocol,
-						node = document.getElementsByTagName('script')[0];
+					useSSL = 'https:' == document.location.protocol,
+					node = document.getElementsByTagName('script')[0];
```
fraszczakszymon: Could you please add info that you can use ?enabled-project on this test page?

src/ad-engine/services/slot-service.js
```diff
@@ -120,6 +120,10 @@ class SlotService {
 
 		return hasConflict;
 	}
+
+	isCatlapsed(adSlot) {
+		return context.get(`slots.${adSlot.getSlotName()}.catlapsed`);
```
fraszczakszymon: We don't need to create additional flag for that. You can use `adSlot.collapse('catlapsed');` and then get this state from slot service by `slotService.getState('incontent_boxad_1');`

src/ad-services/bill-the-lizard/index.js
```diff
@@ -114,17 +114,16 @@ class BillTheLizard {
...
 			utils.logger(logGroup, 'disabled');
 			return new Promise((resolve, reject) => reject(new Error('Disabled')));
 		}
 
 		const host = context.get('services.billTheLizard.host');
 		const endpoint = context.get('services.billTheLizard.endpoint');
-		const parameters = context.get('services.billTheLizard.parameters');
 		const timeout = context.get('services.billTheLizard.timeout');
-		const models = this.projectsHandler.getEnabledModels();
+		const { models, parameters } = this.projectsHandler.getEnabledModels(lazyCallProject);
```
fraszczakszymon: It doesn't look like `getEnabledModels`. What do you think about creating separate method for `getProjectParameters`?

src/ad-services/bill-the-lizard/index.js
```diff
@@ -114,17 +114,16 @@ class BillTheLizard {
 	 * Requests service, executes defined methods and parses response
 	 * @returns {Promise}
 	 */
-	call() {
+	call(lazyCallProject = null) {
```
fraszczakszymon: The idea of enabling models per pv seems outdated. What do you think about making `projectNames` parameter required. So in each call you have to pass list of projects and there is no way to call bill the lizard without project names?

fraszczakszymon: Then we won't need special flag `is_lazy_called`

src/ad-services/bill-the-lizard/projects-handler.js
```diff
@@ -30,28 +30,39 @@ export class ProjectsHandler {
 
 	/**
 	 * Returns all geo-enabled models' definitions based on enabled projects
-	 * @returns {ModelDefinition[]}
+	 * @returns {{models: ModelDefinition[], parameters: Object}}
 	 */
-	getEnabledModels() {
+	getEnabledModels(projectName = null) {
```
fraszczakszymon: Add some unit tests please :)

examples/bidders/reusable-prebid/index.html
dmnsobczak: Done.

examples/context.js
dmnsobczak: Renamed :)

src/ad-engine/services/slot-service.js
dmnsobczak: Great suggestion. Done.

src/ad-services/bill-the-lizard/index.js
dmnsobczak: Ok, done.

src/ad-services/bill-the-lizard/index.js
dmnsobczak: I think it doesn't have sense to have two executionally identical functions, because to prepare a parameters list I need to have enabledModels, but in the meantime those are converted from Object to Array (we are loosing keys) which makes impossible to get values from context without combinations. I just gonna rename this functions, is it fine?

src/ad-engine/services/btf-blocker-service.js
```diff
@@ -69,6 +69,10 @@ class BtfBlockerService {
 				slotService.disable(adSlot.getSlotName(), 'viewport-conflict');
 			}
 
+			if (slotService.isCatlapsed(adSlot)) {
```
fraszczakszymon: I believe we don't need to create special cases here. Slot should be already disabled by billTheLizard.executor at this moment

src/ad-engine/services/slot-service.js
```diff
@@ -120,6 +120,10 @@ class SlotService {
 
 		return hasConflict;
 	}
+
+	isCatlapsed(adSlot) {
```
fraszczakszymon: Seems not needed

examples/bidders/reusable-prebid/script.js
```diff
@@ -40,6 +42,27 @@ context.push('listeners.slot', {
 	}
 });
 
+if (enabledProjects) {
+	enabledProjects.split(',').forEach(name => billTheLizard.projectsHandler.enable(name));
```
fraszczakszymon: Seems not needed. Can we remove "enabling" functionality from projectsHandler?

examples/bidders/reusable-prebid/script.js
```diff
@@ -40,6 +42,27 @@ context.push('listeners.slot', {
...
+
+	billTheLizard.executor.register('logResult', (model, prediction) => {
+		console.log(`🦎 %c${model.name}`, 'font-weight: bold', `predicted ${prediction}`);
+
+		if (prediction === 1) {
```
fraszczakszymon: Could you please create separate method for that? Like:
```
billTheLizard.executor.register('catlapse', (model, prediction) => {
	const slots = document.querySelectorAll('.repeatable-boxad');
	if (slots.length > 0) {
		const slot = slots[slots.length - 1];
		slotService.disable(slot.id, 'catlapsed');
	}
});
```
and then define model in context:
```
	name: 'cheshirecat:0.0.1',
	countries: ['XX'],
	on_0: ['logResult'],
	on_1: ['catlapse', 'logResult']
```

src/ad-engine/services/btf-blocker-service.js
dmnsobczak: That is right. Removed.

src/ad-engine/services/slot-service.js
dmnsobczak: *delete_button_click*

examples/bidders/reusable-prebid/script.js
dmnsobczak: I didn't notice I can do that, sorry. Splited.

PR #181, opened 2018-09-25
ADEN-7692

ThisIsJesus:


tests/specs/floating-rail-ad.test.js
```diff
@@ -0,0 +1,128 @@
...
+
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.newsAndStories);
+		expect(browser.getUrl())
+			.to
+			.equal(helpers.newsAndStories);
+		helpers.closeNewTabs();
+	});
+
+	// Top Boxad tests
```
emadry: You can use multiple describes in one test.
Then you can have before each dedicated to slot and it's cleaner

tests/specs/hivi-uap-ad.test.js
```diff
@@ -0,0 +1,349 @@
...
+
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.newsAndStories);
+		expect(browser.getUrl())
+			.to
+			.equal(helpers.newsAndStories);
+		helpers.closeNewTabs();
+	});
+
+	it('will check if ui shows up after hover in top leaderboard player', () => {
```
emadry: UI is used in other test cases so it's redundant

tests/specs/hivi-uap-ad.test.js
```diff
@@ -0,0 +1,349 @@
...
+		browser.click(hiviUap.closeLeaderboardButton);
+		expect(browser.element(adSlots.topLeaderboard).getAttribute(hiviUap.slotResult))
+			.to
+			.equal(hiviUap.slotCollapsed, 'Slot has not collapsed');
+	});
+
+	it('will test top leaderboard unsticking after scroll', () => {
+		browser.waitForVisible(adSlots.topLeaderboard, timeouts.standard);
+		helpers.slowScroll(1000);
+		helpers.slowScroll(2000);
```
emadry: Remove one of the scrolls

tests/specs/hivi-uap-ad.test.js
```diff
@@ -0,0 +1,349 @@
...
+		}
+
+		expect(tableOfErrors.length, `Errors found: ${tableOfErrors.toString()}`)
+			.to
+			.equal(0);
+	});
+
+	it('will test line item id of top boxad', () => {
+		browser.waitForVisible(adSlots.topBoxad, timeouts.standard);
+		expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.lineItemParam))
```
emadry: Create method in AdSlots where you pass the slot and return bool

tests/specs/hivi-uap-ad.test.js
```diff
@@ -0,0 +1,349 @@
...
+		try {
+			expect(browser.isVisibleWithinViewport(adSlots.incontentBoxad), 'Incontent boxad not in viewport')
+				.to
+				.be
+				.true;
+		} catch (error) {
+			tableOfErrors.push(error.message);
+		}
+
+		expect(tableOfErrors.length, `Errors found: ${tableOfErrors.toString()}`)
```
emadry: Create helper to display tableOfErrors in more readable way (break lines, numeration)

tests/specs/hivi-uap-ad.test.js
```diff
@@ -0,0 +1,349 @@
...
+		expect(browser.getUrl())
+			.to
+			.equal(helpers.lukeSkywalkerLegacy);
+		helpers.closeNewTabs();
+	});
+
+	// Bottom Leaderboard tests
+
+	it('will test bottom leaderboard dimensions and visibility', () => {
+		helpers.slowScroll(6000);
```
emadry: Scroll Values are 'magic numbers' - try to move it to variables

wdio.conf.js
```diff
@@ -36,6 +36,18 @@ exports.config = {
...
 			'./tests/specs/empty-response.test.js',
 		],
+		hiviUapAd: [
+			'./tests/specs/hivi-uap-ad.test.js',
+		],
+		hiviUapStaticAd: [
+			'./tests/specs/hivi-uap-static-ad.test.js',
+		],
+		abcdAd: [
+			'./tests/specs/abcd-ad.test.js',
```
emadry: Group tests and keep one as "current test" to use for writing/fixing

tests/common/helpers.js
```diff
@@ -5,8 +5,13 @@ const pauseBetweenScrolls = 250;
 class Helpers {
 	constructor() {
 		this.newsAndStories = 'http://www.wikia.com/fandom';
+		this.lukeSkywalkerLegacy = 'http://fandom.wikia.com/articles/legacy-luke-skywalker';
+		this.fandomWikia = 'http://fandom.wikia.com/';
```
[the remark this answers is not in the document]
ThisIsJesus: After the discussion on adeng-poz we have 3 choices:
1. To change (unify) these links. 
2. To teach me how to change them.
3. To check if clicking on an ad just opens a new tab, without checking where it takes us.

Up to you which one we choose. I'm thinking about the last one.

tests/common/adSlots.js
```diff
@@ -2,13 +2,22 @@ class AdSlots {
...
 		this.bottomLeaderboard = '#bottom_leaderboard';
-		this.leaderboardWidth = 728; // shared between leaderboards
-		this.leaderboardHeight = 90; // shared between leaderboards
 		this.topBoxad = '#top_boxad';
 		this.incontentBoxad = '#incontent_boxad';
 		this.repeatableBoxad = '#repeatable_boxad_';
+		this.topBoxadRail = '.top-boxad.rail-module';
+		this.leaderboardWidth = 728; // shared between leaderboards
+		this.leaderboardHeight = 90; // shared between leaderboards
+		this.adProductsTopLeaderboardWidth = 1200; // UAP and ABCD leaderboards have the same width
```
nandy-andy: `adProducts` is quite wide term IMHO. I know Ad Operations label mentioned products in Google Ad Manager. What do you think of renaming the property to `this.uapFamilyTopLeaderboardWidth`?

nandy-andy: Although, UAP:JWP won't have top leaderboard, so it's not that good name. D'oh, naming is hard! :)

ThisIsJesus: It is hard as F., which is why everything is so mixed up here. In all honesty though, I try to rename it everytime I do some modifications to better reflect the current state. So if it's right now and won't be later, I will change it to keep it as clear as possible. ;)

tests/pages/hivi-uap-static-ad.page.js
```diff
@@ -0,0 +1,12 @@
+class HiviUapStatic {
+	constructor() {
+		this.pageLink = 'templates/hivi-uap-static/';
+		this.closeLeaderboardButton = 'button';
+		this.slotResult = 'data-slot-result';
```
nandy-andy: You may move it to `adSlots.js` since it's being set for every slot. However, I understand it's not needed now in other tests and that's why you've put it here (where it's needed). So, treat this comment as minor and it's perfectly fine to move it to `AdSlot` once we'll start checking the slot result in other tests :)

ThisIsJesus: Just like with names, I try to move stuff when I see it's needed elsewhere. I'll keep that in mind, though, and definitely consider moving it now too. :)

PR #182, opened 2018-09-28
ADEN-7694 Update A9 apstag with deals

fraszczakszymon:
⚠️ Major due to changed A9 context configuration


examples/context.js
```diff
@@ -12,18 +12,24 @@ export default {
...
-					[728, 90]
-				],
-				top_boxad: [
-					[300, 250]
-				]
-			},
-			slotsVideo: []
+				top_leaderboard: {
+					sizes: [[728, 90]],
+					slotId: 'TOP_LEADERBOARD'
```
fraszczakszymon: Discussed with Nef. We suspect that A9 video issues may be caused with new requesting "new" slot ids (`featured` vs `FEATURED`). Lets try setting up upper case bids on mobile-wiki (leaving lowercase slotnames)

src/ad-bidders/a9/index.js
```diff
@@ -34,15 +36,14 @@ export class A9 extends BaseBidder {
...
-
-		if (!this.loaded) {
+		if (!loaded) {
 			this.insertScript();
 			this.configureApstag();
 
 			const apsConfig = {
 				pubID: this.amazonId,
-				videoAdServer: 'DFP'
+				videoAdServer: 'DFP',
```
nandy-andy: I wonder when they'll change it to `'GAM'` :trollface:

fraszczakszymon: 🙊

PR #183, opened 2018-09-28
ADEN-7702 Use notion of firstCall and ATF correctly.

jacekbj:


fraszczakszymon:
Major? :)

examples/bidders/reusable-prebid/script.js
```diff
@@ -104,7 +104,7 @@ events.on(events.AD_SLOT_CREATED, (slot) => {
 
 new AdEngine().init();
 
-btfBlockerService.finishAboveTheFold();
+btfBlockerService.finishFirstCall();
```
fraszczakszymon: Should we rename service? 🙈

spec/ad-engine/models/ad-slot.spec.js
```diff
@@ -3,16 +3,17 @@ import { AdSlot } from '../../../src/ad-engine/models/ad-slot';
 import { context } from '../../../src/ad-engine/services/context-service';
 import ConfigMock from '../config-mock';
 
+function createAdSlot(/** string */ id) {
```
fraszczakszymon: What do you think about keeping it consistent with other files? (keeping types in jsdoc above method)

src/ad-engine/services/btf-blocker-service.js
```diff
@@ -52,12 +54,15 @@ class BtfBlockerService {
...
-		logger(logGroup, 'ATF queue finished');
+	finishFirstCall() {
+		this.firstCallEnded = true;
+		logger(logGroup, 'first call queue finished');
 
 		if (window.ads.runtime.disableBtf) {
-			disableBtf(this.unblockedSlots);
+			const slotConfigs = context.get('slots');
+			disableSecondCall(
+				this.unblockedSlotNames.filter(adSlotName => AdSlot.isSlotAboveTheFold(slotConfigs[adSlotName])),
```
fraszczakszymon: Shouldn't it base on first call flag instead of ATF?

src/ad-engine/models/ad-slot.js
```diff
@@ -96,8 +96,12 @@ export class AdSlot extends EventEmitter {
 		}
 	}
 
-	isAboveTheFold() {
-		return !!this.config.aboveTheFold;
+	static isSlotAboveTheFold(config) {
```
fraszczakszymon: Seems not needed if: https://github.com/Wikia/ad-engine/pull/183/files#r221507656

src/ad-engine/services/btf-blocker-service.js
fraszczakszymon: Also - previously we were passing here all ATF/BTF slots, now it seems that array will contain ATF only. Is it ok?

examples/bidders/reusable-prebid/script.js
jacekbj: It still blocks BTF. Now it explicitly rules SRA. I think the latter task is much more important and should be mirrored in service name. I did not change it because I was afraid of too broad change scope. I think it should be changed, even though it's more work :|

spec/ad-engine/models/ad-slot.spec.js
jacekbj: Good point! This kind of annotation we can use with anonymous functions defined as callbacks.

src/ad-engine/services/btf-blocker-service.js
jacekbj: I changed this code to respect slots enabled by i.e. templates.

src/ad-engine/services/slot-service.js
```diff
@@ -120,6 +121,15 @@ class SlotService {
 
 		return hasConflict;
 	}
+
+	/**
+	 * Returns ATF slots.
+	 * @returns {AdSlot[]} ATF slots
+	 */
+	getAboveTheFoldSlots() {
```
jacekbj: This could be a getter but getters are hard to mock.

PR #184, opened 2018-10-01
ADEN-7608 Add Labrador:Basset functionality

dmnsobczak:


src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,106 @@
...
 	samplingSeparator = '/';
 
-let geoData = null,
-	cache = {};
+let cache = {},
+	cookieLoaded = false,
+	geoData = null,
+	sessionId = '';
+
+function hasCookie(countryList) {
```
fraszczakszymon: s/hasCookie/hasCache?

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,106 @@
 import Cookies from 'js-cookie';
 import Random from './random';
 
-const earth = 'XX',
+const cookieMarker = '-cached',
```
fraszczakszymon: What about `cacheMarker`?

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,106 @@
...
+		withCookie
 	};
+
+	if (withCookie) {
+		synchronizeCookie();
+	}
+}
+
+function readSession() {
+	sessionId = sessionId || Cookies.get('wikia_session_id') || 'ae3';
```
fraszczakszymon: Can we get name of cookie from context? e.g. `options.sessionIdCookieName`

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,106 @@
...
+
+	Object.keys(cache).forEach((variable) => {
+		if (cache[variable].withCookie) {
+			cachedVariables[variable] = cache[variable];
+		}
+	});
+
+	Cookies.set(`${sessionId}_basset`, JSON.stringify(cachedVariables), {
+		path: '/',
+		domain: domain.length > 1 ? `.${domain[domain.length - 2]}.${domain[domain.length - 1]}` : undefined,
```
fraszczakszymon: What about `context.get('options.session.cookieDomain')`?

src/ad-engine/utils/geo.js
fraszczakszymon: What about storing it in the context? `context.get('options.session.id')`

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,106 @@
...
+	const cachedVariables = {};
+	const domain = (window.location.hostname).split('.');
+
+	Object.keys(cache).forEach((variable) => {
+		if (cache[variable].withCookie) {
+			cachedVariables[variable] = cache[variable];
+		}
+	});
+
+	Cookies.set(`${sessionId}_basset`, JSON.stringify(cachedVariables), {
```
fraszczakszymon: Correct me if I am wrong, but it means that cookie expires with closed browser, right? Session cookie expires after 30mins. Basset cookie should last longer/the same (configurable) time

src/ad-engine/utils/geo.js
dmnsobczak: I used that name at the first attempt :D Renamed.

src/ad-engine/utils/geo.js
dmnsobczak: Si, renamed.

src/ad-engine/utils/geo.js
dmnsobczak: Discussed offline: it's to set proper wikia.com or fandom.com domain. No need to move into app code.

src/ad-engine/utils/geo.js
dmnsobczak: Yes, you right, I missed that part. Unfortunately we have rewrite cookie every pageview to make it up-to-date with session cookie.

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,131 @@
...
+
+	if (withCookie) {
+		synchronizeCookie();
+	}
+}
+
+function readSession() {
+	let sid = sessionId || context.get('options.session.id');
+
+	if (!sid) {
```
fraszczakszymon: What if user is inactive for an hour on mobile-wiki?

src/ad-engine/utils/geo.js
```diff
@@ -1,54 +1,131 @@
...
+		withCookie
 	};
+
+	if (withCookie) {
+		synchronizeCookie();
+	}
+}
+
+function readSession() {
+	let sid = sessionId || context.get('options.session.id');
```
fraszczakszymon: Can we have single getter method which is the only one source of truth? We should re-check value on page change event

dmnsobczak: OK, done.

PR #187, opened 2018-10-03
ADEN-7006 Handle slots with multiple positions in targeting

fraszczakszymon:


spec/ad-engine/services/slot-service.spec.js
```diff
@@ -62,7 +62,27 @@ describe('slot-service', () => {
...
+		expect(slotService.get('FAKE_AD')).to.equal(adSlot);
+	});
+
+	it('getter for slot with multiple positions', () => {
+		slotService.add(adSlot);
+
+		expect(slotService.get('FAKE_AD,FOO')).to.equal(adSlot);
+	});
+
+	it('getter for with different single targeting.pos', () => {
```
nandy-andy: A word is missing here, I'm afraid. Is it supposed to read: `getter for slot with different single targeting.pos`?

PR #188, opened 2018-10-04
Aden 7737 - Track UAP - twitch data to DW

pnather:


src/ad-products/styles/_twitch.scss
```diff
@@ -0,0 +1,13 @@
+.twitch-player-right {
+  z-Index: 1;
```
fraszczakszymon: Should it be `z-index`?

src/ad-products/styles/_twitch.scss
```diff
@@ -0,0 +1,13 @@
...
+  position: absolute;
+  top: 0;
+  right: 0;
+}
+
+.twitch-player-left {
+  z-Index: 1;
```
fraszczakszymon: Same here

PR #190, opened 2018-10-05
ADEN-7747 Default/Resolved state in UAP, UAP JWP, Twitch, AdEngine code refactor

ThisIsJesus:


tests/common/helpers.js
```diff
@@ -94,6 +96,58 @@ class Helpers {
 		});
 		return finalString;
 	}
+
+	/**
+	 * Provides parameters with the example page to load and ad slot to wait for
+	 * @param adPage example page with ads to load
+	 * @param adSlot ad slot to wait for visible
+	 */
+	reloadPage(adPage, adSlot) {
```
fraszczakszymon: what about `reloadPageAndWaitForSlot`?

tests/common/helpers.js
```diff
@@ -94,6 +96,58 @@ class Helpers {
...
+	 */
+	waitForNewTab() {
+		browser.waitUntil(() => browser.getTabIds().length > 1, timeouts.standard, 'Tab has not been opened', timeouts.interval);
+	}
+
+	/**
+	 * Waits until the ad slot to receive its line item id parameter.
+	 * @param adSlot ad slot that should receive the parameter
+	 */
+	waitForLineItemParam(adSlot) {
```
fraszczakszymon: What about `waitForLineItemIdAttribute`?

tests/common/helpers.js
```diff
@@ -94,6 +96,58 @@ class Helpers {
...
+		browser.waitUntil(() => browser.element(adSlot).getAttribute(adSlots.lineItemParam) !== null, timeouts.standard, 'No line item id attribute', timeouts.interval);
+	}
+
+	/**
+	 * Waits until the element is visible and its height is greater than 0.
+	 * @param adSlot ad slot we are waiting for
+	 */
+	waitForExpanded(adSlot) {
```
fraszczakszymon: What about `waitForAdSlotExpanded`?

tests/common/helpers.js
ThisIsJesus: I can do that for readability's sake. Would you like me to change the variable's name too? I think I should to keep it consistent.

tests/specs/abcd-ad.test.js
```diff
@@ -5,144 +5,141 @@ import helpers from '../common/helpers';
...
 
-		it('will test line item id', () => {
-			expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.lineItemParam))
-				.to
-				.equal(abcdAd.topBoxadLineItemId, 'Line item ID mismatch');
-		});
+		const tabIds = browser.getTabIds();
+
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.fandomWord);
```
fraszczakszymon: What's the point of this variable? In case we need to change `"fandom"` to `"foo"` we will need to rename variable also. Shouldn't it be `helpers.companyName = 'fandom';` / `helpers.clickThroughUrlDomain = 'fandom';` or something similar?

tests/common/helpers.js
ThisIsJesus: Can do. :)

tests/common/helpers.js
ThisIsJesus: Changed both the method's name and variable's name.

tests/specs/empty-response.test.js
```diff
@@ -5,29 +5,22 @@ import helpers from '../common/helpers';
...
 
-	it('will test if top leaderboard ad is not visible', () => {
+	it('Check if top leaderboard ad is not visible', () => {
 		browser.isExisting(adSlots.topLeaderboard, timeouts.standard);
-		expect(browser.isExisting(`${adSlots.topLeaderboard}${helpers.classHidden}`), 'Element does not exist')
-			.to
-			.be
-			.true;
+		browser.waitForExist(`${adSlots.topLeaderboard}${helpers.classHidden}`, timeouts.standard);
 		expect(browser.isVisibleWithinViewport(adSlots.topLeaderboard), 'Top leaderboard not visible in viewport')
```
fraszczakszymon: I believe we should check whether slot is collapsed on this page (0 height)

tests/specs/empty-response.test.js
```diff
@@ -5,29 +5,22 @@ import helpers from '../common/helpers';
...
 	});
-	it('will test if top boxad is not visible', () => {
-		browser.isExisting(adSlots.topBoxad, timeouts.standard);
-		expect(browser.isExisting(`${adSlots.topBoxad}${helpers.classHidden}`), 'Top boxad exists in code')
-			.to
-			.be
-			.true;
+	it('Check if top boxad is not visible', () => {
+		browser.waitForExist(`${adSlots.topBoxad}${helpers.classHidden}`, timeouts.standard);
 		expect(browser.isVisibleWithinViewport(adSlots.topBoxad), 'Top boxad is visible in viewport')
```
fraszczakszymon: I believe we should check whether slot is collapsed on this page (0 height)

tests/specs/floating-ad.test.js
```diff
@@ -5,139 +5,139 @@ import helpers from '../common/helpers';
...
-				expect(browser.isVisibleWithinViewport(adSlots.topLeaderboard), 'Top leaderboard not in viewport')
-					.to
-					.be
-					.true;
-			} catch (error) {
-				tableOfErrors.push(error.message);
-			}
-
-			expect(tableOfErrors.length, helpers.errorFormatter(tableOfErrors))
+	it('Check dimensions and visibility', () => {
```
fraszczakszymon: In my opinion we have too many duplicated tests. We don't need to check anything related to TLB on this page. In case of any failure related to TLB we'll get bunch of falsy failures.  This page is about floating boxad.

Could we make tests/assertions more specific for example pages? So once something fails we know that this is related to XYZ

tests/specs/floating-rail-ad.test.js
```diff
@@ -5,123 +5,134 @@ import helpers from '../common/helpers';
 
 const { expect } = require('chai');
 
-describe('It will test floating rail ads', () => {
+let theUrlYouWantToGet;
+
+global.client.on('Network.responseReceived', (params) => {
```
fraszczakszymon: Can we clean it up/move to another branch? ;)

tests/specs/floating-rail-ad.test.js
```diff
@@ -5,123 +5,134 @@ import helpers from '../common/helpers';
...
+	const { url, status } = params.response;
+	if (url.includes('gampad/ads?gdfp')) {
+		theUrlYouWantToGet = url;
+	}
+});
+
+describe('Floating rail ads page: top leaderboard', () => {
```
fraszczakszymon: Not needed imo

tests/common/adSlots.js
```diff
@@ -8,16 +8,22 @@ class AdSlots {
 		this.topBoxadRail = '.top-boxad.rail-module';
 		this.leaderboardWidth = 728; // shared between leaderboards
 		this.leaderboardHeight = 90; // shared between leaderboards
-		this.adProductsTopLeaderboardWidth = 1200; // UAP and ABCD leaderboards have the same width
-		this.uapTopLeaderboardHeight = 300;
+		this.adProductsTopLeaderboardWidth = 1920; // UAP and ABCD leaderboards have the same width
+		this.uapTopLeaderboardHeight = 480;
```
fraszczakszymon: Some of this dimensions are strictly connected to browser size (e.g. this one). On the other hand we know that aspect ratio of ad in default state is 4:1, in resolved 10:1 and 16:9 on mobile. Should we calculate height based on browser width? (`browserWidth * 1/4`)

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
+			expect(defaultSize.width)
+				.to
+				.equal(adSlots.adProductsTopLeaderboardWidth, 'Default width incorrect');
+			expect(defaultSize.height)
+				.to
+				.equal(adSlots.uapTopLeaderboardHeight, 'Default height incorrect');
+		} catch (error) {
+			tableOfErrors.push(error.message);
+		}
+		try {
```
fraszczakszymon: Still I believe we should create a wrapper for that

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
+		browser.waitForVisible(adSlots.incontentBoxad, timeouts.standard);
+	});
+
+	it('Check dimensions and visibility', () => {
+		const size = browser.getElementSize(adSlots.incontentBoxad);
+		const tableOfErrors = [];
 
-		it('will test line item id', () => {
-			expect(browser.element(adSlots.bottomLeaderboard).getAttribute(adSlots.lineItemParam))
+		try {
```
fraszczakszymon: I believe we could have helper for checking dimensions. This code seems to be duplicated many times.

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
 
-		it('will test redirect on click', () => {
-			browser.click(adSlots.bottomLeaderboard);
+		expect(tableOfErrors.length, helpers.errorFormatter(tableOfErrors))
+			.to
+			.equal(0);
+	});
+
+	it('Check line item id', () => {
+		helpers.waitForLineItemParam(adSlots.topLeaderboard);
```
fraszczakszymon: I believe we could have helper for checking line item id. This code seems to be duplicated many times.

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
+		expect(browser.element(adSlots.incontentBoxad).getAttribute(adSlots.lineItemParam))
+			.to
+			.equal(hiviUap.bottomLineItemId, 'Line item ID mismatch');
+	});
 
-			const tabIds = browser.getTabIds();
+	it('Check redirect on click', () => {
+		helpers.waitForLineItemParam(adSlots.incontentBoxad);
```
fraszczakszymon: I believe we could have helper for checking redirection. This code seems to be duplicated many times.

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.fandomWord);
+		expect(browser.getUrl())
+			.to
+			.include(helpers.fandomWord);
+		helpers.closeNewTabs();
+	});
+});
+
+describe('Hivi uap ads page: bottom leaderboard', () => {
```
fraszczakszymon: s/Hivi uap/HiVi UAP

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
+		browser.switchTab(tabIds[1]);
+		helpers.waitForUrl(helpers.fandomWord);
+		expect(browser.getUrl())
+			.to
+			.include(helpers.fandomWord);
+		helpers.closeNewTabs();
+	});
+});
+
+describe('Hivi uap ads page: video player in bottom leaderboard', () => {
```
fraszczakszymon: BFAB deserves more test cases:
- it's in default state on first pv
- it's in resolved state on second pv
- it's in resolved state once video finishes
- video can be replayed

tests/specs/hivi-uap-ad.test.js
```diff
@@ -5,312 +5,396 @@ import helpers from '../common/helpers';
...
-					.true;
-			} catch (error) {
-				tableOfErrors.push(error.message);
-			}
-
-			expect(tableOfErrors.length, helpers.errorFormatter(tableOfErrors))
-				.to
-				.equal(0);
-		});
+describe('Hivi uap ads page: video player in top leaderboard', () => {
```
fraszczakszymon: Can we check whether video can be replayed too?

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,134 @@
+import twitchAd from '../pages/hivi-uap-twitch-ad.page';
+import adSlots from '../common/adSlots';
+import { timeouts } from '../common/timeouts';
+import helpers from '../common/helpers';
+
+const { expect } = require('chai');
+
+describe('Twitch ads page: top leaderboard', () => {
```
fraszczakszymon: Imo we don't need test cases about dimensions/redirect here

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,134 @@
...
+
+	it('Check Twitch player visibility', () => {
+		const myFrame = $(twitchAd.playerFrame).value;
+
+		browser.frame(myFrame);
+		browser.waitForVisible(twitchAd.twitchPlayer, timeouts.standard);
+	});
+});
+
+describe('Twitch ads page: top boxad', () => {
```
fraszczakszymon: Not sure if it's needed (except top_boxad line item id, which confirms that it's from the same campaign)

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,134 @@
...
+		} catch (error) {
+			tableOfErrors.push(error.message);
+		}
+
+		expect(tableOfErrors.length, helpers.errorFormatter(tableOfErrors))
+			.to
+			.equal(0);
+	});
+
+	it('Check line item id', () => {
```
fraszczakszymon: Can we make descriptions more descriptive? E.g. Check if top boxad is from the same campaign

tests/specs/porvata.test.js
```diff
@@ -53,38 +53,26 @@ describe('It will test porvata player', () => {
...
 		browser.click(porvata.fullscreenButton);
-		expect(browser.isExisting(porvata.stopScrolling))
-			.to
-			.be
-			.true;
+		browser.waitForExist(porvata.stopScrolling, timeouts.standard);
 	});
 
-	it('will test closing the player', () => {
+	it('Check test closing the player', () => {
```
fraszczakszymon: Check closing the player

tests/specs/porvata.test.js
```diff
@@ -4,25 +4,25 @@ import helpers from '../common/helpers';
 
 const { expect } = require('chai');
 
-describe('It will test porvata player', () => {
+describe('Porvata player ads', () => {
```
fraszczakszymon: We could test replay functionality too

tests/specs/repeatable-slots.test.js
```diff
@@ -5,23 +5,23 @@ import helpers from '../common/helpers';
 
 const { expect } = require('chai');
 
-describe('It will test repeatable slots ads ', () => {
+describe('Repeatable slots ads', () => {
```
fraszczakszymon: Imo we're missing one important test case - checking whether there is at least one viewport space between slots

tests/specs/viewport-conflict-ad.test.js
```diff
@@ -94,10 +104,10 @@ describe('It will test viewport conflicts ad page', () => {
 		helpers.closeNewTabs();
 	});
 
-	it('will test if top boxad is hidden after clicking the button', () => {
+	it('Check if top boxad is hidden after clicking the button', () => {
 		browser.waitForVisible(viewportConflictAd.hideBoxadButton, timeouts.standard);
 		browser.click(viewportConflictAd.hideBoxadButton);
-		browser.waitUntil(() => browser.element(adSlots.topBoxad).getAttribute(viewportConflictAd.dataSlotResult) === 'collapse', timeouts.standard, 'Ad was not hidden', helpers.interval);
+		browser.waitUntil(() => browser.element(adSlots.topBoxad).getAttribute(adSlots.resultAttribute) === 'collapse', timeouts.standard, 'Ad was not hidden', timeouts.interval);
 		expect(browser.isExisting(`${adSlots.topBoxad}${helpers.classHidden}`))
```
fraszczakszymon: After that we should scroll a bit down and check whether bottom_leaderboard is displayed (it does not conflict with other slots in the same viewport)

wdio.conf.js
```diff
@@ -69,6 +75,10 @@ exports.config = {
 	mochaOpts: {
 		ui: 'bdd',
 		compilers: ['js:babel-core/register'],
-		timeout: 20000
+		timeout: 200000
 	},
+	before() {
+		// eslint-disable-next-line no-undef
```
fraszczakszymon: Should we exclude whole this file from linting?

fraszczakszymon:
Fix linter please

tests/common/adSlots.js
ThisIsJesus: @jacekbj pointed this out earlier. This is why we added this line of code in wdio.conf.js:
```
before() {
		// eslint-disable-next-line no-undef
		browser.windowHandleSize({ width: 1920, height: 1080 });
	}
```

The above is to make sure that there are no discrepancies between devices we run tests on. Nonetheless, your idea of calculating height based on browser's width is also good. I need to consult it and think about it a little bit.

tests/specs/porvata.test.js
ThisIsJesus: Silly me, forgot to remove "test" from here! :< Fixing it right now!

tests/specs/empty-response.test.js
ThisIsJesus: After the discussion we decided to do waitUntil on data-slot-result to see if it is collapsed (slot's dimensions don't change, so it would immediately fail).

tests/specs/empty-response.test.js
ThisIsJesus: After the discussion we decided to do waitUntil on data-slot-result to see if it is collapsed (slot's dimensions don't change, so it would immediately fail).

tests/specs/abcd-ad.test.js
ThisIsJesus: I decided to go with the second option. Thank you! :)

tests/specs/porvata.test.js
ThisIsJesus: Done. I added such case. :)

tests/specs/floating-rail-ad.test.js
ThisIsJesus: Yes, it will be cleaned up.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done, though it hardly made any difference.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done.

tests/common/adSlots.js
ThisIsJesus: We decided to go with the option mentioned by you anyway. :)

tests/specs/floating-rail-ad.test.js
ThisIsJesus: Removed.

tests/common/adSlots.js
ThisIsJesus: I am currently working on refactoring the code to get rid of the fixed value for UAP slots anyway using something similar to what you suggested. I will resolve this conversation as soon as I am done.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Tests for visibility and dimensions are now two separate its, so there is no try catch anymore.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-ad.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-twitch-ad.test.js
ThisIsJesus: Removed from here, remains only in common-slots.

tests/specs/hivi-uap-twitch-ad.test.js
ThisIsJesus: Done. They should provide more information now.

wdio.conf.js
ThisIsJesus: I don't think it's necessary. This is the only line causing problems, hence the ignore over here. I'd leave it like this.

tests/specs/repeatable-slots.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+		expect(browser.isVisibleWithinViewport(adSlots.topBoxad))
+			.to
+			.be
+			.true;
+		expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.lineItemIdAttribute))
+			.to
+			.equal(hiviUapJwp.inHouseLineItemId, 'Wrong ad loaded');
+	});
+
+	it('Check if top boxad shows up after clicking the button, if it was viewed and if it is uap ad', () => {
```
jacekbj: Definately this scenario deserves a separate describe.

describe: click on button
   before: click
   it should be visible in viewport
   it should have result attribute set to ${adSlots.adLoaded}
   it should have line item id ${lineItemIdAttribute}

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		browser.waitForVisible(hiviUapJwp.loadAdsButton, timeouts.standard);
+	});
+	afterEach(() => {
+		browser.scroll(0, 0);
+	});
+
+	it('Check if incontent boxad is not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.incontentBoxad}${adSlots.resultAttribute}`), 'Incontent boxad visible')
```
jacekbj: Mixed concept of existing and visibility.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+			.to
+			.be
+			.false;
+	});
+
+	it('Check if the ad loaded after delay is the inhouse one', () => {
```
jacekbj: it also checks if it's visible - split into separate `it`. It's OK that they depend on each other.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+		expect(browser.isVisibleWithinViewport(adSlots.incontentBoxad))
+			.to
+			.be
+			.true;
+		expect(browser.element(adSlots.incontentBoxad).getAttribute(adSlots.lineItemIdAttribute))
+			.to
+			.equal(hiviUapJwp.inHouseLineItemId, 'Wrong ad loaded');
+	});
+
+	it('Check if incontent boxad shows up after clicking the button, if it was viewed and if it is UAP ad', () => {
```
jacekbj: a separate describe, please :)

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+describe('Twitch ads page: top leaderboard', () => {
+	let adStatus;
+
+	before(() => {
+		browser.url(twitchAd.pageLink);
+	});
+
+	beforeEach(() => {
+		browser.waitForVisible(adSlots.topLeaderboard, timeouts.standard);
```
jacekbj: Do we have to wait for top leaderboard visibility before each test?

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+	it('Check if slot is visible', () => {
+		expect(adStatus.inViewport, 'Not in viewport')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if dimensions are correct', () => {
+		const dimensions = helpers.checkUAPSizeSlotRatio(adSlots.topLeaderboard, 3.88);
```
jacekbj: MAGIC NUMBER DETECTED
Please move the magic number to a const a bit higher up, to page object or at least name it :)

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,107 @@
...
+	});
+
+	it('Check if line item id is from the correct campaign', () => {
+		helpers.waitForLineItemIdAttribute(adSlots.topLeaderboard);
+		expect(helpers.getLineItemId(adSlots.topLeaderboard))
+			.to
+			.equal(twitchAd.topLeaderboardLineItemId, 'Line item ID mismatch');
+	});
+
+	it('Check if leaderboard does not obstruct the navbar', () => {
```
jacekbj: ditto

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,107 @@
...
+		expect(browser.isVisibleWithinViewport(helpers.navbar), 'Navbar not visible')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if redirect on click works properly', () => {
```
jacekbj: Do we have checks if something works incorrectly :trollface: 
I'm OK with the construction "check if ..." but I will write my tests "should redirect on click", which will generate in my opinion nicer test output.

tests/specs/hivi-uap-twitch-ad.test.js
```diff
@@ -0,0 +1,107 @@
...
+	});
+});
+
+describe('Twitch ads page: player', () => {
+	before(() => {
+		browser.url(twitchAd.pageLink);
+	});
+
+	beforeEach(() => {
+		browser.waitForVisible(twitchAd.twitchPlayer);
```
jacekbj: ditto

wdio.conf.js
```diff
@@ -33,23 +36,25 @@ exports.config = {
 			'./tests/specs/porvata.test.js',
 		],
 		currentTest: [
-
+			// spot for a test that is currently being worked on
```
jacekbj: Let's promote running tests instead:
`npm run wdio -- --spec path.to.my.spec.test.js`

it's just to tempting to commit somthing in here

tests/specs/repeatable-slots.test.js
```diff
@@ -80,4 +55,23 @@ describe('It will test repeatable slots ads ', () => {
 			.be
 			.false;
 	});
+
+	it('Check if there is at least one viewport between slots', () => {
+		browser.url(helpers.addParametersToUrl(repeatableSlots.pageLink, [repeatableSlots.setLengthOfContent(5)]));
+		repeatableSlots.scrollBetweenBoxads(repeatableSlots.getRepeatableSlot(1));
```
jacekbj: Something is wrong with `helpers.slowScroll` 

```
slowScroll(px, scrollFromElement = null) {
		if (scrollFromElement !== null) {
			for (let i = (px / valueToDivideBy); i < px; i += i) {
				browser.scroll(scrollFromElement, 0, i);
				browser.pause(pauseBetweenScrolls);
			}
		} else {
			for (let i = (px / valueToDivideBy); i < px; i += i) {
				browser.scroll(0, i);
				browser.pause(pauseBetweenScrolls);
			}
		}
	}
```
shouldn't that be
```
slowScroll(px, scrollFromElement = null) {
		if (scrollFromElement !== null) {
			for (let i = (px / valueToDivideBy); i < px; i += i) {
				browser.scroll(scrollFromElement, 0, i * valueToDivideBy);  // !HERE!
				browser.pause(pauseBetweenScrolls);
			}
		} else {
			for (let i = (px / valueToDivideBy); i < px; i += i) {
				browser.scroll(0, i * valueToDivideBy);
				browser.pause(pauseBetweenScrolls);
			}
		}
	}
```

jacekbj: BTW browser.scroll supports 3 arguments?
http://webdriver.io/api/utility/scroll.html#Usage

jacekbj: Can you add comments on where the viewport is? I can't imagine how the scrolling works (I have a very limited imagination).

package.json
```diff
@@ -13,7 +13,8 @@
     "current-device": "0.7.8",
     "eventemitter3": "3.0.1",
     "js-cookie": "2.2.0",
-    "lodash": "^4.17.4"
+    "lodash": "^4.17.4",
+    "npm": "6.4.1"
```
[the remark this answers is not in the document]
emilmadry: If you mean "npm", then it's not

tests/common/adSlots.js
```diff
@@ -5,19 +5,20 @@ class AdSlots {
...
-		this.leaderboardWidth = 728; // shared between leaderboards
-		this.leaderboardHeight = 90; // shared between leaderboards
-		this.adProductsTopLeaderboardWidth = 1200; // UAP and ABCD leaderboards have the same width
-		this.uapTopLeaderboardHeight = 300;
-		this.uapBottomLeaderboardWidth = 1024;
-		this.uapBottomLeaderboardHeight = 102;
-		this.abcdLeaderboardHeight = 240;
-		this.boxadWidth = 300; // shared between boxads
-		this.boxadHeight = 250; // shared between boxads
+		this.railModule = '.rail-module';
```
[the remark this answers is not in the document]
emadry: adding "selector" would require us to add it in a lot of places which would make it less readable.
"Module" because it's displayed on the template page as "rail module" and it has class .rail-module. 
By standard which was agreed by qa's we should keep naming close to the names in the code

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
-	 * @returns {string}
+	 * Provides parameters with the example page to load and ad slot to wait for.
+	 * @param adPage example page with ads to load
+	 * @param adSlot ad slot to wait for visible
+	 */
+	reloadPageAndWaitForSlot(adPage, adSlot) {
+		browser.reload();
+		browser.windowHandleSize({ width: 1920, height: 1080 });
+		browser.url(adPage); // mandatory, because test page fails to load without it
+		browser.waitForVisible(adSlot, timeouts.standard);
```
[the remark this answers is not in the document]
emadry: Yes. For viewport there's `isVisibleWithinViewport`

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+	}
+
+	/**
+	 * Refreshes the page and pauses all the actions to let them reload properly.
+	 * @param adSlot slot to wait for
+	 * @param timeout duration of the pause
+	 */
+	refreshPageAndWaitForSlot(adSlot, timeout = timeouts.pageReload) {
+		browser.refresh();
+		browser.pause(timeout);
```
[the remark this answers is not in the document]
emadry: Actually it should be changed to standard which takes up to 5 seconds and old one should be removed

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+	 * @param adSlot ad slot that should receive the parameter
+	 */
+	waitForLineItemIdAttribute(adSlot) {
+		browser.waitUntil(() => browser.element(adSlot).getAttribute(adSlots.lineItemIdAttribute) !== null, timeouts.standard, 'No line item id attribute', timeouts.interval);
+	}
+
+	/**
+	 * Returns line item ID of the given slot.
+	 * @param adSlot slot to get line item ID from
+	 * @returns {String|String[]|*|(WebdriverIO.Client<string> & WebdriverIO.Client<null> & string & null)|(WebdriverIO.Client<string[]> & WebdriverIO.Client<null[]> & string[] & null[])|WebdriverIO.Client<any>|string}
```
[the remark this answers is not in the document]
emadry: It was auto-generated by webstorm, so there's a chance, that it will ;)

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+	 * @returns {{status: boolean, capturedErrors: string}} returns false if no errors were found, else returns true. captured errors: returns string with errors
+	 */
+	checkUAPSizeSlotRatio(adSlot, heightRatio, customPrefix = '') {
+		return this.checkSlotRatio(adSlot, browser.getViewportSize('width'), heightRatio, customPrefix);
+	}
+
+	/**
+	 * Checks slot ratio based on a given derivative value.
+	 * @param adSlot slot to measure
+	 * @param sizeDeterminant derivative value for the slot
```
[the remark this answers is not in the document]
emadry: Yes. Description has to be updated

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+		this.waitForLineItemIdAttribute(adSlot);
+		browser.waitForEnabled(adSlot, timeouts.standard);
+		browser.click(adSlot);
+
+		const tabIds = browser.getTabIds();
+
+		browser.switchTab(tabIds[1]);
+		this.waitForUrl(url);
+
+		if (browser.getUrl().includes(url)) {
```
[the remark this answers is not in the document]
emadry: It has to close tab even if url is incorrect so other tests in this session won't fail

tests/specs/abcd-ad.test.js
```diff
@@ -5,144 +5,78 @@ import helpers from '../common/helpers';
 
 const { expect } = require('chai');
 
-describe('It will test abcd ads', () => {
-	beforeEach(() => {
+describe('ABCD ads page: top leaderboard', () => {
```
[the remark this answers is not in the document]
emadry: If I remember correctly - issue with allure reporting. @ThisIsJesus please confirm

tests/specs/abcd-ad.test.js
```diff
@@ -5,144 +5,78 @@ import helpers from '../common/helpers';
...
+		adStatus = helpers.checkSlotStatus(adSlots.topLeaderboard);
+	});
 
-		describe('It will test video player in leaderboard', () => {
-			beforeEach(() => {
-				helpers.waitToStartPlaying();
-			});
+	it('Check if slot is visible in viewport', () => {
+		expect(adStatus.inViewport, 'Not in viewport')
+			.to
```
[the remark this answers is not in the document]
emadry: It came from our standard but it has it's up sides. In case of change you see if entire assertion was changed or only a part of it so I'm not so sure if we should change it

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
 	/**
-	 * Reformats the errors from the array and prints them as a numbered list.
-	 * @param arrayOfErrors
-	 * @returns {string}
+	 * Provides parameters with the example page to load and ad slot to wait for.
+	 * @param adPage example page with ads to load
+	 * @param adSlot ad slot to wait for visible
+	 */
+	reloadPageAndWaitForSlot(adPage, adSlot) {
+		browser.reload();
```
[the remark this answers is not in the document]
emadry: Method for opening new url is `browser.url()` so it would look like `browser.url(url)` so it might not be very clear

tests/specs/common-slots-ad.test.js
```diff
@@ -0,0 +1,211 @@
...
+describe('Common slots: top leaderboard', () => {
+	let adStatus;
+
+	before(() => {
+		browser.url(commonAds.pageLink);
+		adStatus = helpers.checkSlotStatus(adSlots.topLeaderboard);
+	});
+
+	beforeEach(() => {
+		browser.waitForVisible(adSlots.topLeaderboard, timeouts.standard);
```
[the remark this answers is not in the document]
emadry: Some tests have more interactions, so they require page 'ready'

wdio.conf.js
emadry: We're running it with node runner in webstorm which allows us to use debugger. It's not possible with running it from command line

jacekbj: I think you can still use the `--spec` parameter. Please slack me your IDE configuration, I will check it.

tests/specs/abcd-ad.test.js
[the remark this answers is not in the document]
jacekbj: I can see your point, I'm not strongly for or against. :+1:

tests/specs/common-slots-ad.test.js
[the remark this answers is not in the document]
jacekbj: First 3 tests check properties of a slot and don't interact with it, so visibility should not change ;)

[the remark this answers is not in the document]
jacekbj: Only the last test does some interaction with the adslot, so how would it change in-between?

tests/common/helpers.js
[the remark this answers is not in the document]
ThisIsJesus: I think @emilmadry explained it clearly enough, so I don't have to. ;)

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+		browser.pause(timeout);
+		browser.waitForVisible(adSlot);
+	}
+
+	/**
+	 * Waits until the ad slot to receive its line item id parameter.
+	 * @param adSlot ad slot that should receive the parameter
+	 */
+	waitForLineItemIdAttribute(adSlot) {
+		browser.waitUntil(() => browser.element(adSlot).getAttribute(adSlots.lineItemIdAttribute) !== null, timeouts.standard, 'No line item id attribute', timeouts.interval);
```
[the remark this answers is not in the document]
ThisIsJesus: Will do! :)

tests/specs/abcd-ad.test.js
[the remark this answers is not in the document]
ThisIsJesus: I'd leave it like it is, at least for now.

tests/specs/btf-only-ad.test.js
```diff
@@ -1,58 +1,27 @@
...
 import helpers from '../common/helpers';
 
 const { expect } = require('chai');
 
-describe('It will test btf ads', () => {
-	beforeEach(() => {
+describe('BTF Only ads page: incontent boxad', () => {
+	let adStatus;
+
+	before(() => {
```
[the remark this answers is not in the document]
ThisIsJesus: I was thinking about doing that, but I was uncertain. @emadry please confirm it does not violate any rules/conventions.

tests/specs/hivi-uap-twitch-ad.test.js
ThisIsJesus: Yes, I think we do. There are 5 tests here and I think only one of them does not really require visibility. :)

tests/specs/hivi-uap-twitch-ad.test.js
ThisIsJesus: OH NOES! :(
Since you said "Please", I will. ;)

tests/common/helpers.js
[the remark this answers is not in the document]
ThisIsJesus: Standard it is, pageReload has been removed. :)

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+		if (slotSize.width !== width) {
+			result = false;
+			errorMessages += `${customPrefix} Width incorrect: expected ${slotSize.width} to equal ${width}\n`;
+		}
+		if (slotSize.height !== height) {
+			result = false;
+			errorMessages += `${customPrefix} Height incorrect: expected ${slotSize.height} to equal ${height}\n`;
+		}
+		return {
+			status: result,
```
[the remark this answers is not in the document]
ThisIsJesus: Done. :)

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+		return result;
+	}
+
+	/**
+	 * Checks slot\'s status after making sure it exists in the code.
+	 * Returns information about visibility in general, visibility in viewport and about being enabled.
+	 * @param adSlot slot to wait for
+	 * @returns {{visible: *, inViewport: (Boolean|Boolean[]), enabled: (Boolean|Boolean[])}} slot statuses
+	 */
+	checkSlotStatus(adSlot) {
```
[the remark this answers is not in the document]
ThisIsJesus: I think you're right, changed. :)

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+		}
+		this.closeNewTabs();
+		return result;
+	}
+
+	/**
+	 * Checks slot\'s status after making sure it exists in the code.
+	 * Returns information about visibility in general, visibility in viewport and about being enabled.
+	 * @param adSlot slot to wait for
+	 * @returns {{visible: *, inViewport: (Boolean|Boolean[]), enabled: (Boolean|Boolean[])}} slot statuses
```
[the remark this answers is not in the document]
ThisIsJesus: This is something Webstorm returned, so I'm not sure about this one. I'll go with your proposition anyway, as it's easy to revert it.

tests/specs/abcd-ad.test.js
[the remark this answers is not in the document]
ThisIsJesus: Not as much with Allure reporting as with making new sessions or something similar. Some tests required this and when describes were nested it failed to work.

tests/specs/abcd-ad.test.js
```diff
@@ -5,144 +5,78 @@ import helpers from '../common/helpers';
...
 
-	describe('It will test top boxad', () => {
-		beforeEach(() => {
-			browser.waitForVisible(adSlots.topBoxad, timeouts.standard);
-		});
-
-		it('will test dimensions and visibility', () => {
-			const size = browser.getElementSize(adSlots.topBoxad);
-			const tableOfErrors = [];
+	it('Check if line item id is from the proper campaign', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: The proposition was suggested by @fraszczakszymon, what you're saying is practically doing the same thing that assertion does (Error message looks like this: Line item mismatch: expected x to equal y). While I may change it, I am uncertain if it's necessary with the way we print error now.

tests/specs/floating-ad.test.js
```diff
@@ -1,143 +1,20 @@
...
 
 const { expect } = require('chai');
 
-describe('It will test floating ad page', () => {
-	beforeEach(() => {
+describe('Floating ad page: incontent boxad', () => {
+	before(() => {
 		browser.url(floatingAd.pageLink);
-		browser.waitForVisible(helpers.pageBody);
+		helpers.slowScroll(2500);
```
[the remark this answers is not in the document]
ThisIsJesus: Done, though with just one case here it doesn't make that much difference in anything but aesthetics.

tests/specs/floating-rail-ad.test.js
```diff
@@ -1,127 +1,20 @@
...
-			helpers.closeNewTabs();
-		});
-
-		it('will test if rail scrolls with the content', () => {
-			helpers.slowScroll(500);
-			expect(browser.element(floatingRailAd.rail).getAttribute(helpers.classProperty))
-				.to
-				.equal(floatingRailAd.attributeRailScrolling, 'Rail did not scroll');
-		});
+	it('Check if rail scrolls with the content', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: When you scroll, rail's class changes from `rail-static` to `rail-fixed`. This is the only change I can see in the code. If I'm missing something here, I need some guidance.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+const { expect } = require('chai');
+
+describe('HiVi UAP JWP ads page: top leaderboard', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		browser.waitForVisible(hiviUapJwp.loadAdsButton, timeouts.standard);
+	});
+
+	it('Check if top leaderboard is not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.topLeaderboard}${adSlots.resultAttribute}`), 'Top leaderboard visible')
```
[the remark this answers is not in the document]
ThisIsJesus: `Check if top leaderboard is existing, but not immediately visible` then? It's hardly any better, I know, but this is the best I can think of.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+});
+
+describe('HiVi UAP JWP ads page: top boxad', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		browser.waitForVisible(hiviUapJwp.loadAdsButton, timeouts.standard);
+	});
+
+	it('Check if top boxad is not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.topBoxad}${adSlots.resultAttribute}`), 'Top boxad visible')
```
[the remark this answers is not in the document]
ThisIsJesus: Ditto.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+			.to
+			.be
+			.false;
+	});
+
+	it('Check if the ad loaded after delay is visible and if it is the inhouse one', () => {
+		hiviUapJwp.waitToLoadAds();
+		browser.waitForVisible(adSlots.topBoxad, timeouts.standard);
```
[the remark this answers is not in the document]
ThisIsJesus: I got rid of this line and left isVisibleWithinViewport, as I think it's more important. I also added a missing error message.

tests/specs/hivi-uap-jwp-ad.test.js
```diff
@@ -0,0 +1,125 @@
...
+			.be
+			.true;
+		expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.lineItemIdAttribute))
+			.to
+			.equal(hiviUapJwp.inHouseLineItemId, 'Wrong ad loaded');
+	});
+
+	it('Check if top boxad shows up after clicking the button, if it was viewed and if it is uap ad', () => {
+		browser.click(hiviUapJwp.loadAdsButton);
+		browser.waitForVisible(adSlots.topBoxad, timeouts.standard);
```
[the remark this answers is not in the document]
ThisIsJesus: Ditto.

tests/specs/hivi-uap-jwp-ad.test.js
ThisIsJesus: Please refer to my comment left earlier, about you saying "it may be existing, but not visible".

tests/specs/hivi-uap-jwp-ad.test.js
ThisIsJesus: I disagree on this one. I've been told that if such case happens, I should rewrite them to one case, else we'd be testing the same thing twice for no reason. @emadry please correct me if I am wrong.

tests/specs/hivi-uap-jwp-ad.test.js
ThisIsJesus: What for? It's still incontent boxad, why would you want a separate describe for it? :)

tests/specs/hivi-uap-twitch-ad.test.js
ThisIsJesus: I'll go through these descriptions again and change it. I may have overdone it when I was tweaking them. ;)

tests/specs/floating-rail-ad.test.js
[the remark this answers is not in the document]
jacekbj: What I meant was that appending a class does not mean the rail really scrolls with the user. The class responsible for sticking it may be broken.

tests/specs/hivi-uap-jwp-ad.test.js
jacekbj: Too many things are checked in this test case. They can be easily split into 3 cases. For the sake of readability one can group them inside a describe with `before()` doing the click, scrolling and waits.

tests/common/helpers.js
```diff
@@ -80,19 +84,197 @@ class Helpers {
...
+	waitForResult(adSlot, result) {
+		browser.waitUntil(() => browser.element(adSlot).getAttribute(adSlots.resultAttribute) === result, timeouts.standard, `Result mismatch: expected ${result}`, timeouts.interval);
+	}
+
+	/**
+	 * Checks the slot\'s dimensions. Returns result and (if present) error messages.
+	 * @param adSlot slot dimensions are taken from
+	 * @param width slot\'s width
+	 * @param height slot\'s height
+	 * @param customPrefix additional information to add before the error message (e.g. if the size applies to default or resolved state of the slot)
```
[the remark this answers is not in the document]
ThisIsJesus: I think we can get rid of it, but before that I want to double-check if this custom prefix is not needed anywhere. If it is, I'll change the test name so it's no longer necessary.

tests/specs/block-btf-ad.test.js
```diff
@@ -5,56 +5,20 @@ import helpers from '../common/helpers';
...
-		const tabIds = browser.getTabIds();
-
-		browser.switchTab(tabIds[1]);
-		helpers.waitForUrl(helpers.fandomWord);
-		expect(browser.getUrl())
-			.to
-			.include(helpers.fandomWord);
-		helpers.closeNewTabs();
+	beforeEach(() => {
+		helpers.slowScroll(2000);
```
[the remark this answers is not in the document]
ThisIsJesus: Done. :)

tests/specs/abcd-ad.test.js
```diff
@@ -5,144 +5,78 @@ import helpers from '../common/helpers';
...
-			}
-			try {
-				expect(browser.isVisibleWithinViewport(adSlots.topBoxad), 'Top boxad not in viewport')
-					.to
-					.be
-					.true;
-			} catch (error) {
-				tableOfErrors.push(error.message);
-			}
+	it('Check if leaderboard does not obstruct the navbar', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: Done. :)

tests/specs/common-slots-ad.test.js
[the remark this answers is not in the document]
ThisIsJesus: I agree with @jacekbj here, aside from the last test, there doesn't seem to be any interaction with the slot, I'll move it to before().

tests/specs/floating-rail-ad.test.js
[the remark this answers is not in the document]
ThisIsJesus: I realize that, but, as I said, there is really nothing else I can check, not to my current knowledge. :( I will need @emadry to help me with this one.

tests/common/helpers.js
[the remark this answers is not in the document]
ThisIsJesus: Ok, I got rid of it and updated tests that were using it. :)

tests/specs/empty-response.test.js
```diff
@@ -5,30 +5,36 @@ import helpers from '../common/helpers';
...
+});
+
+describe('Empty response page: top boxad', () => {
+	let adStatus;
+
+	before(() => {
+		browser.url(emptyResponse.pageLink);
+		browser.waitForVisible(emptyResponse.article, timeouts.standard);
+		helpers.waitForResult(adSlots.topBoxad, adSlots.adCollapsed);
+		adStatus = helpers.checkSlotStatus(adSlots.topBoxad);
```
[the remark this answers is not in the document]
ThisIsJesus: There is no beforeEach here, I don't see the reason to change anything like I did in previous cases.

tests/specs/empty-response.test.js
```diff
@@ -5,30 +5,36 @@ import helpers from '../common/helpers';
...
-describe('It will test empty response page', () => {
-	beforeEach(() => {
+describe('Empty response page: top leaderboard', () => {
+	let adStatus;
+
+	before(() => {
 		browser.url(emptyResponse.pageLink);
-		browser.waitForVisible(emptyResponse.articleClass, timeouts.standard);
+		browser.waitForVisible(emptyResponse.article, timeouts.standard);
+		adStatus = helpers.checkSlotStatus(adSlots.topLeaderboard);
```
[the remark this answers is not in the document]
ThisIsJesus: There is no beforeEach here, I don't see the reason to change anything like I did in previous cases.

tests/specs/hivi-uap-jwp-ad.test.js
ThisIsJesus: Done.

tests/specs/hivi-uap-jwp-ad.test.js
ThisIsJesus: Done.

tests/common/helpers.js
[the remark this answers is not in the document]
ThisIsJesus: Done.

package.json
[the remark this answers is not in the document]
ThisIsJesus: Removed npm, left lodash.

PR #195, opened 2018-10-12
ADEN-7773 Fix wikia adapter after moving prebid to ae3

nandy-andy:


spec/ad-engine/video/vast-url-builder.spec.js
```diff
@@ -12,7 +12,10 @@ describe('vast-url-builder', () => {
 				adUnitId: '/5441/wka.fandom/{src}/{slotConfig.slotName}'
 			},
 			slots: {
-				top_leaderboard: {}
+				top_leaderboard: {
+					src: 'test',
```
fraszczakszymon: Is this needed? It should be taken from global scope of context (L10)

src/ad-engine/video/vast-url-builder.js
```diff
@@ -17,24 +20,19 @@ function getCustomParameters(slot, extraTargeting = {}) {
...
+		'gdfp_req=1',
+		'impl=s',
+		'unviewed_position_start=1',
+		'sz=640x480',
+		`url=${encodeURIComponent(window.location.href)}`,
+		`description_url=${encodeURIComponent(window.location.href)}`,
+		`correlator=${correlator}`
+	];
+
+	params.push(`iu=${getVideoAdUnit(slotName)}`);
```
fraszczakszymon: It can be added directly in params array (L22)

nandy-andy: Exactly! @jacekbj, look: great minds think alike ;)

PR #202, opened 2018-10-22
ADEN-7842 Support out-of-page slots and add floor-adhesion/interstital templates

fraszczakszymon:


tests/specs/interstitial.test.js
```diff
@@ -0,0 +1,42 @@
...
+		expect(browser.element('html').getAttribute('class'))
+			.to
+			.include('stop-scrolling', 'Document does not have .stop-scrolling class');
+
+		browser.click(interstitial.closeButton);
+		expect(browser.isVisibleWithinViewport(adSlots.invisibleHighImpact, 'Floor adhesion is in the viewport'))
+			.to
+			.be
+			.false;
+		expect(browser.element('html').getAttribute('class'))
```
emadry: You can wait for element to not have class
`browser.waitForExisting('.element.stop-scrolling', timeouts.standard, true);`
The 3rd parameter, boolean is reversing the logic of the wait, so it waits for element to disappear

PR #203, opened 2018-10-23
ADEN-7783 Possibility to configure timeout and randomness for wikia adapter

fraszczakszymon:


tests/common/asserts.js
```diff
@@ -0,0 +1,23 @@
...
+
+class Asserts {
+	assertSlotLineItemId(slotName, expectedLineItemId) {
+		helpers.waitForLineItemIdAttribute(slotName);
+		expect(helpers.getLineItemId(slotName))
+			.to
+			.equal(expectedLineItemId, 'Line item ID mismatch');
+	}
+
+	assertInhouseCreative(slotName) {
```
ThisIsJesus: I'm uncertain about this method's name - the first thing I thought of after seeing it was checking creative ID instead of line item ID. Can we think of something better?

tests/common/asserts.js
```diff
@@ -0,0 +1,23 @@
...
+		this.assertSlotLineItemId(slotName, adSlots.inhouseLineItemId);
+	}
+
+	assertWikiaAdapterCreative(slotName) {
```
ThisIsJesus: Ditto.

tests/specs/prebid-wikia-adapter.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+		asserts.assertWikiaAdapterCreative(adSlots.topLeaderboard);
+	});
+
+	it('Check if wikia adapter creative is rendered for lazy loaded slot when prebid timeouts at the beginning', () => {
+		browser.url(reusablePrebid.getLinkWithWikiaAdapterParameters(1410, 0, true, 3000));
+
+		asserts.assertInhouseCreative(adSlots.topLeaderboard);
+
+		browser.pause(4000);
```
ThisIsJesus: OH NOES! MAGIC NUMBER DETECTED! Move it to Timeouts please.

tests/specs/prebid-wikia-adapter.test.js
```diff
@@ -0,0 +1,41 @@
...
+		helpers.slowScroll(5000);
```
ThisIsJesus: Unless you need the page to scroll slowly (for whatever reason) I suggest using browser.scroll() instead. ;)

tests/specs/prebid-wikia-adapter.test.js
```diff
@@ -0,0 +1,41 @@
...
+		asserts.assertWikiaAdapterCreative(repeatableSlots.getRepeatableSlot(2));
+	});
+
+	it('Check if wikia adapter creative is not rendered when limit of bids is reached', () => {
+		browser.url(reusablePrebid.getLinkWithWikiaAdapterParameters(1410, 2, true));
+
+		asserts.assertWikiaAdapterCreative(adSlots.topLeaderboard);
+		asserts.assertWikiaAdapterCreative(repeatableSlots.getRepeatableSlot(1));
+
+		helpers.slowScroll(5000);
```
ThisIsJesus: Ditto.

tests/specs/prebid-wikia-adapter.test.js
```diff
@@ -0,0 +1,41 @@
+import asserts from '../common/asserts';
+import adSlots from '../common/adSlots';
+import helpers from '../common/helpers';
+import repeatableSlots from '../pages/repeatable-slots.page';
+import reusablePrebid from '../pages/reusable-prebid.page';
+
+describe('Reusable prebid', () => {
+	it('Check if wikia adapter creative is not rendered when there are no bids', () => {
+		browser.url(reusablePrebid.getLinkWithWikiaAdapterParameters());
+
```
ThisIsJesus: Comment below applies to all the cases you wrote there.

So far there was no empty line between instructions and assertion, but at the same time there was no convention about this discussed. I would like @emadry to also make an opinion about this.

tests/common/asserts.js
fraszczakszymon: Makes sense! Thanks, I'm going to change it.

PR #211, opened 2018-11-05
ADEN-7850 Visual regression for wdio tests, UAP Mobile tests, UAP Sticky BFAB Desktop

ThisIsJesus:


tests/common/helpers.js
```diff
@@ -293,6 +295,13 @@ class Helpers {
 
 		browser.frame(frame);
 	}
+
+	/**
+	 * Waits for the slot to collapse.
+	 */
+	waitToCollapse() {
+		browser.pause(timeToCollapse);
```
emadry: you can pass slot here, and check it's size - so wait until size is 0

tests/pages/porvata.page.js
```diff
@@ -32,6 +20,18 @@ class Porvata {
 	waitForVideoToFinish() {
 		browser.pause(this.videoLength);
 	}
+
+	/**
+	 * Returns autoplay on or off
+	 * @param {boolean} autoplay - true/false
```
emadry: boolean means true/false. Write something like: "turns autoplay on/off"

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,180 @@
...
+		expect(browser.isExisting(`${adSlots.topLeaderboard}${adSlots.resultAttribute}`),
+			'Top leaderboard visible')
+			.to
+			.be
+			.false;
+	});
+
+	it('Check if top leaderboard does not load after manually finishing the queue', () => {
+		browser.click(hiviUapJwp.loadAdsButton);
+		expect(browser.isExisting(adSlots.lineItemIdAttribute), 'Top leaderboard has been loaded')
```
emadry: it can be flaky. You are checking it instantly after click

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,180 @@
...
+			.to
+			.equal(hiviUapJwp.inHouseLineItemId, 'Wrong ad loaded');
+	});
+});
+
+describe('Mobile HiVi UAP JWP ads page: top boxad (ads loaded after clicking the button)', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		browser.waitForVisible(hiviUapJwp.loadAdsButton, timeouts.standard);
+		browser.waitUntil(
```
emadry: Move function from this test to the page.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,180 @@
...
+	});
+});
+
+describe('Mobile HiVi UAP JWP ads page: incontent boxad (ads loaded after 10s)', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		hiviUapJwp.waitToLoadAds();
+		browser.scroll(adSlots.railModule);
+		browser.waitForVisible(adSlots.incontentBoxad, timeouts.standard);
+		browser.scroll(adSlots.incontentBoxad); // separate scroll, because this slot is not immediately visible
```
emadry: This scrolling and waiting can be moved to page logic. Also why not slow scroll?

package.json
```diff
@@ -92,8 +94,10 @@
...
     "test-coverage": "nyc mocha-webpack './spec/**/*.spec.js' -r jsdom-global/register --webpack-env.test --reporter mocha-junit-reporter --reporter-options mochaFile=./spec/build/tests.xml",
     "watch": "./node_modules/webpack/bin/webpack.js --progress --colors --hide-modules --watch",
-    "wdio": "BABEL_ENV=test ./node_modules/.bin/wdio  wdio.conf.js",
-    "allure": "cp -R ./tests/allure-results/report/history ./tests/allure-results | true && allure generate ./tests/allure-results --output ./tests/allure-results/report --clean && allure open ./tests/allure-results/report",
+    "wdioDesktop": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.desktop.conf.js",
+    "wdioMobile": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.mobile.conf.js",
+    "wdioAll": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.mobile.conf.js; BABEL_ENV=test ./node_modules/.bin/wdio wdio.desktop.conf.js",
+    "allure": "cp -R ./tests/allure-results/report/history ./tests/allure-results | true; allure generate ./tests/allure-results --output ./tests/allure-results/report --clean; allure open ./tests/allure-results/report",
```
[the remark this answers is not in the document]
ThisIsJesus: Done.

tests/pages/porvata.page.js
ThisIsJesus: Done.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
ThisIsJesus: There is no need for a slow scroll here, we can scroll directly to the desired position. I use slow scroll only when something loads while scrolling to a certain point, when it's not immediately visible. As for the rest, I'll move it now.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
ThisIsJesus: There was literally nothing I could wait for in top leaderboard, so I created a special method for the pause. Both this method and the comment left next to it in test case explains the necessity to use it.

I would gladly accept some tips how to do it differently, though.

tests/pages/hivi-uap-ad.page.js
```diff
@@ -15,7 +15,7 @@ class HiviUap {
 		this.slotCollapsed = 'collapse';
 		this.topLineItemId = '4466763538'; // top leaderboard and top boxad share the same ID
 		this.bottomLineItemId = '4511050296'; // bottom leaderboard and incontent boxad share the same ID
-		this.videoLength = 45000;
+		this.videoLength = 50000; // 5 seconds more than the movie has to make sure it changes dimensions (necessary fix)
```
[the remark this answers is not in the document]
ThisIsJesus: Done (with your help).

tests/common/helpers.js
ThisIsJesus: Done. Slightly modified old waitForExpanded() method, as it uses almost the identical logic.

package.json
```diff
@@ -92,8 +94,10 @@
     "test": "mocha-webpack --full-trace './spec/**/*.spec.js' -r jsdom-global/register --webpack-env.test",
     "test-coverage": "nyc mocha-webpack './spec/**/*.spec.js' -r jsdom-global/register --webpack-env.test --reporter mocha-junit-reporter --reporter-options mochaFile=./spec/build/tests.xml",
     "watch": "./node_modules/webpack/bin/webpack.js --progress --colors --hide-modules --watch",
-    "wdio": "BABEL_ENV=test ./node_modules/.bin/wdio  wdio.conf.js",
+    "wdioDesktop": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.desktop.conf.js",
```
fraszczakszymon: Make it consistent with other script names please (camel case vs dash)

.eslintignore
```diff
@@ -1,3 +1,4 @@
+allure-report/
```
fraszczakszymon: Not needed anymore. It's handled in L7

wdio.mobile.conf.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const merge = require('deepmerge');
+const wdioConf = require('./wdio.conf.js');
+
+exports.config = merge(wdioConf.config, {
```
fraszczakszymon: Not sure if we need separate deepmerge dependency here. What about `Object.assign({}, wdioConf.config, { specs: ... }`

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
ThisIsJesus: I was shown another way to solve this problem (thank you!) and used that solution. Done.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
ThisIsJesus: Moved. :)

package.json
ThisIsJesus: Will do, thanks for pointing this out! :)

wdio.mobile.conf.js
ThisIsJesus: It's currently based on examples provided [here](http://webdriver.io/guide/testrunner/organizesuite.html), but I'll take a look at your idea as well.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+describe('Mobile HiVi UAP JWP ads page: incontent boxad (ads loaded after 10s)', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+	});
+	afterEach(() => {
+		browser.scroll(0, 0);
+	});
+
+	it('Check if incontent boxad is existing, but not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.incontentBoxad}${adSlots.resultAttribute}`),
```
jacekbj: ditto if valid

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+		hiviUapJwp.waitForAdsAfterDelayAndScrollToAdSlotOnMobile(adSlots.railModule, adSlots.incontentBoxad);
+		browser.waitForVisible(adSlots.incontentBoxad, timeouts.standard);
+		browser.scroll(adSlots.incontentBoxad); // separate scroll, because this slot is not immediately visible
+		helpers.waitForLineItemIdAttribute(adSlots.incontentBoxad);
+		expect(browser.isVisibleWithinViewport(adSlots.incontentBoxad))
+			.to
+			.be
+			.true;
+		expect(browser.element(adSlots.incontentBoxad)
+			.getAttribute(adSlots.lineItemIdAttribute))
```
jacekbj: we do it so often we could have a helper for that

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+			.be
+			.true;
+	});
+
+	it('Check if slot was viewed', () => {
+		helpers.waitForViewed(adSlots.incontentBoxad);
+		expect(browser.element(adSlots.incontentBoxad).getAttribute(adSlots.resultAttribute))
+			.to
+			.equal(adSlots.adLoaded, 'Incontent boxad slot failed to load');
+		expect(browser.element(adSlots.incontentBoxad).getAttribute(adSlots.viewedAttribute))
```
jacekbj: ditto

tests/specs/templates/hivi-uap-static-ad.desktop.test.js
```diff
@@ -93,9 +92,19 @@ describe('HiVi UAP static ads page: top leaderboard', () => {
...
 			.equal(hiviUapStatic.slotCollapsed, 'Top leaderboard has not been closed');
 	});
+
+	it('Check visual regression in top leaderboard (default)', () => {
+		helpers.reloadPageAndWaitForSlot(hiviUapStatic.pageLink, adSlots.topLeaderboard);
+		browser.checkElement(adSlots.topLeaderboard);
+	});
+
+	it('Check visual regression in top leaderboard (resolved)', () => {
+		helpers.refreshPageAndWaitForSlot(adSlots.topLeaderboard, timeouts.standard);
```
jacekbj: I think time required for resolving should have it's own definition instead of using `timeouts.standard`.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
jacekbj: Oh we do have! `getLineItemId` :D

tests/specs/templates/hivi-uap-static-ad.mobile.test.js
```diff
@@ -0,0 +1,130 @@
...
+	});
+
+	it('Check if navbar is visible in viewport', () => {
+		expect(browser.isVisibleWithinViewport(helpers.navbar), 'Navbar not visible')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if redirect on tap works', () => {
```
jacekbj: d

tests/specs/templates/hivi-uap-static-ad.mobile.test.js
```diff
@@ -0,0 +1,130 @@
...
+		expect(helpers.adRedirect(adSlots.topLeaderboard), 'Wrong link after redirect')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if closing top leaderboard works', () => {
```
jacekbj: d

tests/specs/templates/hivi-uap-static-ad.mobile.test.js
```diff
@@ -0,0 +1,130 @@
...
+		browser.waitForVisible(adSlots.topBoxad, timeouts.standard);
+	});
+
+	afterEach(() => {
+		browser.scroll(0, 0);
+	});
+
+	it('Check if line item id is from the same campaign', () => {
+		helpers.waitForLineItemIdAttribute(adSlots.topBoxad);
+		expect(helpers.getLineItemId(adSlots.topBoxad))
```
jacekbj: This looks like a perfect candidate for custom assertion @emadry

wdio.conf.js
```diff
@@ -44,6 +46,18 @@ exports.config = {
...
+
+	visualRegression: {
+		compare: new VisualRegressionCompare.LocalCompare({
+			referenceName: getScreenshotName(path.join(process.cwd(), 'tests/screenshots/reference')),
+			screenshotName: getScreenshotName(path.join(process.cwd(), 'tests/screenshots/current')),
+			diffName: getScreenshotName(path.join(process.cwd(), 'tests/screenshots/diff')),
+			misMatchTolerance: 0.01,
+		}),
+		viewportChangePause: 300,
+		viewports: [{ width: 1920, height: 1080 }],
```
jacekbj: I have offline questions regarding viewports and regression for mobile

package.json
```diff
@@ -92,8 +94,10 @@
     "test": "mocha-webpack --full-trace './spec/**/*.spec.js' -r jsdom-global/register --webpack-env.test",
     "test-coverage": "nyc mocha-webpack './spec/**/*.spec.js' -r jsdom-global/register --webpack-env.test --reporter mocha-junit-reporter --reporter-options mochaFile=./spec/build/tests.xml",
     "watch": "./node_modules/webpack/bin/webpack.js --progress --colors --hide-modules --watch",
-    "wdio": "BABEL_ENV=test ./node_modules/.bin/wdio  wdio.conf.js",
+    "wdio-desktop": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.desktop.conf.js",
+    "wdio-mobile": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.mobile.conf.js",
+    "wdio-all": "BABEL_ENV=test ./node_modules/.bin/wdio wdio.mobile.conf.js; BABEL_ENV=test ./node_modules/.bin/wdio wdio.desktop.conf.js",
```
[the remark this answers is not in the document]
ThisIsJesus: It does work, though not only slightly slower (there is a 1 - 2 second break between runs), but also the results are shown as two separate runs instead of one. I think we should leave it like it is for readibility's sake.

tests/common/helpers.js
```diff
@@ -256,6 +256,7 @@ class Helpers {
 		this.waitForLineItemIdAttribute(adSlot);
 		browser.waitForEnabled(adSlot, timeouts.standard);
 		browser.click(adSlot);
+		browser.pause(timeouts.standard); // TODO remove this workaround after chromedriver update for opening new pages
```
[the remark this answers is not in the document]
ThisIsJesus: ...and not on this branch. ;)

tests/pages/hivi-uap-jwp-ad.page.js
```diff
@@ -25,6 +27,55 @@ class HiviUapJwp {
 			'Ads not loaded',
 			timeouts.interval);
 	}
+
+	/**
+	 * Waits for the delay to pass so ads can load and scrolls to the desired ad slot.
+	 * This method is desired for mobile tests.
+	 * @param adSlot first ad slot/module to scroll to
+	 */
+	waitForAdsAfterDelayAndScrollToAdSlotOnMobile(adSlot) {
```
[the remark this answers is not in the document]
ThisIsJesus: This (and AfterClick) method vary slightly, whether it's used in desktop or mobile. I couldn't find a way to make it any more universal than this, hence the name.

tests/pages/hivi-uap-jwp-ad.page.js
```diff
@@ -25,6 +27,55 @@ class HiviUapJwp {
...
+		this.waitToLoadAds();
+		browser.scroll(adSlot);
+	}
+
+	/**
+	 *Waits for the delay to pass so ads can load and scrolls to the desired slot.
+	 * This method is desired for desktop tests.
+	 * @param adSlot slot we want to be visible
+	 */
+	waitForAdsAfterDelayAndScrollToAdSlotOnDesktop(adSlot) {
```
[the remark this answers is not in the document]
ThisIsJesus: As said before, there is a slight difference between these two. I may take one more look at these two and try to do something with them.

tests/pages/hivi-uap-ad.page.js
```diff
@@ -24,6 +27,25 @@ class HiviUap {
 	waitForVideoToFinish() {
 		browser.pause(this.videoLength);
 	}
+
+	/**
+	 * Returns current slot size if it equals the one we desire based on ratio.
```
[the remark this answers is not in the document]
ThisIsJesus: "If slot size ratio" what? Exists? Checks something? There is obviously something missing from this sentence, please clarify.

tests/pages/hivi-uap-jwp-ad.page.js
```diff
@@ -25,6 +27,55 @@ class HiviUapJwp {
...
+
+	/**
+	 * Clicks on the button to load ads. After the click, scrolls to the desired slot.
+	 * This method is desired for desktop tests.
+	 * @param adSlot slot we want to be visible
+	 */
+	waitForAdsAfterClickAndScrollToAdSlotOnDesktop(adSlot) {
+		browser.waitForVisible(this.loadAdsButton);
+		browser.click(this.loadAdsButton);
+		helpers.slowScroll(1000);
```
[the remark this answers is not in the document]
ThisIsJesus: Now it should! :)

tests/pages/porvata.page.js
```diff
@@ -32,6 +20,18 @@ class Porvata {
 	waitForVideoToFinish() {
 		browser.pause(this.videoLength);
 	}
+
+	/**
+	 * Turns autoplay on/off
+	 * @param {boolean} autoplay - true/false
```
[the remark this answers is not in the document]
ThisIsJesus: Then what would you like to add to it? What am I missing?

tests/pages/hivi-uap-ad.page.js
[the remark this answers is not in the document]
jacekbj: What I meant was "It checks slot size ratio." :D

tests/pages/porvata.page.js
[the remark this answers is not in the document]
jacekbj: Emil mentioned that there is no need to write '- true/false' - it's defined as boolean type.

tests/specs/templates/hivi-uap-ad.mobile.test.js
```diff
@@ -0,0 +1,174 @@
...
+	});
+
+	it('Check if redirect on tap works', () => {
+		expect(helpers.adRedirect(adSlots.topLeaderboard), 'Wrong link after redirect')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if closing top leaderboard works properly', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: Slightly modified, should look good now.

tests/specs/templates/hivi-uap-ad.mobile.test.js
```diff
@@ -0,0 +1,174 @@
...
+	});
+
+	it('Check if navbar is visible in viewport', () => {
+		expect(browser.isVisibleWithinViewport(helpers.navbar), 'Navbar not visible')
+			.to
+			.be
+			.true;
+	});
+
+	it('Check if redirect on tap works', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: Changed, but I certainly hope I won't have to revert it.

tests/specs/templates/hivi-uap-ad.mobile.test.js
```diff
@@ -0,0 +1,174 @@
...
+		browser.waitForVisible(adSlots.topLeaderboard);
+		helpers.waitToStartPlaying();
+		browser.click(`${adSlots.topLeaderboard} ${hiviUap.videoPlayer}`);
+	});
+
+	it('Check if opening the full screen player works properly', () => {
+		browser.waitForEnabled(`${adSlots.topLeaderboard} ${hiviUap.playerFullscreenButton}`,
+			timeouts.standard);
+		browser.click(`${adSlots.topLeaderboard} ${hiviUap.playerFullscreenButton}`);
+		browser.waitForExist(hiviUap.fullScreen, timeouts.standard);
```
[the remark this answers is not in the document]
ThisIsJesus: Because we need to wait for the certain class to exis in DOM, waitForVisible wouldn't work here.

tests/specs/templates/hivi-uap-ad.mobile.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+describe('Mobile HiVi UAP ads page: video player in top leaderboard', () => {
+	beforeEach(() => {
+		browser.url(hiviUap.pageLink);
+		browser.waitForVisible(adSlots.topLeaderboard);
+		helpers.waitToStartPlaying();
+		browser.click(`${adSlots.topLeaderboard} ${hiviUap.videoPlayer}`);
+	});
+
+	it('Check if opening the full screen player works properly', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: fixed fixed fixed fixed

tests/common/helpers.js
```diff
@@ -293,6 +294,18 @@ class Helpers {
 
 		browser.frame(frame);
 	}
+
+	/**
+	 * Waits until the slot has collapsed (its size equals 0).
+	 * @param adSlot ad slot we are waiting for
+	 */
+	waitToCollapse(adSlot) {
```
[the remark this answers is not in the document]
ThisIsJesus: Good point, I forgot we can use waitForVisible with reverse flag to check quite the opposite thing. If that proves to be correct, then this method will be completely useless here.

wdio.conf.js
ThisIsJesus: When I was working on both of these, I spoke to @emadry - he told me that we're only gonna use FHD resolution for desktop/visual regression. There will be no VR for mobile (though I believe it'll be subject to change for quite some time), the resolution for mobile was picked by me, but I have nothing against changing it later.

tests/specs/templates/hivi-uap-static-ad.desktop.test.js
ThisIsJesus: Why? 5 seconds is quite ok for a timeout and there is no point in creating yet another variable with the same value. Still, I would like to hear your proposition, what do you have in your mind?

tests/pages/hivi-uap-ad.page.js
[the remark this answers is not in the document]
ThisIsJesus: Done.

tests/pages/porvata.page.js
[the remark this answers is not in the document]
ThisIsJesus: We have two choices here:
- leave it like it is
- removed it from the documentation completely (else eslint will throw an error)

What should it be?

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+const { expect } = require('chai');
+
+describe('Mobile HiVi UAP JWP ads page: top leaderboard', () => {
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+		browser.waitForVisible(hiviUapJwp.loadAdsButton, timeouts.standard);
+	});
+
+	it('Check if top leaderboard is existing, but not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.topLeaderboard}${adSlots.resultAttribute}`),
```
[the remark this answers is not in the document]
ThisIsJesus: Nowhere. :( Added missing lines both in desktop and mobile tests (decided to go with simple waitForVisible, no need for assertion). Also, fixed its description so it better reflects what happens in this test.

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+	beforeEach(() => {
+		browser.url(hiviUapJwp.pageLink);
+	});
+
+	afterEach(() => {
+		browser.scroll(0, 0);
+	});
+
+	it('Check if top boxad is existing, but not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.topBoxad}${adSlots.resultAttribute}`), 'Top boxad visible')
```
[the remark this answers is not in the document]
ThisIsJesus: Ditto (because previous remark regarding visibility was valid)

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+			.be
+			.true;
+	});
+
+	it('Check if slot was viewed', () => {
+		helpers.waitForViewed(adSlots.topBoxad);
+		expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.resultAttribute))
+			.to
+			.equal(adSlots.adLoaded, 'Top boxad slot failed to load');
+		expect(browser.element(adSlots.topBoxad).getAttribute(adSlots.viewedAttribute))
```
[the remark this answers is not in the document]
ThisIsJesus: That's exactly why I removed unnecessary lines after reading this comment. :)

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
ThisIsJesus: Ditto, because valid.

tests/specs/templates/hivi-uap-static-ad.desktop.test.js
jacekbj: Even if for now it's the same value, we may want to change the value of "standard" timeout (since it's not tightly connected to any feature). However, resolving of UAP is a well-defined value, hence I think it should be defined in a separate variable.

wdio.conf.js
jacekbj: My question is: since we have separate files for desktop and mobile, why do we keep desktop resolution in the common file?

tests/pages/porvata.page.js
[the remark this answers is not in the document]
ThisIsJesus: Solved - simply removed true/false and just left "autoplay".

tests/specs/templates/hivi-uap-jwp-ad.mobile.test.js
```diff
@@ -0,0 +1,164 @@
...
+	});
+
+	it('Check if top boxad is existing, but not immediately visible', () => {
+		expect(browser.isExisting(`${adSlots.topBoxad}${adSlots.resultAttribute}`), 'Top boxad visible')
+			.to
+			.be
+			.false;
+	});
+
+	it('Check if the ad loaded after delay is visible and if it is the inhouse one', () => {
```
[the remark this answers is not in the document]
ThisIsJesus: Done! :)

tests/pages/hivi-uap-jwp-ad.page.js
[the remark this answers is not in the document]
ThisIsJesus: Agreed with @jacekbj that we will not work on it now, if we decide to do it, we are gonna do it later.

tests/pages/hivi-uap-jwp-ad.page.js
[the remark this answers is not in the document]
ThisIsJesus: I did take a look - while I can scroll directly to the slot on mobile (because it's already visible), I can't always do it on desktop (which is why these are different). I may refactor them, but definitely not now, as stated before.

tests/pages/hivi-uap-jwp-ad.page.js
```diff
@@ -25,6 +27,55 @@ class HiviUapJwp {
...
+
+	/**
+	 * Clicks on the button to load ads. After the click, scrolls to the desired ad slot.
+	 * This method is desired for mobile tests.
+	 * @param adSlot ad slot to scroll to
+	 */
+	waitForAdsAfterClickAndScrollToAdSlotOnMobile(adSlot) {
+		browser.waitForVisible(this.loadAdsButton, timeouts.standard);
+		browser.waitUntil(
+			() => browser.getText(this.loadAdsButton) === 'Load UAP:JWP (7s)',
```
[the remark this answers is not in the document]
ThisIsJesus: There is a slight difference in scroll. Just like with the previous ones, we agreed to leave it like this for now and maybe work on it a little bit later.

As for the missing wait for text - added it now, because it should be there indeed.

wdio.mobile.conf.js
ThisIsJesus: Actually, we do need deepmerge here. I tested your solution with @jacekbj and we stumbled upon some problems. I believe he will explain everything

jacekbj: deepmerge is more convenient when the object has many levels - it can for example merge arrays (so cool, check out first example: https://www.npmjs.com/package/deepmerge)

tests/pages/hivi-uap-jwp-ad.page.js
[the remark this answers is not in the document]
emadry: It still just scrolls through the page but after that slot might not be in the viewport. Slow scroll does trigger the slot, but you have to scroll to it after

tests/pages/porvata.page.js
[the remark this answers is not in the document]
emadry: Unless there's problem with github you either not pushed it or you haven't changed it
`	/**
	 * Turns autoplay on/off
	 * @param {boolean} autoplay - true/false
	 * @returns {string} link with correct parameter
	 */`

This is still there. And from two choices you've mentioned I would like to pick third. Change it.
I even gave you the example.
`
	 * Turns autoplay on/off - change to: provides query param with autoplay
 	 * @param {boolean} autoplay - true/false - change to: sets autoplay to 1 if true
`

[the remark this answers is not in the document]
ThisIsJesus: Done, though I had to leave autoplay too, else eslint would moan about parameter missing.

PR #213, opened 2018-11-09
ADEN-7809 Unstick TLB on window resize

pnather:


src/ad-products/templates/uap/themes/hivi/stickiness.js
```diff
@@ -73,6 +75,15 @@ export class Stickiness extends EventEmitter {
 		}
 	}
 
+	revertStickinessOnResize() {
+		if (this.unstickOnResize) {
+			window.addEventListener('resize', () => {
```
jacekbj: Is there a possible scenario where this listener is still live after changing page in SPA? I'd add `once: true` option (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) to make 100% it does not mess things up.

PR #216, opened 2018-11-14
ADEN-7818 Move jwplayer ads video module to ad-products

fraszczakszymon:


examples/video/jwplayer/script.js
```diff
@@ -0,0 +1,70 @@
...
+context.set('custom.device', utils.client.getDeviceType());
+context.set('custom.adLayout', 'article');
+context.set('options.tracking.kikimora.player', true);
+context.set('options.video.isMidrollEnabled', utils.queryString.get('midroll') === '1');
+context.set('options.video.isPostrollEnabled', utils.queryString.get('postroll') === '1');
+context.set('options.video.adsOnNextVideoFrequency', parseInt(utils.queryString.get('capping'), 10) || 3);
+
+if (f15sVideoId) {
+	const map = {};
+	map[f15sVideoId] = 5.0;
```
dmnsobczak: This variable is unnecessary, but it's just a example page so it's doesn't matter I guess ¯\_(ツ)_/¯

src/ad-engine/services/slot-service.js
```diff
@@ -41,6 +41,10 @@ function setState(slotName, state, status = null) {
 }
 
 class SlotService {
+	/**
+	 * Add new slot to register
+	 * @param {AdSlot} adSlot
```
dmnsobczak: 💎

src/ad-products/tracking/video/twitch-tracker.js
```diff
@@ -0,0 +1,43 @@
+import { context } from '@wikia/ad-engine';
+import playerEventEmitter from './player-event-emitter';
+import videoEventDataProvider from './video-event-data-provider';
+
+/**
+ * Ads tracker for Twitch
+ */
+export class TwitchTracker {
```
dmnsobczak: Twitch tracker is an exact copy of Porvata tracker. Can't we avoid that?

But on the other hand we gonna make a Twitch cleanup so... ;)

PR #217, opened 2018-11-14
ADEN-7809-2	Resize adslot

pnather:


src/ad-products/templates/sticky-ad.js
```diff
@@ -83,6 +87,10 @@ export class StickyAd {
 				scrollListener.removeCallback(this.scrollListener);
 			}
 		});
+
+		window.addEventListener('resize', () => {
```
dmnsobczak: You can simplify this to: `window.addEventListener('resize', this.adjustAdSlot);`

PR #220, opened 2018-11-16
ADEN-7907 StickyAd enhancements.

jacekbj:


src/ad-products/templates/sticky-ad.js
```diff
@@ -103,54 +103,106 @@ export class StickyAd {
...
+		this.config.moveNavbar(this.adSlot.getElement().scrollHeight, SLIDE_OUT_TIME);
+
+		if (document.hidden) {
+			await utils.once(window, 'visibilitychange');
+		}
+
+		return slotTweaker.makeResponsive(this.adSlot);
+	}
+
+	onCloseClicked() {
```
fraszczakszymon: Do we need this "wrapper"?

src/ad-products/templates/sticky-ad.js
```diff
@@ -1,94 +1,94 @@
...
+	static getDefaultConfig() {
+		return {
+			enabled: true,
+			desktopNavbarWrapperSelector: '.wds-global-navigation-wrapper',
+			mobileNavbarWrapperSelector: '.global-navigation-mobile-wrapper',
+			mainContainer: document.body,
+			handleNavbar: false,
+			stickyAdditionalTime: 0,
+			stickyUntilSlotViewed: true,
+			stickinessAllowed: true, // TODO
```
fraszczakszymon: What's left to do? Should we allow stickiness only for defined slot names?

jacekbj: I am removing this property.
The TLB should enable this (or any other) template in each app in order not to store app-dependent slot names in ad-engine.

PR #221, opened 2018-11-19
ADEN-7933 Adding stickyAd events

dmnsobczak:


src/ad-products/templates/sticky-ad.js
```diff
@@ -58,6 +59,7 @@ export class StickyAd {
 			return;
 		}
 
+		this.adSlot.setStatus(StickyAd.SLOT_STICKY_READY_STATE);
```
fraszczakszymon: I have doubts about setting status. It overrides info about success/collapse :( (it also overrides `data-slot-result`. I'd vote for refactoring this part of code and give possibility to separately:
* set status based on received creative
* dispatch single event for ad slot

But I understand that it would require much more time. What do you think?

dmnsobczak: Will do.

src/ad-engine/models/ad-slot.js
```diff
@@ -170,4 +170,10 @@ export class AdSlot extends EventEmitter {
 		slotTweaker.hide(this);
 		this.setStatus(status);
 	}
+
+	trackEvent(event = null) {
```
fraszczakszymon: s/trackEvent/emitEvent
s/event/eventName

What do you think?

dmnsobczak: Done.

src/ad-engine/models/ad-slot.js
```diff
@@ -170,4 +170,10 @@ export class AdSlot extends EventEmitter {
 		slotTweaker.hide(this);
 		this.setStatus(status);
 	}
+
+	emitEvent(eventName = null) {
+		if (eventName !== null) {
```
jacekbj: How about undefined :trollface: ? `if (eventName != null)` would fix that

dmnsobczak:
Major release: sticky/unsticky BFAB events logic is changed and it will cause regression without changes on app, mobile-wiki and f2.

PR #228, opened 2018-11-23
ADEN-7995 Refactor A9

Bielik20:


src/ad-bidders/a9/index.js
```diff
@@ -140,27 +108,65 @@ export class A9 extends BaseBidder {
...
+				this.targetingKeys.push(key);
+			}
+			this.bids[slotName][key] = bidTargeting[key];
+		});
 	}
 
-	getTargetingKeysToReset() {
-		return this.targetingKeys;
+	async callBids(onResponse) {
+		if (this.cmp) {
```
jacekbj: It's assigned in the constructor, can it ever be false?

Bielik20: Tak powinno być:
```suggestion
		if (this.cmp.exists) {
```
Dlatego że `window.__cmp` może być falsy.

src/ad-bidders/a9/index.js
```diff
@@ -14,106 +15,73 @@ export class A9 extends BaseBidder {
...
-			onResponse();
-		});
+	insertScript() {
+		utils.scriptLoader.loadScript('//c.amazon-adsystem.com/aax2/apstag.js', 'text/javascript', true, 'first');
 	}
 
-	configureApstag() {
-		window.apstag = window.apstag || {};
-		window.apstag._Q = window.apstag._Q || [];
+	getApsConfig(consentData) {
```
dmnsobczak: Aps = Apstag? Why not to keep full name?

src/ad-bidders/a9/index.js
```diff
@@ -14,106 +15,73 @@ export class A9 extends BaseBidder {
...
+			videoAdServer: 'DFP',
+			deals: !!this.bidderConfig.dealsEnabled,
+			gdpr: this.getGdprIfEnabled(consentData)
+		};
+	}
 
-		if (typeof window.apstag.init === 'undefined') {
-			window.apstag.init = (...args) => {
-				this.configureApstagCommand('i', args);
+	getGdprIfEnabled(consentData) {
```
dmnsobczak: Maybe not `enabled` because CMP is enabled globally, but `applicable`.

src/ad-bidders/a9/index.js
```diff
@@ -140,27 +108,65 @@ export class A9 extends BaseBidder {
 		return definition;
 	}
 
-	getBestPrice(slotName) {
-		return this.priceMap[slotName] ? { a9: this.priceMap[slotName] } : {};
+	getBidTargetingAndKeys(bid) {
```
dmnsobczak: It's like `getAnimalAndDog`, keys are a part of targeting, so maybe just that?

src/ad-bidders/a9/index.js
```diff
@@ -140,27 +108,65 @@ export class A9 extends BaseBidder {
...
+
+	getTargetingParams(slotName) {
+		return this.bids[slotName] || {};
 	}
 
 	isSupported(slotName) {
 		return !!this.slots[slotName];
 	}
+
+	// DISCUSS: Is it used somewhere?
```
dmnsobczak: It can be used by old or new Prebid Oasis integration. We will finally clean-up that in Q1 after old Prebid cleanup.

dmnsobczak:
Same here. We're not gonna merge this until Q1. Let's use that time to test it and make sure that it won't brake A9 integration.
Also: any idea how to roll this out gradually, not everywhere at the same time?

src/ad-engine/wrappers/apstag.js
```diff
@@ -0,0 +1,56 @@
+import { bindCallback } from '../utils';
+
+export class Apstag {
```
fraszczakszymon: ad-engine package shouldn't be "aware" of apstag. A9 is implemented in ad-products

src/ad-engine/index.js
```diff
@@ -16,4 +16,5 @@ export * from './models';
 export * from './providers';
 export * from './services';
 export * from './video';
+export * from './wrappers';
```
fraszczakszymon: ad-engine package shouldn't be "aware" of apstag. A9 is implemented in ad-products

I have mixed feelings about cmp, not sure if anything from ad-engine uses __cmp

src/ad-bidders/a9/index.js
```diff
@@ -1,8 +1,9 @@
-import { context, slotService, utils } from '@wikia/ad-engine';
+import { context, slotService, utils, apstag, cmp } from '@wikia/ad-engine';
```
fraszczakszymon: alphabetical order please

examples/bidders/a9/script.js
```diff
@@ -1,4 +1,4 @@
-import { AdEngine, context, events, utils } from '@wikia/ad-engine';
+import { AdEngine, context, events, utils, apstag, cmp } from '@wikia/ad-engine';
```
fraszczakszymon: alphabetical order please

src/ad-bidders/a9/index.js
```diff
@@ -14,106 +15,73 @@ export class A9 extends BaseBidder {
 		this.priceMap = {};
 		this.slotNamesMap = {};
 		this.targetingKeys = [];
-		this.timeout = timeout;
+		this.apstag = apstag;
```
fraszczakszymon: Why do we need to keep apstag/cmp property in A9 instance?

src/ad-bidders/a9/index.js
```diff
@@ -14,106 +15,73 @@ export class A9 extends BaseBidder {
...
 
-	configureApstag() {
-		window.apstag = window.apstag || {};
-		window.apstag._Q = window.apstag._Q || [];
+	getApstagConfig(consentData) {
+		return {
+			pubID: this.amazonId,
+			videoAdServer: 'DFP',
+			deals: !!this.bidderConfig.dealsEnabled,
+			gdpr: this.getGdprIfApplicable(consentData)
```
fraszczakszymon:
```
{ pubID: ..., videoAdServer: ..., deals: ... }
```
vs
```
{ pubID: ..., videoAdServer: ..., deals: ..., gdpr: undefined }
```

I'm not sure if it's treated in the same way. Worth to verify before merging.

src/ad-bidders/a9/index.js
```diff
@@ -140,27 +108,63 @@ export class A9 extends BaseBidder {
 		return definition;
 	}
 
-	getBestPrice(slotName) {
-		return this.priceMap[slotName] ? { a9: this.priceMap[slotName] } : {};
+	getBidTargetingWithKeys(bid) {
+		return this.bidderConfig.dealsEnabled
```
fraszczakszymon: Wouldn't 
```
if (foo) { 
  return 1 
} 
return 2 
``` 
be more readable?

src/ad-bidders/a9/index.js
Bielik20: Propose lint rule

src/ad-bidders/a9/index.js
Bielik20: To not refer to it globally. Better practice, you know just by looking at constructor what given class is using.

src/ad-bidders/a9/index.js
Bielik20: It is overriding gdpr to be undefined. It is not the same as delete, but close enough.

src/ad-engine/index.js
Bielik20: But it was anyway:
```
window.apstag.init(apsConfig);
```
```
configureApstag() {
  ...
}
```
etc.
Check commit "ADEN-7995 Refactor A9"

It is the same with cmp:
```
if (window.__cmp)
```

The fact that it is hidden in global/window doesn't mean that it is not "aware" of it. It merely clouds our perception. It is better to have it said explicitly. Everyone can clearly see what are the class dependencies this way.

src/ad-bidders/a9/index.js
Bielik20: Here I think it boils down to personal taste. To me it is very readable and expressive, return this or that. From what I have read it is good to use the ternary operator for assignments, and regular `if` for operations. As we have an assignment in here (we do not "do" anything) I would leave it that way.

src/ad-bidders/a9/index.js
fraszczakszymon: 🤔 I have mixed feeling about this practice. Given class is using also context, slotService and utils.

src/ad-bidders/a9/index.js
fraszczakszymon: Good point. `sort-imports` and `memberSyntaxSortOrder ` option

src/ad-engine/index.js
Bielik20: We could move wrapper somewhere else. What do you propose?

src/ad-bidders/a9/index.js
```diff
@@ -140,27 +108,63 @@ export class A9 extends BaseBidder {
...
+
+	getTargetingParams(slotName) {
+		return this.bids[slotName] || {};
 	}
 
 	isSupported(slotName) {
 		return !!this.slots[slotName];
 	}
+
+	// Should mark external methods, so that we know they cannot be easily renamed
```
fraszczakszymon: ... or treat renaming of any public method as breaking change

src/ad-bidders/a9/index.js
Bielik20: I think it conflicts with "import/order". Moving discussion to slack.

src/ad-engine/index.js
fraszczakszymon: src/ad-products/wrappers

src/ad-bidders/a9/index.js
Bielik20: Changed

src/ad-bidders/a9/index.js
Bielik20: 4:1 you won.

src/ad-bidders/a9/index.js
Bielik20: Of course, still it would be nice to have something like c# "internal" access operator.

src/ad-bidders/wrappers/apstag.js
```diff
@@ -0,0 +1,56 @@
...
+		window.apstag._Q.push([command, args]);
+	}
+
+	/**
+	 * @param {object} bidsConfig configuration of bids
+	 * @param {function(object)} cb Callback receiving current bids
+	 * @returns {!Promise} If `cb` has been omitted
+	 */
+	fetchBids(bidsConfig, cb = null) {
+		return utils.getPromiseAndExecuteCallback((resolve) => {
```
fraszczakszymon: What's the benefit? Why don't we call
```
window.apstag.fetchBids(bidsConfig, cb);
```
Returned promise seems not needed

src/ad-bidders/wrappers/cmp.js
```diff
@@ -0,0 +1,23 @@
+import { utils } from '@wikia/ad-engine';
+
+export class Cmp {
+	get exists() {
+		return !!window.__cmp;
+	}
+
+	/**
+	 * @param {function(object)} cb Callback receiving current bids
```
fraszczakszymon: Incorrect description

src/ad-bidders/wrappers/apstag.js
Bielik20: It is not needed if callback is preset.
The benefit is that now we can use also promise, not only callback. So we are able to do this:
```ts
const currentBids = await this.apstag.fetchBids({ slots, timeout: this.timeout });
```

src/ad-bidders/wrappers/cmp.js
Bielik20: What would be correct?

fraszczakszymon: @Bielik20 I mean missing `param` variable

PR #233, opened 2018-11-28
ADEN-7950 BTL status per request

jacekbj:


src/ad-services/bill-the-lizard/index.js
```diff
@@ -127,126 +139,211 @@ class BillTheLizard {
...
-		if (targeting.length > 0) {
-			context.set('targeting.btl', targeting);
+	/**
+	 * Sets DFP targeting in context.
+	 */
+	setTargeting() {
+		const targeting = this.getTargeting();
+		if (Object.keys(targeting).length > 0) {
+			context.set(
+				'targeting.btl',
```
fraszczakszymon: Just wonder if we should set those results on slot level params instead of page level

jacekbj: I don't know of any use of the targeting at all

PR #236, opened 2018-11-30
ADEN-7989 Add option to stick ads with line item id equal to "null".

jacekbj:


src/ad-products/templates/sticky-base.js
```diff
@@ -0,0 +1,89 @@
...
+	static SLOT_UNSTICK_IMMEDIATELY = 'force-unstick';
+	static ADX = 'AdX';
+
+	/**
+	 * Base class for sticky ads
+	 * @param {AdSlot} adSlot
+	 */
+	constructor(adSlot) {
+		this.adSlot = adSlot;
+		this.lineId = adSlot.lineItemId === 'null' ? StickyBase.ADX : adSlot.lineItemId;
```
fraszczakszymon: What about 
> event:isEmpty != empty?

part?

Flag is exposed in `event` object:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/listeners/slot-listener.js#L18

However -- what do you think about saving `AdX` information here:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/listeners/slot-listener.js#L78
e.g.
```
adSlot.creativeId = !event.isEmpty && response.creativeId === null ? 'AdX' : response.creativeId;
adSlot.lineItemId = !event.isEmpty && response.lineItemId === null ? 'AdX' :  ? response.lineItemId;
```

Having that we'd have info about AdX also in DOM element slot `data-*` params and StickyBase wouldn't require any "AdX" related logic.

src/ad-products/templates/sticky-tlb.js
```diff
@@ -58,14 +44,27 @@ export class StickyTLB {
 		};
 	}
 
+	constructor(adSlot) {
+		super(adSlot);
+		this.container = document.getElementById(this.adSlot.getSlotName());
```
fraszczakszymon: Can we use `this.adSlot.getElement()`?

src/ad-products/templates/sticky-base.js
```diff
@@ -0,0 +1,89 @@
...
+
+	isEnabled() {
+		const isEnabledInContext = context.get(`templates.${this.getName()}.enabled`);
+		const isLineAndGeo = StickyBase.isLineAndGeo(this.lineId, this.lines);
+		const isEnabled = isEnabledInContext && isLineAndGeo;
+
+		if (isEnabled) {
+			utils.logger(logGroup, `enabled with line item id ${this.lineId}`);
+
+			// DISCUSS: Should we set lineItemId to ADX for all slots at creation time
```
fraszczakszymon: +1 for all of them
Then StickyBase doesn't need to know about AdX at all

src/ad-engine/services/slot-data-params-updater.js
```diff
@@ -2,6 +2,8 @@ import { GptSizeMap } from '../providers';
 import { slotTweaker } from './slot-tweaker';
 import { context } from './context-service';
 
+export const ADX = 'AdX';
```
fraszczakszymon: What about moving it to ad-slot / gpt-provider?

src/ad-engine/services/slot-data-params-updater.js
```diff
@@ -13,8 +15,14 @@ class SlotDataParamsUpdater {
 
 	updateOnRenderEnd(adSlot, event) {
 		if (event) {
-			slotTweaker.setDataParam(adSlot, 'gptLineItemId', event.lineItemId);
```
fraszczakszymon: What about 
```
slotTweaker.setDataParam(adSlot, 'gptLineItemId', adSlot.lineItemId);
slotTweaker.setDataParam(adSlot, 'gptCreativeId', adSlot.creativeId);
```

PR #241, opened 2018-12-05
Aden 7984

Bielik20:


src/ad-engine/ad-engine.js
```diff
@@ -102,8 +108,16 @@ export class AdEngine {
 		}
 	}
 
-	getProvider(name) {
-		return this.providers.get(name);
+	/**
+	 * Returns GptProvider if it is set in context.
```
jacekbj: It returns any provider set (can be Prebidium). Please correct the description.

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+
+	/** @private */
+	getAdId(adSlot) {
+		const winning = context.get(`slots.${adSlot.config.slotName}.targeting`);
+		return winning.hb_adid;
+	}
+
+	@decorate(postponeExecutionUntilPbjsLoads)
+	flush() {
+		// TODO: Something should be here
```
jacekbj: Where is it then? :trollface: 3

Bielik20: In GtpProvider it uses google tag to refresh ads. Since we do not have google tag it is not that clear how should we approach it. Moreover in the example pages we cannot emulate mobile wikia behavior (SPA) so it is hard to test what are actual consequences of not having that implemented.

src/ad-engine/ad-engine.js
Bielik20: Yeah, that is correct. I have changed it recently.

src/ad-engine/utils/timer.js
```diff
@@ -0,0 +1,35 @@
+export class Timer {
```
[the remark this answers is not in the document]
Bielik20: For debug, now it is not used anywhere. We can delete it but I thought it may be of use in the future.

src/ad-engine/utils/timer.js
```diff
@@ -0,0 +1,35 @@
...
+	constructor() {
+		this.start = 0;
+		// this.clock = performance;
+		this.clock = {
+			now: () => 0
```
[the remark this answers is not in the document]
Bielik20: Yeah, that's a workaround. The commented line is correct but when running specs it screams that it cannot find performance. W spent line 15-30 minutes trying to find more elegant solution but wasn't successful. Either we need to set it somehow in webpack or ignore this file from specs entirely.

src/ad-engine/utils/lazy-queue.js
```diff
@@ -14,3 +14,60 @@ export function makeLazyQueue(queue, callback) {
 		throw new Error('LazyQueue requires an array as the first parameter');
 	}
 }
+
+// TODO: Proposal
+// example: https://stackblitz.com/edit/wikia-lazy-queue
+export class LazyQueue {
```
jacekbj: Looks really promising!
It's missing validation of callback type - is it on purpose?

Bielik20: On the one hand, good point, I could add that. On the other I am not in favor of checking input arguments, see [Design by Contract](https://www.eiffel.com/values/design-by-contract/introduction/).

So what is your opinion? Should I add validation?

src/ad-engine/ad-engine.js
```diff
@@ -53,20 +52,27 @@ export class AdEngine {
...
 		this.adStack = context.get('state.adStack');
-
 		if (!this.adStack.start) {
 			makeLazyQueue(this.adStack, (ad) => {
-				const gpt = this.providers.get('gpt');
-
-				fillInUsingProvider(ad, gpt);
+				fillInUsingProvider(ad, this.provider);
 
 				if (this.adStack.length === 0) {
```
fraszczakszymon: Imo we should move it to `gpt-provider` (after filling in the slot) and then we wouldn't need to implement dummy method in `prebidium` https://github.com/Wikia/ad-engine/pull/241/files?utf8=%E2%9C%93&diff=unified&w=1#diff-2da0029f3d403fa3dbce0fb40665c5f9R54

src/ad-engine/ad-engine.js
```diff
@@ -53,20 +52,27 @@ export class AdEngine {
 	}
 
 	setupProviders() {
-		this.providers.set('gpt', new GptProvider());
+		const providerName = context.get('state.provider');
```
fraszczakszymon: What about preventing breaking-changs and keeping `gpt` as default?

src/ad-engine/ad-engine.js
```diff
@@ -102,8 +108,16 @@ export class AdEngine {
...
 	}
 
-	getProvider(name) {
-		return this.providers.get(name);
+	/**
+	 * Returns GptProvider if it is set in context.
+	 * @param name - deprecated, works only for gpt
+	 * @returns {any}
+	 */
+	getProvider(name = undefined) {
```
fraszczakszymon: It seems that we can remove this method

src/ad-engine/providers/gpt-provider.js
```diff
@@ -1,15 +1,17 @@
 import { decorate } from 'core-decorators';
-import { logger, defer } from '../utils';
+import { defer, logger } from '../utils';
```
fraszczakszymon: 🙇

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	/** @private */
+	getAdId(adSlot) {
+		const winning = context.get(`slots.${adSlot.config.slotName}.targeting`);
+		return winning.hb_adid;
+	}
+
+	@decorate(postponeExecutionUntilPbjsLoads)
+	flush() {
```
fraszczakszymon: We may remove it (https://github.com/Wikia/ad-engine/pull/241/files?utf8=%E2%9C%93&diff=unified&w=1#r239388006)

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+	/** @private */
+	getIframeDoc(adSlot) {
+		const iframe = this.iframeBuilder.create(adSlot);
+		return iframe.contentWindow.document;
+	}
+
+	/** @private */
+	getAdId(adSlot) {
+		const winning = context.get(`slots.${adSlot.config.slotName}.targeting`);
+		return winning.hb_adid;
```
fraszczakszymon: What about `return context.get(`slots.${adSlot.getSlotName()}.targeting.hb_adid`);`?

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+		const adId = this.getAdId(adSlot);
+
+		window.pbjs.renderAd(doc, adId);
+		logger(logGroup, adSlot.getSlotName(), 'slot added');
+
+		this.forceWrappedFillInCallback(adSlot);
+	}
+
+	/**
+	 * This method causes wrappedFillInCallback method in BtfBlockerService["push"] to fire.
```
fraszczakszymon: Imo we should avoid using BtfBlockerService in prebidium. There is no point to postpone BTF

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+
+	@decorate(postponeExecutionUntilPbjsLoads)
+	flush() {
+		// TODO: Something should be here
+	}
+}
+
+class IframeBuilder {
+	create(adSlot) {
+		const doc = document.getElementById(`${adSlot.config.slotName}`);
```
fraszczakszymon: adSlot.config.slotName -> adSlot.getSlotName()

fraszczakszymon: Or even
```
const container = adSlot.getElement()
```

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+		const iframe = doc.appendChild(document.createElement('iframe'));
```
fraszczakszymon: For slightly better perf: create iframe, style it and append it at the end of operations please

src/ad-engine/services/btf-blocker-service.js
```diff
@@ -83,11 +86,12 @@ class BtfBlockerService {
 		}
 
 		if (!this.firstCallEnded && !adSlot.isFirstCall()) {
-			this.slotsQueue.push({ adSlot, fillInCallback: wrappedFillInCallback });
+			this.slotsQueue.push({ adSlot, fillInCallback: (...args) => wrappedFillInCallback(...args) });
 			logger(logGroup, adSlot.getSlotName(), 'second call slot pushed to queue');
 			return;
 		}
 
+		// TODO: 'wrappedFillInCallback' doesn't take argument
```
fraszczakszymon: That's right. Would you remove it please?

src/ad-engine/ad-engine.js
```diff
@@ -22,7 +22,7 @@ function fillInUsingProvider(ad, provider) {
 
 	slotService.add(adSlot);
 
-	btfBlockerService.push(adSlot, provider.fillIn.bind(provider));
+	btfBlockerService.push(adSlot, arg => provider.fillIn(arg));
```
fraszczakszymon: What about naming it `adSlot`? (instead of `arg`)

src/ad-engine/providers/prebidium-provider.js
fraszczakszymon: It will require moving `btfBlockerService` to gpt-provider (from ad-engine), which seems to be fine because only gpt-provider is going to use btf blocker (?)

src/ad-engine/utils/timer.js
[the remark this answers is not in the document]
fraszczakszymon: YAGNI :)

src/ad-engine/utils/lazy-queue.js
fraszczakszymon: Isn't it a bit of overengineering? I think there is no usecase to have multiple itemFlushCallbacks. Also I'm not sure about defining lazy queue without "flushCallback". It may turn out that you'll shift all items without executing anything and noticing that. I believe that the idea of having array instance from the beginning and turning it to 'callbacks queue' after "flushing" is pretty well-known in our env (e.g. `pbjs.que`, `googletag.cmd`) so won't it be misleading once we create second interface for that?

src/ad-engine/providers/prebidium-provider.js
fraszczakszymon: On the other hand -- we don't need to take it from targeting, we can use pbjs methods

src/ad-engine/utils/lazy-queue.js
Bielik20: I wouldn't call it overengineering. The design is quite simple. True there are more lines of code but are quite easy to read. I thought about using real command pattern, but discarded it because that would be in fact everengineering.

I though about `pbjs.que` and `googletag.cmd`. The difference is that you don't specify what should happen with those queues during "creation". I think this subtle difference is quite important. Moreover you don't know their exact specification (what they do under the hood). They also use this design because of lack of alternatives. They are not fired on flush/start, they are fired when they can, when library loads thus they cannot provide different way of interacting with their library.

The current solution has few drawbacks:
- It is unpure function. It modifies input argument in _unpredictable_ way.
- You specify a callback in a way that is not obvious for developer what it actually does. Just by looking at `makeLazyQueue` you don't know when this callback is going to fire. Of course you know that because you wrote it, but I was very confused first few times I saw it.
- The life cycle of that creation is also quite unclear for the developer. You basically monkey patching an array. I would avoid monkey patching unless absolutely necessary.
- You need to create this lazy queue in the place of its consumption i.e. only creator can pass callback. I bet this is not always desirable. 
- Should someone use `append` or extend array in any other way than `push`, then behaviour would not be desirable.

I know you are concerned about having queue without a callback. I think it is valid concern. Still I think that it does not belong in the creation. It is not very verbose. 

As for "no usecase to have multiple itemFlushCallbacks", true but still it doesn't hurt us and this functionality comes free without penalties with this design.

src/ad-engine/ad-engine.js
Bielik20: gpt is default, it is set in the context. The ratianale for the change is that there was a dictionary (providers), but entire application was dependent on gpt. So i figured it would be cleaner to just use `provider`.

src/ad-engine/ad-engine.js
Bielik20: But I don't know who is using it. You said yourself that changing any public method is a braking change.

src/ad-engine/ad-engine.js
Bielik20: Yeah, I was thinking about moving it to the base class, I must have forgotten about it during the process. I will do that.

src/ad-engine/providers/prebidium-provider.js
Bielik20: No, it is used by lazy queue in `AdEngine`->`setupQueue`.
Damian said the `flush` from `GptProvider` is used by mobile wiki to refresh adds. I had a hard time trying to test that :/

src/ad-engine/ad-engine.js
Bielik20: There would have been new key anyway. I could add gpt if none found, would that be ok?

```
default:
  this.provider = new GptProvider();
```

fraszczakszymon: :+1:

src/ad-engine/providers/prebidium-provider.js
fraszczakszymon: `flush` is used to request first bulk of ads (1 for mobile-wiki) at once
We can remove it assuming that we move btfBlockedService into gpt-provider, so it won't be used in the `AdEngine`->`setupQueue`, right?

fraszczakszymon: I mean this conversation: https://github.com/Wikia/ad-engine/pull/241#discussion_r239388006

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,74 @@
...
+		iframe.frameBorder = 0;
+		iframe.onload = () => this.removeBodyMargin(iframe);
+		return iframe;
+	}
+
+	/** @private */
+	removeBodyMargin(iframe) {
+		// eslint-disable-next-line prefer-destructuring
```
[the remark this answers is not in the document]
Bielik20: I am not sure it would keep a reference. That's why it is called deconstructing syntax. Still I am not sure I could test that.

src/ad-engine/providers/prebidium-provider.js
Bielik20: That you, I will change that.

If I had only had types :P

src/ad-engine/providers/prebidium-provider.js
Bielik20: I don't think there is a difference. How js works:
- there is an event loop
- it executes all synchronous code
- once in a while it executes things from tasks queue (aka setTimeout)
- once in a while (but not as frequently) it updates UI (aka requestAnimationFrame).

Since this is synchronous code it doesn't matter the order, it will be updated (the UI) in next cycle. For more information: https://www.youtube.com/watch?v=cCOL7MC4Pl0

src/ad-engine/services/btf-blocker-service.js
Bielik20: Question is what do we want. Do we want it not to take an argument, so that it is delivered by `push` method of `btf-blocker`. Or do we want the caller to provide it.

In other words, do we want to remove or add it?

src/ad-engine/ad-engine.js
Bielik20: There is `adSlot` above. It would shadow variable. Do note it is not `fillInUsingProvider` that is providing the argument but `btfBlockerService.push` method. In this case the argument is the same but still. This is very similar to the `wrappedFillInCallback`. One reason why I am not a fan of callbacks, they get messy really fast.

In my opinion the cleanest solution is:
```
(...args) => provider.fillIn(...args)
```
Then doesn't matter the change that is done to `fillIn` method it will always work. This is also direct transfer of functionality provided by:
```
provider.fillIn.bind(provider)
```

fraszczakszymon: Sorry, I didn't catch that adSlot is already defined in this scope 🙈

src/ad-engine/providers/prebidium-provider.js
Bielik20: I would like discuss it in person. I will leave that for tomorrow.

src/ad-engine/utils/timer.js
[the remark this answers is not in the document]
Bielik20: Well it is not the fist time I have implemented it but I will listen to your best judgment. Should I remove it?

src/ad-engine/providers/prebidium-provider.js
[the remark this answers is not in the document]
Bielik20: Yeah, it works.

src/ad-engine/providers/prebidium-provider.js
Bielik20: Will `pbjs` return winning bid? Isn't that done at `ad-bidders`?

fraszczakszymon: Oh, you're right. I didn't thought about that. Now I started thinking what would be the best place for `prebidium-provider` - `ad-engine` vs `ad-bidders`

Bielik20: `src/ad-engine/providers` is where `GptProvider` is stored. Do you consider moving them both or just `PrebidiumProvider`?

fraszczakszymon: Just `PrebidiumProvider`, but then we'd need to create some kind of way to register providers for further usage

src/ad-engine/providers/prebidium-provider.js
Bielik20: `ad-engine` still uses `btfBlockerService.init();` in its init method.

src/ad-engine/ad-engine.js
Bielik20: Fixed in `ADEN-7984 Move fillInUsingProvider to provider (breaking change)`

spec/ad-engine/providers/prebidium-provider.spec.js
```diff
@@ -0,0 +1,106 @@
+import sinon from 'sinon';
+import { assert } from 'chai';
+import { context, PrebidiumProvider, slotService } from '@wikia/ad-engine';
+
+describe('PrebidiumProvider', () => {
+	let sandbox;
+	let prebidiumProvider;
+	const stub = {
+		// utils: {
```
dmnsobczak: Let's remove unused comments

spec/ad-engine/providers/prebidium-provider.spec.js
```diff
@@ -0,0 +1,106 @@
...
+	};
+
+	beforeEach(() => {
+		sandbox = sinon.sandbox.create();
+		prebidiumProvider = new PrebidiumProvider();
+
+		stubIframeBuilder();
+		stub.slotService.add = sandbox.stub(slotService, 'add');
+		stub.context.get = sandbox.stub(context, 'get').returns(mock.adId);
+		// stub.utils.logger = sandbox.stub(utils, 'logger');
```
dmnsobczak: ToDo: Remove

spec/ad-engine/providers/prebidium-provider.spec.js
```diff
@@ -0,0 +1,106 @@
...
+		});
+
+		it('should call context get with correct argument', () => {
+			const argument = stub.context.get.getCall(0).args[0];
+
+			assert(stub.context.get.calledOnce);
+			assert.equal(argument, `slots.${mock.slotName}.targeting.hb_adid`);
+		});
+
+		// it('should call logger', () => {
```
dmnsobczak: Why it is commented out?

src/ad-engine/providers/iframe-builder.js
```diff
@@ -0,0 +1,32 @@
+// TODO: Think of a better place for this class
```
dmnsobczak: Should we left ToDo's in code?

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -1,6 +1,7 @@
 import { decorate } from 'core-decorators';
 import { logger } from '../utils';
 import { context, slotService } from '../services';
+import { IframeBuilder } from './iframe-builder';
```
dmnsobczak: Why not `iframeBuilder`?

src/ad-engine/providers/iframe-builder.js
Bielik20: utils?

src/ad-engine/providers/prebidium-provider.js
```diff
@@ -0,0 +1,40 @@
+import { decorate } from 'core-decorators';
+import { IframeBuilder, logger } from '../utils';
+import { context, slotService } from '../services';
+
+const logGroup = 'prebidium-provider';
+
+// TODO: Duplicate from ad-bidders/prebid/index.js
```
nandy-andy: Can we create a ticket for it and put it in the comment? `// TODO`s are risky. To minimise risk we can:
* resolve TODO within the current task,
* create a follow-up ticket and put it's number like ADEN-XXXX after TODO

examples/slots/common-slots/index.html
```diff
@@ -14,7 +14,7 @@
 			googletag.cmd = googletag.cmd || [];
 
 			(function() {
-				var gads = document.createElement('script'),
+				const gads = document.createElement('script'),
```
fraszczakszymon: It's not going to work on IE, am I right?

src/ad-engine/ad-engine.js
```diff
@@ -55,21 +45,27 @@ export class AdEngine {
...
+		const providerName = context.get('state.provider');
+
+		switch (providerName) {
+			case 'gpt':
+				this.provider = new GptProvider();
+				break;
+			case 'prebidium':
+				this.provider = new PrebidiumProvider();
+				break;
+			default:
```
fraszczakszymon: It can be simplified:
```
case 'gpt':
deafault: 
  this.provider = new GptProvider();
```

src/ad-engine/ad-engine.js
```diff
@@ -55,21 +45,27 @@ export class AdEngine {
...
 			makeLazyQueue(this.adStack, (ad) => {
-				const gpt = this.providers.get('gpt');
-
-				fillInUsingProvider(ad, gpt);
+				const adSlot = new AdSlot(ad);
 
-				if (this.adStack.length === 0) {
-					gpt.flush();
-				}
+				this.provider.fillIn(adSlot);
```
fraszczakszymon: It seems that we can add slot to slotService before filling in
```
slotService.add(adSlot);
```
because both prebidium and gpt providers are doing it at the beginning

src/ad-engine/utils/iframe-builder.js
```diff
@@ -0,0 +1,31 @@
...
+
+		return doc.appendChild(this.createEmptyIframe());
+	}
+
+	/**
+	 * @private
+	 * Inspired by `createInvisibleIframe` method from Prebid.js
+	 */
+	createEmptyIframe() {
+		const f = document.createElement('iframe');
```
fraszczakszymon: s/f/iframe please

PR #247, opened 2018-12-13
ADEN-8061 Adding Nielsen DCR integration code

dmnsobczak:


src/ad-services/nielsen/static-queue-script.js
```diff
@@ -0,0 +1,4 @@
+// NIELSEN CODE START
+// eslint-disable-next-line
+export function initNielsenStaticQueue(){!function(t,n){t[n]=t[n]||{nlsQ:function(e,o,c,r,s,i){return s=t.document,r=s.createElement("script"),r.async=1,r.src=("http:"===t.location.protocol?"http:":"https:")+"//cdn-gl.imrworldwide.com/conf/"+e+".js#name="+o+"&ns="+n,i=s.getElementsByTagName("script")[0],i.parentNode.insertBefore(r,i),t[n][o]=t[n][o]||{g:c||{},ggPM:function(e,c,r,s,i){(t[n][o].q=t[n][o].q||[]).push([e,c,r,s,i])}},t[n][o]}}}(window,"NOLBUNDLE");}
```
fraszczakszymon: Is seems that this code loads the JS resource and creates NOLBUNDLE lazy queue. What do you think about rewriting it? 🙈

src/ad-services/nielsen/index.js
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	call(nielsenMetadata) {
+		const nielsenKey = context.get('services.nielsen.appId');
+
+		if (!context.get('services.nielsen.enabled') || !nielsenKey) {
+			utils.logger(logGroup, 'disabled');
+		}
+
+		if (!this.nlsnInstance) {
+			this.nlsnInstance = createInstance(nielsenKey);
```
fraszczakszymon: How does it work? `createInstance` doesn't return anything

dmnsobczak: It is now, I forgot about `return`.

PR #248, opened 2018-12-13
ADEN-7541 Add A9 to incontent_boxads. Enable bid refreshing for A9.

jacekbj:


examples/context.js
```diff
@@ -208,7 +212,7 @@ export default {
 			},
 		},
 		repeatable_boxad_1: {
-			bidderAlias: 'top_boxad',
+			bidderAlias: 'mobile_in_content',
```
fraszczakszymon: It's going to break prebid example pages because:
https://github.com/Wikia/ad-engine/blob/dev/examples/context.js#L43
there is no mobile_in_content in the wikia adapters definition. Am I right?

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+	}
+
+	/**
+	 * Creates A9 slot definition from slot name.
+	 *
+	 * @param {string} slotName
+	 * @returns {A9SlotDefinition | null} Returns null i
+	 */
+	createSlotDefinition(slotName) {
+		const slotAlias = this.getSlotAlias(slotName);
```
fraszczakszymon: I'm not sure if I understand it correctly. Why do we need to care about slotAlias at this point? Can it work in described way:

1. request bids for normal slot (e.g. mobile_in_content)
2. create slot with alias (e.g. incontent_boxad_1 with alias mobile_in_content)
3. add bid targeting to incontent_boxad_1 using alias (so it takes bids from mobile_in_content)
4. render a9 ad in incontent_boxad_1 using bid for mobile_in_content

For me `createSlotDefinition` is responsible for point 1.

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+			.loadScript('//c.amazon-adsystem.com/aax2/apstag.js', 'text/javascript', true, 'first')
+			.then(() => this.overwriteRenderImp());
 	}
 
+	/**
+	 * Wraps apstag.renderImp
+	 *
+	 * Calls this.refreshBid() if bids refreshing is enabled.
+	 */
+	overwriteRenderImp() {
```
fraszczakszymon: Isn't it sensitive to race conditions? Look at the `configureApstag` please. It might throw excpetion that original `window.apstag.renderImp` is not defined

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+
+		return renderedSlot;
+	}
+
+	/**
+	 * Refreshes bid for given slot.
+	 *
+	 * @param {string | number} impId
+	 */
+	refreshBid(slot) {
```
fraszczakszymon: Should we emit `BIDS_REFRESH` somewhere here?

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+		if (!this.shouldRefreshSlot(slot)) {
+			return;
+		}
+
+		const slotDef = this.createSlotDefinition(slot.getSlotName());
+
+		if (slotDef) {
+			utils.logger(logGroup, 'refresh bids for slot', slotDef);
+			this.fetchBids([slotDef]);
```
fraszczakszymon: Just curious: is it safe for the rest of bids? Won't it clear all other slots?

src/ad-bidders/base-bidder.js
```diff
@@ -56,14 +56,14 @@ export class BaseBidder {
 		return this.response;
 	}
 
-	isSlotSupported(slotName) {
-		if (this.isSupported) {
-			return this.isSupported(slotName);
-		}
-
+	isSupported() {
```
fraszczakszymon: :+1:

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+	/**
+	 * Creates A9 slot definition from slot name.
+	 *
+	 * @param {string} slotName
+	 * @returns {A9SlotDefinition | null} Returns null i
+	 */
+	createSlotDefinition(slotName) {
+		const slotAlias = this.getSlotAlias(slotName);
+		const config = this.slots[slotAlias];
+		const slotID = config.slotId || this.slots[slotAlias];
```
fraszczakszymon: Is it fine? `slotId` can be `config.slotId` or `this.slots[slotAlias]` (which equals `config`)

src/ad-bidders/a9/index.js
```diff
@@ -154,19 +152,136 @@ export class A9 extends BaseBidder {
...
+			slotName: slotID,
+		};
+
+		if (!slotService.getState(slotID)) {
+			return null;
+		}
+
+		this.slotNamesMap[slotID] = slotAlias;
+
+		// DISCUSS Do we enable A9 video bidder anywhere?
```
fraszczakszymon: Ask Nef please. At this moment we have two separate instant globals for that

examples/context.js
jacekbj: Surprisingly it works in for both /bidders/reusable-prebid/ and /bidders/prebid/. It definately needs checking, I might be missing something.

src/ad-bidders/a9/index.js
jacekbj: It's called after apstag script `onload` event is triggered. I did not experience race condition when developing locally.

src/ad-bidders/a9/index.js
jacekbj: Yes, it is. Function called once bids are fetched updates only the slots for which bids were fetched.

src/ad-bidders/a9/index.js
jacekbj: 🔭  good spot

src/ad-bidders/a9/index.js
jacekbj: I realized that it does not break anything. I will delete this comment.

examples/context.js
jacekbj: One more thing: setting bidderAlias to a slot which exists already on the page results in two bid responses writing to the same targeting. This is what happened in this case.

src/ad-bidders/a9/index.js
jacekbj: I refactored `createSlotDefinition` to accept slot name, which is already transformed into alias if necessary.

PR #249, opened 2018-12-17
ADEN-8070 Get last id from wrappers

fraszczakszymon:


src/ad-engine/video/vast-parser.js
```diff
@@ -1,6 +1,22 @@
 import { queryString } from '../utils';
 
 class VastParser {
+	/**
+	 * @private
+	 */
+	getLastNumber(possibleValues) {
```
nandy-andy: Ditto: https://github.com/Wikia/app/pull/16498#discussion_r242213311

PR #254, opened 2018-12-19
Aden 8038

Bielik20:


src/ad-products/templates/sticky-base.js
```diff
@@ -53,24 +52,20 @@ export class StickyBase {
...
 
-		lineId = lineId.toString();
-
-		lines.forEach((line) => {
-			line = line.split(':', 2);
-
-			if (line[0] === lineId && (!line[1] || utils.isProperGeo([line[1]]))) {
-				found = true;
-			}
+			return +lineId === +this.lineId && (!geo || utils.isProperGeo([geo]));
```
fraszczakszymon: praseInt / toString would be way more readable
lineId can contain `AdX` value -- we should map it to string

Bielik20: Then why do we assume it may be number? Shouldn't it always be string?

src/ad-products/templates/sticky-base.js
```diff
@@ -53,24 +52,23 @@ export class StickyBase {
...
 
-		lineId = lineId.toString();
+		this.lines.forEach((line) => {
+			const [lineId, geo] = line.split(':', 2);
 
-		lines.forEach((line) => {
-			line = line.split(':', 2);
-
-			if (line[0] === lineId && (!line[1] || utils.isProperGeo([line[1]]))) {
+			if (+lineId === this.lineId && (!geo || utils.isProperGeo([geo]))) {
```
jacekbj: Shouldn't `this.lineId` be also cast to Number?

jacekbj: It's `this.lineId` that should be cast to `String`.

PR #256, opened 2018-12-20
ADEN-7957 | Create killcode for all ads but top leaderboard

bckatarzyna:


src/ad-engine/services/btf-blocker-service.js
```diff
@@ -45,10 +46,12 @@ class BtfBlockerService {
...
 		logger(logGroup, 'first call queue finished');
 
-		if (window.ads.runtime.disableBtf) {
+		if (window.ads.runtime.disableSecondCall) {
+			this.disableSecondCall([]);
+		} else if (window.ads.runtime.disableBtf) {
 			this.disableSecondCall([
 				...this.unblockedSlotNames,
-				...slotService.getAtfSlotConfigs().map((slot) => slot.name),
+				...slotService.getAtfSlotConfigs().map((slot) => slot.slotName),
```
Bielik20: How do we know that a given slot has a `slotName`?

Bielik20:
https://github.com/Wikia/ad-engine/blob/0dd654385f0d4de42dfb15e71af4a3af7946bee9/tests/specs/slots/kill-codes.desktop.test.js#L10-L45

Consider placing "Check BTF boxad", "Check ATF boxad" and "Check ATF and BTF boxads" inside separate `describe`s. It would be easier to read as `it` message could be shorter.

PR #262, opened 2019-01-04
Aden 8145

Bielik20:


src/ad-engine/ad-engine.js
```diff
@@ -58,69 +64,89 @@ export class AdEngine {
...
 
-		if (promises.length > 0) {
-			Promise.all(promises).then(() => {
+		if (delayModulesPromises.length > 0) {
+			Promise.all(delayModulesPromises).then(() => {
 				logger(logGroup, 'startAdQueue', 'All modules ready');
-				startAdQueue();
+				this.flushAdStackQueue(timeout);
 			});
 			timeout = setTimeout(() => {
```
fraszczakszymon: What do you think about changing it to "timeoutPromise" and running `this.flushAdStackQueue()` on `Promise.race([Promise.all(delayModulesPromises), timeoutPromise]`? Then we don't need to keep `timeout` and `this.started` flag

Bielik20: Done, but I don't know about `this.started` flag. `runAdQueue` method is public so I am not sure it isn't called more than once.

fraszczakszymon: Good point, let's keep it then

PR #272, opened 2019-01-14
ADEN-8168 Adding Confiant service caller

dmnsobczak:


src/ad-services/confiant/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+		if (!context.get('services.confiant.enabled') || !propertyId || !mapping || !activation) {
+			utils.logger(logGroup, 'disabled');
+
+			return Promise.resolve();
+		}
+
+		utils.logger(logGroup, 'loading');
+
+		/* eslint-disable */
```
fraszczakszymon: Why? :p we can defined _clrm as global

src/ad-services/confiant/index.js
```diff
@@ -0,0 +1,57 @@
...
+		window._clrm = window._clrm || {};
+		window._clrm.gpt = {
+			propertyId: propertyId,
+			confiantCdn: scriptDomain,
+			sandbox: 0,
+			mapping: mapping,
+			activation: activation,
+			callback: function(blockingType, blockingId, isBlocked, wrapperId, tagId, impressionData) {
+				console.log('w00t one more bad ad nixed.', arguments);
```
fraszczakszymon: Is it needed?

src/ad-services/confiant/index.js
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	call() {
+		const propertyId = context.get('services.confiant.propertyId');
+		const mapping = context.get('services.confiant.mapping');
+		const activation = context.get('services.confiant.activation');
+
+		if (!context.get('services.confiant.enabled') || !propertyId || !mapping || !activation) {
+			utils.logger(logGroup, 'disabled');
+
+			return Promise.resolve();
```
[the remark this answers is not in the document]
jacekbj: In Krux and MoatYI we return `Promise.resolve()`.

src/ad-services/confiant/index.js
dmnsobczak: It's a part of code copied from Confiant ready-to-paste script tag, so I wanted to mark this as an alien code. But you are probably right :)

src/ad-services/confiant/index.js
dmnsobczak: It's a part of code copied from Confiant ready-to-paste script tag. Looks that we can remove it, but guess I was not brave enough :<

src/ad-services/confiant/index.js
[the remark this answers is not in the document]
dmnsobczak: +1, I keeped consistency with other modules.

src/ad-services/confiant/index.js
```diff
@@ -34,19 +34,16 @@ class Confiant {
...
 		window._clrm = window._clrm || {};
+		// eslint-disable-next-line  no-underscore-dangle
 		window._clrm.gpt = {
-			propertyId: propertyId,
+			propertyId,
 			confiantCdn: scriptDomain,
 			sandbox: 0,
-			mapping: mapping,
-			activation: activation,
-			callback: function(blockingType, blockingId, isBlocked, wrapperId, tagId, impressionData) {
```
dmnsobczak: I wonder if removing this callback parameter won't break the Confiant code.

PR #277, opened 2019-01-23
ADEN-8235 Wdio tests fix

emadry:


tests/common/ad-slots.js
```diff
@@ -46,6 +46,15 @@ class AdSlots {
 		);
 	}
 
+	waitForSlotCollapsedManually(adSlot) {
+		browser.waitUntil(
+			() => browser.getElementSize(adSlot, 'height') < 2,
```
nandy-andy: 🤔 if we can describe it somewhere, so we can edit it once we get rid of 1x1 slots. I think the only ones right now are: `INVISIBLE_SKIN` and `INVISIBLE_HIGH_IMPACT` which we want to remove this year and replace them with `out-of-page` slots. There are also 2x2 and 3x3 but those are special case ones for UAP. So, now `< 2` makes sense since we may have 1x1 size in some slots but in future `< 1` will be enough, I guess.

tests/specs/video/porvata.desktop.test.js
```diff
@@ -51,10 +48,12 @@ describe('Porvata player', () => {
 		browser.waitForExist(`${porvata.unmuteButton}${porvata.iconHidden}`, timeouts.standard);
 	});
 
-	it('Check if opening full screen and redirect on fullscreen player works', () => {
+	// TODO wf fix of ADEN-8239
```
nandy-andy: 💎 I have no reason against TODOs with "a ticket linked" ;-) That's because we have a strong culture of cleaning up and I'm confident the ticket won't be closed without checking if this TODO is done or not needed anymore 👍

nandy-andy:
Feel free to bring timeout changes on board with the dev team. I'm all for changing it if the tests pass thanks to it but we need to have a curiosity and ask ourselves why it worked before and now it doesn't. Is it because the overall load time of our page increased? Is it because Jenkins machines are slower than our laptops where we used to run wdio tests? Perfect opportunity for an interesting discussion and hopefully actions we could take to improve things around us ;)

PR #282, opened 2019-01-30
ADEN-8228 Exposing some functions for AE3 Homad integration

dmnsobczak:


src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -70,6 +77,14 @@ function shouldPlayPostroll(videoDepth) {
 	return context.get('options.video.isPostrollEnabled') && canAdBePlayed(videoDepth);
 }
 
+function getCurrentVast(placement) {
+	const vast = vastUrls[placement];
+
+	vastUrls.last = vast;
```
jacekbj: Why is this value set here?

Aside, it feels bad to set a value in a method whose name begins with `get`.

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -90,6 +105,10 @@ function getVastUrl(slot, position, depth, correlator, slotTargeting) {
 			slotTargeting,
 		),
 	});
+
+	vastUrls[position] = vast;
```
jacekbj: ditto

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -266,6 +285,19 @@ function create(options) {
...
 
+		if (context.get('opts.wadHMD')) {
+			document.addEventListener('hdPlayerEvent', (event) => {
+				if (event.detail.slotStatus) {
+					updateSlotParams(slot, event.detail.slotStatus.vastParams);
+					slot.setStatus(event.detail.slotStatus.statusName);
+				}
+
+				if (event.detail.name) {
+					tracker.emit(event.detail.name, event.detail.errorCode);
```
jacekbj: Is `event.detail.errorCode` always present if `event.detail.name` is set? Is it possible that this piece of code will emit empty error codes?

src/ad-products/video/jwplayer-ads-factory.js
dmnsobczak: Maybe you have right, but idea of it was to save last requested vast and I found this unnecessary to complicate this more.

src/ad-products/video/jwplayer-ads-factory.js
dmnsobczak: Vast url is now builded and forgotten. We want to save it, because it will be needed again for Homad configuration.

src/ad-products/video/jwplayer-ads-factory.js
dmnsobczak: Param `errorCode` can be and most of the time will be empty. But that not an issue - in that case it's just not gonna be tracker (empty field value):
https://github.com/Wikia/ad-engine/blob/dev/src/ad-products/tracking/video/jwplayer-tracker.js#L159

src/ad-products/video/jwplayer-ads-factory.js
nandy-andy: I think it may still be worth keeping the logic clean. I imagine you can remove this line from the getter, create a setter with two parameters: position and vast. And re-use the setter in the hooks like:
```
const fillInSlot = () => {
   const vastUrl = getVastUrl(slot, 'preroll', depth, correlator, targeting);
   player.playAd(vastUrl);
   setVastUrl('preroll', vastUrl);
};
```

dmnsobczak: Ack, changed.

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -266,6 +294,19 @@ function create(options) {
 			slot.setStatus('error');
 		});
 
+		if (context.get('opts.wadHMD')) {
```
fraszczakszymon: `context` doesn't have `opts` key. There are `options`, using existing one please

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -12,6 +12,13 @@ import {
 import { JWPlayerTracker } from '../tracking/video/jwplayer-tracker';
 import featuredVideo15s from './featured-video-f15s';
 
+const vastUrls = {
```
fraszczakszymon: Shouldn't we keep it in the scope of `create` method?

src/ad-products/video/jwplayer-ads-factory.js
dmnsobczak: Done.

src/ad-products/video/jwplayer-ads-factory.js
dmnsobczak: Good idea, moved.

dmnsobczak: Discussed offline. Reverting.

PR #284, opened 2019-02-04
ADEN-8279 Add inViewportWith(out)Offer tracking

dmnsobczak:


fraszczakszymon:
Why existing logic:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-products/templates/outstream/porvata-template.js#L109-L115
does not work?
porvata.js module is used in UAPs, is it safe for them? Shouldn't we work on porvata-template.js module?

dmnsobczak:
You right, my idea was wrong. This PR is unnecessary. Closing.

PR #285, opened 2019-02-04
ADEN-8232 Export Stickiness in ad-products.

jacekbj:


src/ad-engine/models/ad-slot.js
```diff
@@ -122,6 +122,15 @@ export class AdSlot extends EventEmitter {
 		}
 	}
 
+	/**
+	 * Decode dataset value stored under provided key.
+	 *
+	 * @param {string | number} key
+	 */
+	getDatasetValue(key) {
```
fraszczakszymon: Don't make logic based on debug attributes please. More description in:
https://github.com/Wikia/mobile-wiki/pull/1361/files#r253500562

src/ad-engine/models/ad-slot.js
```diff
@@ -189,6 +200,15 @@ export class AdSlot extends EventEmitter {
 		}
 	}
 
+	updateWinningPbBidderDetails(targeting) {
+		if (targeting.hb_bidder && targeting.hb_pb) {
```
fraszczakszymon: What about `this.targeting`?

src/ad-engine/models/ad-slot.js
fraszczakszymon: This one seems not needed now

PR #286, opened 2019-02-05
ADEN-8232 Store winning Prebid bidder info in AdSlot.

jacekbj:


src/ad-bidders/prebid/index.js
```diff
@@ -12,6 +12,16 @@ function postponeExecutionUntilPbjsLoads(method) {
 	};
 }
 
+events.on(events.VIDEO_AD_IMPRESSION, (adSlot, vastParams) => {
```
fraszczakszymon: I believe we should stick to `VIDEO_AD_REQUESTED` because you won't get AD_IMPRESSION if there is an error in VAST, so player will try to use that bid all the time. What do you think?

src/ad-bidders/prebid/index.js
```diff
@@ -12,6 +12,16 @@ function postponeExecutionUntilPbjsLoads(method) {
...
+	// Mark ad as rendered
+	if (vastParams.customParams && vastParams.customParams.hb_adid) {
+		if (window.pbjs && typeof window.pbjs.markWinningBidAsUsed === 'function') {
+			window.pbjs.markWinningBidAsUsed({ adId: vastParams.customParams.hb_adid });
```
fraszczakszymon: 💪

src/ad-bidders/prebid/index.js
```diff
@@ -12,6 +12,16 @@ function postponeExecutionUntilPbjsLoads(method) {
 	};
 }
 
+events.on(events.VIDEO_AD_IMPRESSION, (adSlot, vastParams) => {
+	// Mark ad as rendered
+	if (vastParams.customParams && vastParams.customParams.hb_adid) {
```
fraszczakszymon: What about 
```
	const adId = context.get(`slots.${adSlot.getSlotName()}.targeting.hb_adid`);
	if (adId) {
```

src/ad-bidders/index.js
```diff
@@ -9,7 +9,11 @@ const realSlotPrices = {};
 const logGroup = 'bidders';
 
 events.on(events.VIDEO_AD_REQUESTED, (adSlot) => {
-	resetTargetingKeys(adSlot.getSlotName());
+	adSlot.updateWinningPbBidderDetails();
```
fraszczakszymon: s/WinningPbBidderDetails/WinningBidDetails
what do you think?

jacekbj: It's only Prebid bidder price we know. A9 is also considered a "bidder", so it would be ambiguous.

src/ad-bidders/prebid/index.js
jacekbj: 🙇

src/ad-bidders/index.js
fraszczakszymon: KK, I just don't like 'Pb' in name :D let's leave it for now

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -12,6 +12,14 @@ import {
 import { JWPlayerTracker } from '../tracking/video/jwplayer-tracker';
 import featuredVideo15s from './featured-video-f15s';
 
+// We need to rely on message instead of error code
+// because JWPlayer returns:
+// {adErrorCode: 21009, code: 900}
+// instead of 303.
+const EMPTY_TAG_MESSAGE = 'Ad Error: The VAST response document is empty.';
```
fraszczakszymon: According to JWPlayer docs 21009/900 stands for empty response. Can't we use that?
https://developer.jwplayer.com/jw-player/docs/developer-guide/advertising/ads_errors_reference/

src/ad-products/video/jwplayer-ads-factory.js
```diff
@@ -253,17 +261,30 @@ function create(options) {
...
+			events.emit(events.VIDEO_AD_IMPRESSION, slot);
 		});
 
 		player.on('adError', (event) => {
 			const vastParams = vastParser.parse(event.tag, {
 				imaAd: event.ima && event.ima.ad,
 			});
 
-			vastDebugger.setVastAttributesFromVastParams(videoContainer, 'error', vastParams);
+			if ([AdSlot.STATUS_COLLAPSE, AdSlot.STATUS_ERROR].indexOf(slot.getStatus()) !== -1) {
```
fraszczakszymon: Is it safe for mid/post roll?

src/ad-services/confiant/index.js
```diff
@@ -44,6 +44,7 @@ class Confiant {
 			mapping,
 			activation,
 			callback: (...args) => {
+				// eslint-disable-next-line  no-console
```
fraszczakszymon: As far as I know it has been changed in other pull request

PR #287, opened 2019-02-05
ADEN-8292 Fix Krux

nandy-andy:


src/ad-services/krux/index.js
```diff
@@ -82,12 +82,13 @@ class Krux {
...
-		const user = getKruxData('kxuser');
+		const oldUser = getKruxData('kxuser');
+		const newUser = getKruxData('kxwikia_user');
 		const segments = getKruxData('kxsegs');
 
-		context.set('targeting.kuid', user || null);
+		context.set('targeting.kuid', oldUser || newUser || null);
 		context.set('targeting.ksg', segments ? segments.split(',') : []);
-		utils.logger(logGroup, 'data set', user, segments);
+		utils.logger(logGroup, 'data set', oldUser, segments);
```
jacekbj: Shouldn't that be `utils.logger(logGroup, 'data set', oldUser, segments);` -> `utils.logger(logGroup, 'data set', oldUser || newUser, segments);`?

src/ad-services/krux/index.js
```diff
@@ -82,12 +82,13 @@ class Krux {
 	 * @returns {void}
 	 */
 	importUserData() {
-		const user = getKruxData('kxuser');
+		const oldUser = getKruxData('kxuser');
```
fraszczakszymon: What about 
```suggestion
		const user = getKruxData('kxuser') || getKruxData('kxwikia_user');
```

src/ad-services/krux/index.js
```diff
@@ -82,12 +82,13 @@ class Krux {
...
+		const newUser = getKruxData('kxwikia_user');
 		const segments = getKruxData('kxsegs');
```
fraszczakszymon: Should we handle segments too?
```suggestion
		const segments = getKruxData('kxsegs') || getKruxData('kxwikia_segs');
```

src/ad-services/krux/index.js
```diff
@@ -82,8 +82,8 @@ class Krux {
 	 * @returns {void}
 	 */
 	importUserData() {
-		const user = getKruxData('kxuser');
-		const segments = getKruxData('kxsegs');
+		const user = getKruxData('kxuser') || getKruxData('kxwikia_user');
```
fraszczakszymon: Just wonder if we shouldn't make it configurable in context (e.g. `services.krux.userKey`) and use it as a fallback 
```suggestion
		const user = getKruxData('kxuser') || getKruxData(context.get('services.krux.userKey'));
```

PR #289, opened 2019-02-06
ADEN-8302 Switch Beachfront to via-DFP integration

dmnsobczak:


src/ad-bidders/prebid/adapters/beachfront.js
```diff
@@ -16,6 +16,7 @@ export class Beachfront extends BaseAdapter {
 			code,
 			mediaTypes: {
 				video: {
+					context: 'instream',
```
fraszczakszymon: Shouldn't we use 'outstream'? Shouldn't it be configurable per slot?

dmnsobczak: If you use `outstream` then Prebid won't use VAST cache and will try to return bid + renderer/player. We ask for `instream` and then we do our custom in-DFP magic.

src/ad-services/confiant/index.js
```diff
@@ -44,7 +44,7 @@ class Confiant {
 			mapping,
 			activation,
 			callback: (...args) => {
-				console.log('w00t one more bad ad nixed.', args);
```
nandy-andy: 😆

dmnsobczak: It was messing with my linter output :x

PR #294, opened 2019-02-13
ADEN-8258 Possibility to configure slot class names on create

fraszczakszymon:


examples/context.js
```diff
@@ -158,6 +158,7 @@ export default {
 					sizes: [[728, 90], [3, 3]],
 				},
 			],
+			defaultClasses: ['i-am-the-default-class-added-on-create'],
```
dmnsobczak: 😄

nandy-andy: Explicit enough ;)

PR #306, opened 2019-02-19
ADEN-8183 Build ad-engine dist in prepare step

fraszczakszymon:
Remove dist/* directory from git versioning and build ad-engine package during `prepare` steps (which runs every time we install dependency).

We need to get rid of `window.ads.adEngineCommit` because there are no .git files in the yarn installation.

Since it shouldn't change behaviour in any of client's repo I believe we don't need to add any release-label.


package.json
```diff
@@ -95,14 +100,14 @@
...
-		"build": "./node_modules/webpack/bin/webpack.js --progress --colors --hide-modules && bash ./maintenance/build-bundle.sh",
+		"build": "webpack --progress --colors --hide-modules && node ./maintenance/build-bundle.js",
 		"ci-check": "npm run test && npm run lint",
 		"dev": "npm run serve",
 		"lint": "tslint --project . && ./maintenance/detect-incorrect-usages.sh",
 		"lintf": "tslint --project . --fix && ./maintenance/detect-incorrect-usages.sh",
 		"prettier": "prettier --write '**/*.{ts,js,json,css,scss,html}'",
 		"format": "npm run prettier && npm run lintf",
-		"prepare": "echo 'This is a test'",
+		"prepare": "echo '🍷 Building Wikia/ad-engine' && npm run build --production",
```
jacekbj: 🍷

PR #307, opened 2019-02-20
Ts cleanup 1

jacekbj:
Please review commit-by-commit as changes are not overlapping.


src/ad-engine/global.d.ts
```diff
@@ -0,0 +1,19 @@
...
+		ima: typeof google.ima;
+	};
+	[key: string]: any;
+}
+
+declare interface AdDisplayContainer extends google.ima.AdDisplayContainer { }
+
+declare var BlockAdBlock: any;
+
+declare var NOLBUNDLE: any;
```
nandy-andy: Is it `NOLBUNDLE` really or a typo of `NOBUNDLE`?

jacekbj: It's `NOLBUNDLE`: https://github.com/Wikia/ad-engine/blob/7b3908c5a7a866994c0b837d6b9b6996f0018b4a/src/ad-services/nielsen/index.ts#L1

PR #308, opened 2019-02-20
ADEN-8377 Fix A9 not rendering ads.

jacekbj:


src/ad-engine/models/ad-slot.ts
```diff
@@ -246,11 +246,15 @@ export class AdSlot extends EventEmitter {
 	/**
 	 * Appends gpt-ad class to adSlot node.
 	 */
-	addAdClass() {
+	addClass(cls: string): boolean {
```
fraszczakszymon: s/cls/className

jacekbj: fixed

PR #310, opened 2019-02-21
ADEN-8382 Fix FV lineItem tracking

dmnsobczak:


src/ad-engine/video/vast-parser.ts
```diff
@@ -9,7 +10,7 @@ class VastParser {
 		let value = '';
 
 		for (i = 0; i < possibleValues.length; i += 1) {
-			if (!isNaN(possibleValues[i])) {
+			if (possibleValues[i] && Number.isInteger(parseInt(possibleValues[i]))) {
```
bckatarzyna: Why not `!isNaN(parseInt(possibleValues[i]))`?

dmnsobczak: I don't trust `isNaN` after this :D Yeah, but you right, changed!

PR #311, opened 2019-02-22
ADEN-8407 Refactor EventService.

jacekbj:


src/ad-services/index.ts
```diff
@@ -1,6 +1,6 @@
-export * from './bill-the-lizard';
+export { billTheLizard, BillTheLizard, events as billTheLizardEvents } from './bill-the-lizard';
 export * from './confiant';
 export * from './geo-edge';
 export * from './krux';
-export * from './moat-yi';
+export { moatYi, events as moatYiEvents } from './moat-yi';
```
fraszczakszymon: What do you think about naming it `moatYiEvents` within `moat-yi/index.ts`? It will simplify all reexports

src/ad-services/index.ts
```diff
@@ -1,6 +1,6 @@
-export * from './bill-the-lizard';
+export { billTheLizard, BillTheLizard, events as billTheLizardEvents } from './bill-the-lizard';
```
fraszczakszymon: Same here

src/ad-products/tracking/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './video/jwplayer-tracker';
 export * from './video/porvata-tracker';
 export * from './video/twitch-tracker';
+export { events as playerEvents } from './video/player-event-emitter';
```
fraszczakszymon: Same here

examples/bidders/a9/script.ts
```diff
@@ -1,5 +1,10 @@
+<<<<<<< HEAD
```
fraszczakszymon: Fix me please

src/ad-engine/global.d.ts
```diff
@@ -0,0 +1,19 @@
+///<reference path="../../node_modules/@alugha/ima/typings/ima.d.ts"/>
```
Bielik20: What about doing it like that? https://github.com/Microsoft/TypeScript-Node-Starter#what-if-a-library-isnt-on-definitelytyped

Global types are discouraged as stated in here: https://stackoverflow.com/questions/42233987/how-to-configure-custom-global-interfaces-d-ts-files-for-typescript

tsconfig.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"outDir": "./dist/",
 		"sourceMap": true,
-		"module": "commonjs",
+		"module": "es2015",
```
Bielik20: Then this wouldn't be necessary, right?

src/ad-services/index.ts
jacekbj: Done

src/ad-services/index.ts
jacekbj: Done

src/ad-products/tracking/index.ts
jacekbj: Done

examples/bidders/a9/script.ts
jacekbj: Done

src/ad-engine/global.d.ts
jacekbj: I couldn't get that work. Can you please experiment and add a commit on top of my work?

Bielik20: Yeah, it was just a suggestion. I will experiment but I don't think it should block your PR.

PR #312, opened 2019-02-25
Aden 8206 TSLint

Bielik20:


examples/bidders/a9/script.ts
```diff
@@ -1,5 +1,5 @@
+import { Apstag, bidders, cmp } from '@wikia/ad-bidders';
```
fraszczakszymon: 🎉

spec/ad-engine/video/player/porvata/ima/google-ima.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* tslint:disable:function-name */
```
fraszczakszymon: What do you think about adding it to spec/tslint.json?

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -186,20 +186,23 @@ export class GptProvider {
...
-					if (!slotId) {
-						logger(logGroup, 'destroySlots', "slot doesn't return element id", slot);
-					} else if (slotNames.indexOf(slotId) > -1) {
-						return true;
-					}
-
-					return false;
-				  })
-				: allSlots;
+		let slotsToDestroy;
```
fraszczakszymon:
```suggestion
		let slotsToDestroy = allSlots;
```
by default? (without else statement)

spec/ad-engine/video/player/porvata/ima/google-ima.spec.ts
Bielik20: This is actually not necessary (good catch 👍 ) because of `// tslint:disable-next-line` latter in the files. Unfortunately the second one is required.

PR #315, opened 2019-02-27
ADEN-8429 Finish first call queue if there are no first call slots en…

jacekbj:
…abled.


examples/slots/disabled-first-call/script.ts
```diff
@@ -0,0 +1,11 @@
+import { AdEngine, btfBlockerService, context } from '@wikia/ad-engine';
+import { SlotConfig } from '../../../src/ad-engine/models';
```
fraszczakszymon: Not needed

examples/slots/disabled-first-call/script.ts
```diff
@@ -0,0 +1,11 @@
+import { AdEngine, btfBlockerService, context } from '@wikia/ad-engine';
```
fraszczakszymon: Not needed `btfBlockerService`

examples/slots/disabled-first-call/script.ts
fraszczakszymon: Shouldn't lint catch that?

PR #317, opened 2019-03-01
ADEN-8383 Add option to disable onScrollPush for SlotInjector

dmnsobczak:


src/ad-engine/services/slot-injector.ts
```diff
@@ -16,19 +16,22 @@ function findNextSuitablePlace(anchorElements = [], conflictingElements = []) {
...
+	if (!disablePushOnScroll) {
+		context.push('events.pushOnScroll.ids', slotName);
+	}
 
 	return container;
 }
 
 class SlotInjector {
-	inject(slotName, insertBelowScrollPosition = false) {
+	inject(slotName, insertBelowScrollPosition = false, disablePushOnScroll = false) {
```
fraszczakszymon: I believe we could get both `insertBelowScrollPosition` and `disablePushOnScroll` directly from config which is available in the line below

dmnsobczak: You right, changed.

PR #318, opened 2019-03-01
ADEN-8464 Add VMG adapter

fraszczakszymon:


src/ad-bidders/prebid/adapters/vmg.ts
```diff
@@ -0,0 +1,25 @@
+import { BaseAdapter } from './base-adapter';
+
+export class Vmg extends BaseAdapter {
+	constructor(options) {
```
jacekbj: Please replace constructor with:
```
bidderName = 'vmg';
```

PR #320, opened 2019-03-01
Add some utils types

Bielik20:
As far as TS in general is concerned. I have came up with some "guidelines" on how to use types. Unfortunately I wasn't able to find a rule that would do this automatically.

### Guidelines Proposition

```ts
// BAD (this hurts my eyes to read)
let labels: Map<string, string> = new Map<string, string>();
// GOOD (type is obvious)
let labels = new Map<string, string>();
```

```ts
// BAD (in a diff, it's not obvious what this type is)
let labels = this.buildLabels();
// GOOD
let labels: Map<string, string> = this.buildLabels();
```

```ts
// BAD - IDE kind of looses types with that syntax
array.map((x) => x * 2);
// GOOD - way better, it is obvious what we are dealing with
array.map((x: number) => x * 2);
```

```ts
// GOOD
array.map((x = 2) => x * 2);
// BAD
array.map((x: number = 2) => x * 2);
```

```ts
// GOOD? - there is no good type for reject or resolve
// Or maybe there is?
new Promise<void>((resolve, reject) => ...)
```
### Plus Actual Rule

```json
"typedef": {
	"severity": "warning",
	"options": ["call-signature"]
},
```
```ts
// BAD
function noop() {
    return;
}
// GOOD
function noop(): void {
    return;
}
```

### No Implicit Any

Additionally I strongly encourage everyone to fix types with "noImplicitAny" option set to true. You set it in `tsconfig.json`:

```json
{
	"compilerOptions": {
                 ...
		"noImplicitAny": true,
                 ...
	},
}
```

We do not have this turned on right now, because it will almost double a number of TypeScript errors. Still if you are working with types, you should turn it on at least temporarily.

## Summary

I would like us to discuss this proposition as a "general" rule for ad-engine types.


src/ad-engine/utils/document.ts
```diff
@@ -23,5 +23,5 @@ export function getDocumentVisibilityStatus() {
 			status = VISIBILITY_STATUS.notImplemented;
 	}
 
-	return status;
+	return status as VisibilityStatusType;
```
jacekbj: Asserting type (I think) can be avoided by defining `VisibilityStatusType` the following way:
```type VisibilityStatusType = VISIBILITY_STATUS.hidden | VISIBILITY_STATUS.visible | VISIBILITY_STATUS.notImplemented```. Am I right?

File: src/ad-engine/utils/flow-control.ts [hunk 2]
```diff
@@ -20,10 +22,11 @@ export const defer = (fn, ...args) =>
 		setTimeout(() => resolve(fn(...args)), 0);
 	});
 
-export function once(emitter, eventName, options = {}) {
-	const isObject = typeof emitter === 'object';
-	const hasAddEventListener = isObject && typeof emitter.addEventListener === 'function';
-	const hasOnce = isObject && typeof emitter.once === 'function';
+export function once(emitter: EventEmitter, eventName: string, options = {}): Promise<any> {
```
jacekbj: `emitter` can be i.e. `window` object. Are you sure it's not raising any errors?

jacekbj: One solution would be to set type as `emitter: EventEmitter | Window`.

src/global-types/current-device.d.ts
```diff
@@ -0,0 +1,55 @@
+declare module 'current-device' {
```
jacekbj: open source is waiting for your PR!
https://github.com/matthewhudson/current-device/issues/181

src/ad-engine/utils/document.ts
Bielik20: Cool idea, but it doesn't work... at least for me.

PR #327, opened 2019-03-07
ADEN-8426 Pass campaignId to slot listeners

dmnsobczak:


fraszczakszymon:
Would you please expose it to DOM too?
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/slot-data-params-updater.ts#L19

PR #328, opened 2019-03-07
ADEN-8485 Add default values for context.get

Bielik20:


fraszczakszymon:
Discussed offline: we decided to extend context.get method so it takes second argument -- `defaultValue = null`.

PR #331, opened 2019-03-11
Annotate AdSlot class.

jacekbj:


jacekbj:
Synthetic tests pass 👍

PR #333, opened 2019-03-12
ADEN-8495 Possibility to use Prebid logic for targeting params

fraszczakszymon:
Breaking changes:
* rubicon display configuration key changed from `rubiconDisplay` to `rubicon_display`
* changed interface of `getBidByAdId` method


src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -1,18 +1,57 @@
...
+		};
+	});
+
+	return adaptersAdServerTargeting;
+}
+
+export function getSettings(): PrebidSettings {
 	return {
 		standard: {
-			alwaysUseBid: false,
```
fraszczakszymon: This flag isn't working on `standard` level anymore

src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -30,19 +69,9 @@ export function getSettings() {
 					key: 'hb_size',
 					val: ({ size }) => size,
 				},
-				{
-					key: 'hb_uuid',
```
fraszczakszymon: I decided to remove `hb_uuid` from this place because it's misleading. We have to get this param as late as possible. More info: https://github.com/Wikia/ad-engine/pull/333/files#diff-e3dad39a0ac5f69980e6bb79928a85f1R222

src/ad-bidders/prebid/index.ts
```diff
@@ -51,7 +52,7 @@ export class Prebid extends BaseBidder {
 			debug:
 				utils.queryString.get('pbjs_debug') === '1' ||
 				utils.queryString.get('pbjs_debug') === 'true',
-			enableSendAllBids: true,
```
fraszczakszymon: It seems it wasn't used at all - we weren't using `pbjs.getAdserverTargeting*` methods in code. In fact we don't want to send all bids key-values.

examples/bidders/prebid/script.ts
```diff
@@ -20,7 +20,7 @@ const biddersDelay: DelayModule = {
 		}),
 };
 
-context.set('options.maxDelayTimeout', 1000);
+context.set('options.maxDelayTimeout', 2000);
```
jacekbj: I think this value could be exported from ad-engine as `DEFAULT_MAX_DELAY` some day

spec/ad-bidders/prebid/adapters/wikia-video.spec.ts
```diff
@@ -90,7 +90,7 @@ describe('WikiaVideo bidder adapter', () => {
 		const mocks = getMocks();
 
 		sinon.stub(global, 'setTimeout').callsFake(mocks.setTimeout);
-		sinon.stub(wikiaVideo, 'getVastUrl').returns(mocks.fakeVastUrl);
+		sinon.stub(WikiaVideo, 'getVastUrl').returns(mocks.fakeVastUrl);
```
jacekbj: 🙈

spec/ad-bidders/prebid/index.spec.ts
```diff
@@ -11,4 +13,19 @@ describe('Prebid bidder', () => {
 	it('can be initialized', () => {
 		new Prebid(bidderConfig);
 	});
+
+	it('returns all pbjs keys to reset', () => {
```
jacekbj: Please structure the test make it really tell what it tests:
```
describe('getTargetingKeys', () => {
    it('returns all pbjs keys to reset', () => {
        ...
    });
});
```

This will give a nice test summary:
`Prebid bidder getTargetingKeys  returns all pbjs keys to reset`

src/ad-bidders/prebid/adapters/audience-network.ts
```diff
@@ -1,15 +1,21 @@
 import { utils } from '@wikia/ad-engine';
-import { BaseAdapter } from './base-adapter';
+import { AdUnitConfig, BaseAdapter } from './base-adapter';
 
 export class AudienceNetwork extends BaseAdapter {
+	static bidderName = 'audienceNetwork';
+	testMode: boolean;
```
jacekbj: Please use a default value for a boolean properties - this will allow confusion resulting from the variable having 3 states instead of 2.

src/ad-bidders/prebid/adapters/rubicon-display.ts
```diff
@@ -1,18 +1,25 @@
 import { getTargeting } from '../prebid-helper';
-import { BaseAdapter } from './base-adapter';
+import { AdUnitConfig, BaseAdapter } from './base-adapter';
 
 export class RubiconDisplay extends BaseAdapter {
+	static bidderName = 'rubicon_display';
+	accountId: number;
+	aliases: { [key: string]: string[] };
```
jacekbj: We could have an interface for `Aliases`.

src/ad-bidders/prebid/adapters/wikia-video.ts
```diff
@@ -1,16 +1,30 @@
 import { buildVastUrl, context, utils } from '@wikia/ad-engine';
-import { BaseAdapter } from './base-adapter';
+import { AdUnitConfig, BaseAdapter } from './base-adapter';
 
 const price = utils.queryString.get('wikia_video_adapter');
```
jacekbj: Can these parameters go into constructor?

src/ad-bidders/prebid/adapters/wikia-video.ts
```diff
@@ -36,35 +54,28 @@ export class WikiaVideo extends BaseAdapter {
 		};
 	}
 
-	getSpec() {
+	getSpec(): { [key: string]: string | string[] } {
```
jacekbj: What does it mean `getSpec`?

src/ad-bidders/prebid/adapters/wikia.ts
```diff
@@ -1,16 +1,50 @@
 import { utils } from '@wikia/ad-engine';
-import { BaseAdapter } from './base-adapter';
+import { AdUnitConfig, BaseAdapter } from './base-adapter';
 
 const price = utils.queryString.get('wikia_adapter');
```
jacekbj: ditto

src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -1,18 +1,56 @@
...
 import { transformPriceFromBid } from './price-helper';
 
-const dfpVideoBidders = [
-	{ bidderCode: 'appnexusAst', contextKey: 'custom.appnexusDfp' },
-	{ bidderCode: 'beachfront', contextKey: 'custom.beachfrontDfp' },
-	{ bidderCode: 'lkqd', contextKey: 'custom.lkqdDfp' },
-	{ bidderCode: 'rubicon', contextKey: 'custom.rubiconDfp' },
-	{ bidderCode: 'pubmatic', contextKey: 'custom.pubmaticDfp' },
-];
+type ValueFunction = (bidResponse: any) => string;
```
jacekbj: You don't need to define a type for that, it can be put directly in line 11. `ValueFunction` doesn't really tell me anything until I see how it's used.

src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -1,18 +1,56 @@
...
+
+export interface PrebidTargeting {
+	hb_adid?: string;
+	hb_bidder?: string;
+	hb_pb?: string;
+	hb_size?: string;
+	[key: string]: string | string[];
+}
+
+function createAdServerTargetingForDeals(): PrebidSettings {
```
jacekbj: s/PrebidSettings/PrebidSettings | undefined/

src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -1,18 +1,56 @@
...
+	const adaptersAdServerTargeting = {};
+
+	if (!context.get('bidders.prebid.useBuiltInTargetingLogic')) {
+		return;
+	}
+
+	Object.keys(adapters).forEach((key) => {
```
jacekbj: `Object.values(adapters).forEach((adapter) => {` but that's just sugar

jacekbj: What if there is more than just an adapter class exported from one of the adapter files?

src/ad-bidders/prebid/prebid-helper.ts
```diff
@@ -39,6 +40,16 @@ export function setupAdUnits(adaptersConfig, lazyLoad = 'off') {
...
+
+	if (bid.mediaType === videoType) {
+		return bid.videoCacheKey;
+	}
+
+	return 'disabled';
+}
+
 export function getBidByAdId(adId) {
 	if (!window.pbjs || typeof window.pbjs.getBidResponses !== 'function') {
```
jacekbj: Just curious - why do we check if `window.pbjs` is set here? Shouldn't this be checked only once in the entry point to Prebid bidder? Returning a null from this function will require us to handle type `Bid | null` in the rest of the code.

src/ad-bidders/prebid/prebid-settings.ts
jacekbj: If we passed adapters as arguments then this function could actually have a unit test.

jacekbj:
One solution to handling `static bidderName` is to create a map (`Map<string, BaseAdapter>`) mapping bidder names to bidder adapter classes in prebid/adapters/index.ts. This way we would not have to iterate over imports later in the code.

src/ad-bidders/prebid/adapters/wikia-video.ts
fraszczakszymon: Then this information will be lost between consecutive pvs on mobile-wiki 🙈

src/ad-bidders/prebid/adapters/wikia-video.ts
fraszczakszymon: It's required by Prebid when we need to create custom adapter (however it isn't really documented well). When it's not declared video bidder does not work:
image

src/ad-bidders/prebid/adapters/wikia.ts
fraszczakszymon: Ditto :)

src/ad-bidders/prebid/prebid-settings.ts
fraszczakszymon: Good catch, I protected it with checking property `bidderName`

src/ad-bidders/prebid/prebid-helper.ts
fraszczakszymon: This method is exported, which means that pbjs may not exist in some conditions. Yes, we should handle both `Bid` and `null` because null can be returned not only when there is no pbjs, but also when the `adId` is wrong

spec/ad-bidders/prebid/adapters/wikia-video.spec.ts
fraszczakszymon: It has been changed to static 👍

src/ad-bidders/base-bidder.ts
```diff
@@ -1,10 +1,10 @@
-import { context, utils } from '@wikia/ad-engine';
+import { context, DEFAULT_MAX_DELAY, utils } from '@wikia/ad-engine';
 
 /**
  * @abstract
  */
 export class BaseBidder {
-	constructor(name, bidderConfig, timeout = 2000) {
+	constructor(name, bidderConfig, timeout = DEFAULT_MAX_DELAY) {
```
jacekbj: 💎

src/ad-bidders/prebid/adapters/wikia-video.ts
jacekbj: getters maybe?

spec/ad-bidders/prebid/index.spec.ts
```diff
@@ -1,4 +1,6 @@
+import { expect } from 'chai';
 import { Prebid } from '../../../src/ad-bidders/prebid/index';
+import { context } from '../../../src/ad-engine/services/context-service';
```
Bielik20: I would consider imports in spec using `@wikia/ad-engine` etc.

PR #335, opened 2019-03-15
Add Utils Types and Dictionary

Bielik20:


src/ad-engine/utils/iframe-builder.ts
```diff
@@ -1,22 +1,23 @@
...
+		const iframe: HTMLIFrameElement = document.createElement('iframe');
 
-		iframe.height = 0;
-		iframe.width = 0;
-		iframe.border = '0px';
-		iframe.hspace = '0';
-		iframe.vspace = '0';
+		iframe.height = '0';
+		iframe.width = '0';
+		(iframe as any).border = '0px';
```
jacekbj: We should probably style these with CSS - these properties are deprecated :D

src/ad-engine/utils/logger.ts
```diff
@@ -7,7 +7,7 @@ if (debugGroup !== '') {
 	window.console.info('AdEngine debug mode - groups:', debugGroup === '1' ? 'all' : groups);
 }
 
-export function logger(logGroup, ...logValues) {
+export function logger(logGroup: string, ...logValues: any): void {
```
jacekbj: `...logValues: any` -> `...logValues: any[]`
tsc will complaint about it if we generate types from ad-engine and try to use it in another project

src/ad-engine/utils/query-string.ts
```diff
@@ -1,14 +1,18 @@
+interface StringMap {
```
jacekbj: `type StringMap = Dictionary<string>`?

jacekbj: No real benefit though from using Dictionary, so let's leave it as Interface.

src/ad-engine/utils/query-string.ts
```diff
@@ -1,14 +1,18 @@
...
+	[key: string]: string;
+}
+
 class QueryString {
-	getValues(input = null) {
-		const path = input || window.location.search.substr(1);
-		const queryParameters = {};
-		const queryString = path.split('&');
+	getValues(input?: string): StringMap {
```
jacekbj: return type is `StringMap | null`.

src/ad-engine/utils/query-string.ts
```diff
@@ -1,14 +1,18 @@
...
+		const path: string = input || window.location.search.substr(1);
+		const queryStringParameters: string[] = path.split('&');
+		const queryParameters: StringMap = {};
 
-		if (queryString === '') {
+		if (!queryStringParameters) {
```
jacekbj: It will never be falsy =` queryStringParameters` will always be an array.

src/ad-engine/utils/script-loader.ts
```diff
@@ -1,45 +1,45 @@
...
-	createScript(src, type = 'text/javascript', isAsync = true, node = null, parameters = {}) {
-		const script = document.createElement('script');
-
-		node =
+	createScript(
+		src: string,
+		type = 'text/javascript',
+		isAsync = true,
+		node: HTMLElement | string = null,
+		parameters = {},
```
jacekbj: We can use `Partial<HTMLScriptElement>`.

src/ad-engine/utils/try-property.ts
```diff
@@ -1,29 +1,21 @@
...
-	}
-
-	return null;
+		return property in obj;
+	});
 }
 
-export function tryProperty(obj, properties = []) {
-	const property = whichProperty(obj, properties);
+export function tryProperty(obj: {}, properties: string[] = []): boolean | undefined {
```
jacekbj: Please use `obj: Dictionary`

PR #340, opened 2019-03-15
Annotate floating ad.

jacekbj:


src/ad-engine/templates/floating-ad.ts
```diff
@@ -6,11 +8,13 @@ export class FloatingAd {
 		return 'floating-ad';
 	}
 
+	adSlot: AdSlot;
+
 	constructor(adSlot) {
```
Bielik20: public adSlot: AdSlot

src/ad-engine/templates/floating-ad.ts
```diff
@@ -1,3 +1,5 @@
+import { AdSlot } from '../models';
+
```
Bielik20: Remove line here.

src/ad-engine/templates/floating-ad.ts
```diff
@@ -33,10 +37,9 @@ export class FloatingAd {
 			end = containerOffset + container.offsetHeight - slotHeight;
 
 			start = containerOffset;
-			if (slotNode.previousElementSibling) {
-				start =
-					getTopOffset(slotNode.previousElementSibling) +
-					slotNode.previousElementSibling.offsetHeight;
+			const previousElementSibling = slotNode.previousElementSibling as HTMLElement | null;
```
Bielik20: const previousElementSibling: HTMLElement.

src/ad-engine/templates/floating-ad.ts
```diff
@@ -6,11 +8,13 @@ export class FloatingAd {
...
 
+	adSlot: AdSlot;
+
 	constructor(adSlot) {
 		this.adSlot = adSlot;
 	}
 
-	init() {
+	init(): void {
 		const slotNode = document.getElementById(this.adSlot.getSlotName());
```
Bielik20: type

src/ad-engine/templates/floating-ad.ts
jacekbj: Here we accept that it can be either `HTMLElement` or `null`. In next line we check its truthiness to make sure it's `HTMLElement`. Therefore I let it infer type from `slotNode.previousElementSibling`.

src/ad-engine/templates/floating-ad.ts
jacekbj: Why would I retype it? I think it's redundant, type is inferred from `document.getElementById(...)`.

src/ad-engine/templates/floating-ad.ts
jacekbj: Fixed.

src/ad-engine/templates/floating-ad.ts
jacekbj: Fixed.

PR #341, opened 2019-03-18
ADEN-8500 | Set likho local storage

bckatarzyna:


src/ad-engine/services/slot-tweaker.ts
```diff
@@ -133,6 +133,22 @@ export class SlotTweaker {
 		});
 	}
 
+	setLikhoStorage(likhoType: string): void {
+		const after24hTime = Date.now() + 24 * 3600;
+		const likhoStorage = JSON.parse(localStorage.getItem('likho')) || [];
```
jacekbj: Please define interface for the data we keep in localStorage and use it inside this method.

src/ad-engine/services/slot-tweaker.ts
```diff
@@ -133,6 +133,22 @@ export class SlotTweaker {
 		});
 	}
 
+	setLikhoStorage(likhoType: string): void {
```
jacekbj: Is SlotTweaker the right place to put this method?

PR #343, opened 2019-03-20
ADEN-8507 Add logic for below the fold slots injection

dmnsobczak:


src/ad-engine/services/slot-injector.ts
```diff
@@ -54,6 +61,10 @@ class SlotInjector {
 			return null;
 		}
 
+		if (dryRun) {
+			return true;
```
fraszczakszymon: Method returns now null|boolean|DOMElement.

src/ad-engine/services/slot-injector.ts
```diff
@@ -54,6 +61,10 @@ class SlotInjector {
 			return null;
 		}
 
+		if (dryRun) {
```
fraszczakszymon: As far as I understand `dryRun` might be not needed (?)
https://github.com/Wikia/app/pull/16842/files#r267385572

dmnsobczak: Right, reverted.

src/ad-engine/services/slot-injector.ts
dmnsobczak: N/A

PR #346, opened 2019-03-21
ADEN-8515 Fix blockadblock import and return promise on check

fraszczakszymon:


src/ad-engine/utils/client.ts
```diff
@@ -18,26 +20,35 @@ class Client {
...
+		return new Promise((resolve) => {
+			if (bab === null) {
+				if (typeof BlockAdBlock === 'undefined') {
+					resolve(true);
+					enabled();
+
+					return;
+				}
+
+				bab = new BlockAdBlock({
```
jacekbj: It's a good occasion, please move adblock initialization into a separate method.

src/ad-engine/utils/client.ts
```diff
@@ -18,26 +20,35 @@ class Client {
...
-			if (typeof AdBlockDetect === 'undefined' || typeof BlockAdBlock === 'undefined') {
-				if (enabled !== null) enabled();
-
-				return;
+	checkBlocking(enabled: () => {}, disabled: () => {}): Promise<{}> {
+		return new Promise((resolve) => {
+			if (bab === null) {
+				if (typeof BlockAdBlock === 'undefined') {
+					resolve(true);
+					enabled();
```
jacekbj: Please move `enabled` and `disabled` execution into `then` part of the returned promise.

src/ad-engine/utils/client.ts
```diff
@@ -1,4 +1,6 @@
-import AdBlockDetect from 'blockadblock';
+// blockadblock doesn't export anything meaningful
+// it sets blockAdBlock and BlockAdBlock properties on window
+import 'blockadblock';
 import currentDevice from 'current-device';
 
 let bab: BlockAdBlock = null;
```
jacekbj: Let's make life better by avoiding `null`.
Let's add a variable `isBabInitialized` and use it later so that instead of
```
			if (bab === null) {
```
there will be
```
			if (!isBabInitialized) {
```

Using null and undefined as state is very error-prone.
I know that it's not in the scope of this task but since we're already here...

PR #355, opened 2019-04-02
ADEN-8559 Pass context templates.bfaa.slotsToDisable to UAP init().

jacekbj:


examples/templates/abcd/script.ts
```diff
@@ -4,6 +4,11 @@ import customContext from '../../context';
 import '../../styles.scss';
 
 customContext.targeting.artid = '390';
+customContext.targeting.dmn = 'fandomcom';
+customContext.templates.bfaa = {
+	...BigFancyAdAbove.getDefaultConfig(),
```
fraszczakszymon: Is it needed? `template-service` does it for you:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/template-service.ts#L18

jacekbj: It didn't! Template was throwing errors i.e. "onInit is not a function". Sth we should check?

jacekbj: Mystery solved, thanks for help!

PR #356, opened 2019-04-02
ADEN-8562 Fix template slotName value

fraszczakszymon:


src/ad-engine/services/template-service.ts
```diff
@@ -26,13 +26,20 @@ class TemplateService {
...
 
 		if (!templates[name]) {
 			throw new Error(`Template ${name} does not exist.`);
 		}
 
+		// override params.slotName by splitting it by comma
+		// and picking first occurrence so we can rely on it
+		// in used creative templates
+		if (params && typeof params.slotName === 'string') {
+			params.slotName = params.slotName.split(',').shift();
```
fraszczakszymon: I had a mixed feeling about overriding incoming params. But on the other hand we've been touched by this issue a few times already and we always forget that slot may have `pos` key-val with many values (array which is passed here as comma separated string). Another approach would be to fix all existing creative templates so they handle slotName:
https://github.com/Wikia/dfp-creative-templates/blob/f785217f0e49b50119d12c7e46ed4693b1f5b64f/hivi-takeover-bfaa-dev.html#L260
like e.g. porvata3 does:
https://github.com/Wikia/dfp-creative-templates/blob/f785217f0e49b50119d12c7e46ed4693b1f5b64f/incontent-porvata-2.html#L51
But then we'll have to remember to do it in all new templates. So probably we'll forget it soon.

Feel free to advise better solution if you see one ;)

src/ad-engine/services/template-service.ts
```diff
@@ -26,13 +26,20 @@ class TemplateService {
...
 		logger(logGroup, 'Load template', name, slot, params);
 
 		if (!templates[name]) {
 			throw new Error(`Template ${name} does not exist.`);
 		}
 
+		// override params.slotName by splitting it by comma
+		// and picking first occurrence so we can rely on it
+		// in used creative templates
+		if (params && typeof params.slotName === 'string') {
```
jacekbj: if (params && (typeof params.slotName === 'string' || params.slotName instanceof String)) {

src/ad-engine/services/template-service.ts
```diff
@@ -26,13 +26,20 @@ class TemplateService {
 		templates[name] = template;
 	}
 
-	init(name: string, slot: AdSlot = null, params = {}): void {
+	init(name: string, slot: AdSlot = null, params: { [key: string]: any } = {}): void {
```
jacekbj: You can use new type `Dictionary` from ad-engine/models.

src/ad-engine/services/template-service.ts
nandy-andy: I agree. Cost of keeping such tricky logic here < cost of keeping updating all the templates and keeping it in mind. Do you think we can unit test this part as a documentation of this conversation? ;)

PR #357, opened 2019-04-04
ADEN-8580 Remove built in logic and grab hb_deal_* manually

fraszczakszymon:
## Changelog
- handle all prebid key-vals manually
  - removed `bidders.prebid.useBuiltInTargetingLogic` configuration

### DIFF
--- a/src/ad-bidders/prebid/index.ts
+++ b/src/ad-bidders/prebid/index.ts
@@ -175,6 +175,18 @@ export class Prebid extends BaseBidder {
 		return allTargetingKeys.filter((key) => key.indexOf('hb_') === 0);
 	}
 
+	getDealsTargetingFromBid(bid): PrebidTargeting {
+		const keyValueParis = {};
+
+		Object.keys(bid.adserverTargeting).forEach((key) => {
+			if (key.indexOf('hb_deal_') === 0) {
+				keyValueParis[key] = bid.adserverTargeting[key];
+			}
+		});
+
+		return keyValueParis;
+	}
+
 	/**
 	 * @inheritDoc
 	 */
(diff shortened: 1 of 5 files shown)


src/ad-bidders/prebid/prebid-settings.ts
```diff
@@ -23,10 +22,6 @@ export interface PrebidTargeting {
 export function createAdapterSpecificSettings(adaptersList): PrebidSettings | undefined {
 	const adaptersAdServerTargeting = {};
 
-	if (!context.get('bidders.prebid.useBuiltInTargetingLogic')) {
```
nandy-andy: Do we have a follow-up to remove `wgAdDriverPrebidBuiltInTargetingCountries` in future? :)

src/ad-bidders/prebid/index.ts
```diff
@@ -181,30 +181,42 @@ export class Prebid extends BaseBidder {
...
-
-				bids.forEach((param) => {
-					if (!bidParams) {
-						bidParams = param;
-					} else if (bidParams.cpm === param.cpm) {
-						bidParams = bidParams.timeToRespond > param.timeToRespond ? param : bidParams;
-					} else {
-						bidParams = bidParams.cpm < param.cpm ? param : bidParams;
+				// ... However we need to take care of all hb_deal_* keys manually then
+				Object.keys(param.adserverTargeting).forEach((key) => {
```
nandy-andy: We may extract it to a new method, so unit testing it will be easy. Won't it?

src/ad-bidders/prebid/prebid-settings.ts
fraszczakszymon: Not yet 🙈 I'll create one

PR #361, opened 2019-04-09
ADEN-8562 Clear Porvata/IMA objects on slot destroy

fraszczakszymon:


examples/templates/outstream/index.html
```diff
@@ -72,6 +72,7 @@
 					Without the wikiaVideo adapter the Direct VAST line item should be served (creative with a
 					rabbit).
 				</p>
+				<p><strong>Press "D" on keyboard in order to destroy outstream ad slot</strong></p>
```
jacekbj: It's not `D`, it's `d` ;)

PR #362, opened 2019-04-11
ADEN-8651 Check if slot should be always bidded

dmnsobczak:


jacekbj:
Please add this property to `SlotConfig` definition:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/models/ad-slot.ts#L24
and to README.

PR #366, opened 2019-04-12
ADEN-8582 Fix slot injecting around hidden conflicting elements

dmnsobczak:


src/ad-engine/utils/dimensions.ts
```diff
@@ -9,11 +9,20 @@ function getElementOffset(element: HTMLElement, offsetParameter: OffsetParameter
 	const elementWindow: WindowProxy = element.ownerDocument.defaultView;
 	let currentElement: HTMLElement = element;
 	let hideAgain = false;
+	const previousStyles = {
```
jacekbj: Please add type `const previousStyles: Partial< CSSStyleDeclaration> = {`.

dmnsobczak: Done, nice catch!

PR #368, opened 2019-04-15
ADEN-8539 Don't bid A9 for disabled slots

dmnsobczak:


src/ad-bidders/a9/index.ts
```diff
@@ -116,7 +115,25 @@ export class A9 extends BaseBidder {
...
+					return bidderAlias === slot.slotID && !slotService.getState(slotName);
+				});
+
+				const isDisabledByConfig = Object.keys(this.slots).some((slotName) => {
+					const slotId = this.slots[slotName].slotId;
+
+					return slotId && slotId === slot.slotID && !slotService.getState(slotName);
+				});
+
+				return slotService.getState(slot.slotName) && !isDisabledByAlias && !isDisabledByConfig;
```
fraszczakszymon:
```suggestion
				return !isDisabledByAlias && !isDisabledByConfig;
```
isn't sufficient?

src/ad-bidders/a9/index.ts
```diff
@@ -116,7 +115,25 @@ export class A9 extends BaseBidder {
 		return slotsNames
 			.map((slotName) => this.getSlotAlias(slotName))
 			.map((slotAlias) => this.createSlotDefinition(slotAlias))
-			.filter((slot) => slot !== null);
+			.filter((slot) => {
+				if (slot === null) {
+					return false;
+				}
+
+				const isDisabledByAlias = Object.keys(context.get('slots')).some((slotName) => {
```
fraszczakszymon: Can we keep it consistent and name it `isUsedAsAlias`? (with changed return value)

fraszczakszymon: We need to make sure that at least one alias is *enabled*. Not the opposite scenario. In this case we'll disable bidding for specific slot when there is at least one disabled slot with given alias.

fraszczakszymon: Can we extract it to separate method and name it `isUsedAsAlias`?

dmnsobczak: Outstanding catch! Reversed and extracted.

src/ad-bidders/a9/index.ts
dmnsobczak: Actually it's the same. `slotService` will return `true` for undefined slot by default, but yes: version above should be more clear. Changed.

PR #377, opened 2019-04-25
ADEN-8664 Increase indexExchange and wikia max cpm

fraszczakszymon:


src/ad-bidders/prebid/adapters-registry.ts
```diff
@@ -41,6 +41,10 @@ class AdaptersRegistry {
 		WikiaVideo,
 	];
 
+	getAdapter(bidderName: string): BaseAdapter {
```
jacekbj: I know that `BaseAdapter | undefined` is not required as return type right now but it's a good habit. If one day (and it will be a great day for sure 🍻 🎆 ) we enable "strictNullChecks" then this line will glow red.

PR #381, opened 2019-05-07
ADEN-8718 Remove default sticky time from StickyTLB template

dmnsobczak:


src/ad-products/templates/sticky-base.ts
```diff
@@ -92,12 +92,16 @@ export class StickyBase {
 	 * @protected
 	 */
 	addUnstickLogic() {
-		const { stickyAdditionalTime, stickyUntilSlotViewed } = this.config;
+		const { stickyAdditionalTime, stickyDefaultTime, stickyUntilSlotViewed } = this.config;
 		const whenSlotViewedOrTimeout = async () => {
+			const unstickTime =
+				stickyAdditionalTime +
+				(stickyDefaultTime !== undefined ? stickyDefaultTime : StickyBase.DEFAULT_UNSTICK_DELAY);
```
fraszczakszymon: What do you think about putting `StickyBase.DEFAULT_UNSTICK_DELAY` in the `sticky-ad.ts` default config. Then we can simplify above code to:
```suggestion
			const stickTime = stickyAdditionalTime + stickyDefaultTime;
```

dmnsobczak: Modified offline and changed, thanks!

PR #388, opened 2019-05-13
ADEN-8706 | Add AdSlotData interface

bckatarzyna:


src/ad-engine/listeners/slot-listener.ts
```diff
@@ -8,6 +8,21 @@ interface AdditionalEventData {
 	event: googletag.events.SlotRenderEndedEvent;
 }
 
+export interface AdSlotData {
```
Bielik20: Please if you can change it so that files are not optional. I know it is not said anywhere but if interface consists only of optional elements we tend to declare them as not optional and use something like: `Partial<AdSlotData>`

PR #391, opened 2019-05-14
ADEN-8774 | Add Browsi ad service

bckatarzyna:


src/ad-engine/utils/script-loader.ts
```diff
@@ -19,8 +19,16 @@ class ScriptLoader {
 		script.type = type;
 		script.src = src;
 
+		if (parameters.dataset) {
+			Object.keys(parameters.dataset).forEach((parameter) => {
```
dmnsobczak: I would do this in a single foreach but I have nothing against this approach ¯\_(ツ)_/¯

src/ad-services/browsi/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Injects browsi script
+ */
+function loadScript(): Promise<Event> {
+	const browsiLibraryUrl = `https://middycdn-a.akamaihd.net/bootstrap/bootstrap.js`;
+
+	return utils.scriptLoader.loadScript(browsiLibraryUrl, 'text/javascript', true, 'first', {
+		id: 'browsi-tag',
+		dataset: {
```
fraszczakszymon: Shouldn't it be configured in context?

src/ad-services/browsi/index.ts
```diff
@@ -0,0 +1,39 @@
+import { context, utils } from '@wikia/ad-engine';
+
+const logGroup = 'browsi';
+
+/**
+ * Injects browsi script
+ */
+function loadScript(): Promise<Event> {
+	const browsiLibraryUrl = `https://middycdn-a.akamaihd.net/bootstrap/bootstrap.js`;
```
fraszczakszymon: Shouldn't it be configured in context?

PR #395, opened 2019-05-21
ADEN-8791 Add option to force safe frame per slot

dmnsobczak:


src/ad-engine/providers/gpt-provider.ts
```diff
@@ -114,6 +114,7 @@ export class GptProvider implements Provider {
 		gptSlot.addService(window.googletag.pubads()).setCollapseEmptyDiv(true);
 
 		this.applyTargetingParams(gptSlot, targeting);
+		this.setSafeFrameStatus(gptSlot, adSlot);
```
fraszczakszymon:
```suggestion
		if (adSlot.getConfigProperty('forceSafeFrame')) {
			this.setSafeFrameStatus(gptSlot);	
		}
```
What do you think?

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -142,6 +143,15 @@ export class GptProvider implements Provider {
 		Object.keys(targeting).forEach((key) => gptSlot.setTargeting(key, targeting[key]));
 	}
 
+	setSafeFrameStatus(gptSlot: googletag.Slot, adSlot: AdSlot) {
```
fraszczakszymon: We can make this method `adSlot`-independent:
https://github.com/Wikia/ad-engine/pull/395/files#r285945976
What do you think?

src/ad-engine/providers/gpt-provider.ts
dmnsobczak: Why not, changed.

src/ad-engine/providers/gpt-provider.ts
dmnsobczak: Changed too.

PR #396, opened 2019-05-23
ADEN-8657 Improve async in StickyTLB.

jacekbj:


src/ad-engine/models/ad-slot.ts
```diff
@@ -230,10 +247,6 @@ export class AdSlot extends EventEmitter {
 		context.set(`slots.${this.config.slotName}.${key}`, value);
 	}
 
-	onLoad(): Promise<HTMLIFrameElement> {
```
fraszczakszymon: It is kind of breaking change, isn't it?

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -144,7 +144,14 @@ export class GptProvider implements Provider {
 	}
 
 	applyTargetingParams(gptSlot: googletag.Slot, targeting: Targeting) {
-		Object.keys(targeting).forEach((key) => gptSlot.setTargeting(key, targeting[key]));
+		Object.keys(targeting).forEach((key) => {
+			let value = targeting[key];
+
+			if (!Array.isArray(value)) {
+				value = value.toString();
```
fraszczakszymon: Why are we doing it? GPT seems to handle arrays properly

src/ad-engine/models/ad-slot.ts
jacekbj: It would be if `onLoad` was used anywhere apart from ad-engine.

src/ad-engine/providers/gpt-provider.ts
jacekbj: It seems to but documentation clearly states `string | string[]`
https://developers.google.com/doubleclick-gpt/reference#googletag.PassbackSlot_setTargeting

src/ad-products/templates/sticky-base.ts
```diff
@@ -89,28 +82,24 @@ export class StickyBase {
...
 		const {
 			stickyAdditionalTime,
 			stickyDefaultTime = StickyBase.DEFAULT_UNSTICK_DELAY,
 			stickyUntilSlotViewed,
 		} = this.config;
 		const whenSlotViewedOrTimeout = async () => {
-			await (stickyUntilSlotViewed && !this.adSlot.isViewed()
-				? utils.once(this.adSlot, AdSlot.SLOT_VIEWED_EVENT)
+			await (stickyUntilSlotViewed
+				? this.adSlot.loaded.then(() => this.adSlot.viewed)
```
fraszczakszymon: 💎

jacekbj: 🙇

src/ad-engine/models/ad-slot.ts
jacekbj: I agree that we can change the label to major. What do you think @fraszczakszymon?

PR #399, opened 2019-05-27
ADEN-8827 | Expose string builder for browsi

bckatarzyna:


src/ad-services/browsi/index.ts
```diff
@@ -29,6 +29,7 @@ class Browsi {
 		}
 
 		utils.logger(logGroup, 'loading');
+		window.buildAdUnitString = utils.stringBuilder.build;
```
Bielik20: Maybe
```ts
window.buildAdUnitString = (<arg1>, <arg2>) => utils.stringBuilder.build
```

Where arg1 and arg2 are names of the arguments. But I am not sure. You can leave as it is.

PR #400, opened 2019-05-27
ADEN-8807 Implement instant config service 🚀

fraszczakszymon:


src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,99 @@
...
+const overrideInstantConfig = (config: Config): Config => {
+	const newConfig = {};
+	const queryParams = utils.queryString.getValues();
+
+	Object.keys(queryParams)
+		.filter((instantGlobalKey: string) => {
+			return instantGlobalKey.indexOf(instantGlobalsQueryParamPrefix) === 0;
+		})
+		.map((instantGlobalKey) => {
+			const [, key] = instantGlobalKey.split('.');
```
jacekbj: I had no idea it can be done this way 👍

src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+			return key;
+		})
+		.forEach((key) => {
+			newConfig[key] = parseValue(queryParams[`${instantGlobalsQueryParamPrefix}.${key}`]);
```
jacekbj: If you return `[key, instantGlobalKey]` from previous `map` then you can use it to extract value from `queryParams`.

src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const logGroup = 'instant-config';
+const instantGlobalsQueryParamPrefix = 'InstantGlobals';
+
+type Variable = boolean | string | string[] | number | number[] | object | null;
+interface Config {
+	[key: string]: Variable;
+}
+
+const parseValue = (value: string): Variable => {
```
jacekbj: This one has to be exported in order to be covered by unit tests. Unless you plan to cover it with unit tests by testing `instantConfig` instance.

src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	try {
+		return JSON.parse(value);
+	} catch (ignore) {
+		return value || null;
+	}
+};
+
+const overrideInstantConfig = (config: Config): Config => {
```
jacekbj: ditto

tests/pages/instant-config.page.js
```diff
@@ -0,0 +1,22 @@
...
+		this.testQueryParamsButton = '#test-params';
+		this.breakConfigButton = '#break-config';
+	}
+
+	enableTestQueryParams() {
+		$(this.testQueryParamsButton).click();
+	}
+
+	enableBrokenConfigRequest() {
+		$(this.breakConfigButton).click();
```
jacekbj: BTW breaking the connection can be achieved with response interception in WDIO. This allows for testing multiple scenarios like timeout, internet disconnected and so on.
 
https://chromedevtools.github.io/devtools-protocol/tot/Network

tests/specs/services/instant-config.desktop.test.js
```diff
@@ -0,0 +1,41 @@
+import { expect } from 'chai';
+import { helpers } from '../../common/helpers';
+import { instantConfig } from '../../pages/instant-config.page';
+
+describe('It will test instant config page', () => {
```
jacekbj: When designing test cases please take into consideration another person reading the resulting test names. For example, with the design we have here the test names will be:
`It will test instant config page test example query param`
`It will test instant config page test fallback config`
`It will test instant config page test overriding config`

What do you think about:
```
describe('Instant Config', 
  it('should read query params from URL and override configuration', ...)
  describe('on broken connection', ...
    it('should use fallback config', ...)
    it('should override fallback config with query params', ...)
``` 
This would build names:
`Instant Config should read query params from URL and override configuration`
`Instant Config on broken connection should use fallback config`
`Instant Config on broken connection should override fallback config with query params`

fraszczakszymon: Absolutely makes sense! Thanks

src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,90 @@
...
+	config: Config = null;
+
+	async get(variableName: string, defaultValue: ConfigValue = null): Promise<ConfigValue> {
+		const config = await this.getConfig();
+
+		return config[variableName] || defaultValue;
+	}
+
+	async getConfig(): Promise<Config> {
+		if (this.config === null) {
```
jacekbj: `if (this.config === null)` -> `if (this.config)`?

src/ad-services/instant-config/index.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * InstantConfig service
+ */
+class InstantConfig {
+	config: Config = null;
+
+	async get(variableName: string, defaultValue: ConfigValue = null): Promise<ConfigValue> {
+		const config = await this.getConfig();
+
+		return config[variableName] || defaultValue;
```
jacekbj: What if value of `config[variableName]` is `0` or `False`?

tests/pages/instant-config.page.js
jacekbj: We don't have WDIO 5 Network fully configured.

src/ad-services/instant-config/index.ts
fraszczakszymon: 💎

PR #401, opened 2019-05-28
ADEN-8621 Adding GumGum Prebid.js adapter

dmnsobczak:


examples/slots/safe-frame/script.ts
```diff
@@ -1,5 +1,13 @@
-import { AdEngine, context, templateService } from '@wikia/ad-engine';
-import { FloorAdhesion } from '@wikia/ad-products';
+import { bidders } from '@wikia/ad-bidders';
```
fraszczakszymon: I would create separate test page for that so it is clear for testing prebid in safe frame, like `bidders/prebid-safe-frame` where we can force top_leaderboard to be SafeFrame. Now it slightly mixed with the floorAdhesion and its template

PR #404, opened 2019-05-29
ADEN-8842 AdSlot getTopOffset handles no slot element.

jacekbj:
If slot has no element, then it returns -1.

### DIFF
--- a/src/ad-engine/models/ad-slot.ts
+++ b/src/ad-engine/models/ad-slot.ts
@@ -225,8 +225,15 @@ export class AdSlot extends EventEmitter {
 		return JSON.parse(JSON.stringify(this.config));
 	}
 
+	/**
+	 * Returns offset of slot from top of the page.
+	 *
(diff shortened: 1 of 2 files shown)


src/ad-engine/models/ad-slot.ts
```diff
@@ -224,8 +224,15 @@ export class AdSlot extends EventEmitter {
...
+	/**
+	 * Returns offset of slot from top of the page.
+	 *
+	 * Returns -1 if slot has no element.
+	 */
 	getTopOffset(): number {
-		return getTopOffset(this.getElement());
+		const element = this.getElement();
+
+		return element ? getTopOffset(element) : -1;
```
fraszczakszymon: `-1` value might be misleading in case of using this method somewhere in the slot-related-logic (like getting top offset for viewport observer). Discussed on slack: we decided that it would be better to throw an exception here and handle it properly in the slot-tracker.

PR #405, opened 2019-05-30
ADEN-8761 | Add slotsSet configuration param to BFAA

bckatarzyna:


src/ad-products/templates/uap/universal-ad-package.ts
```diff
@@ -249,7 +250,12 @@ function updateSlotsTargeting(lineItemId, creativeId) {
 	});
 }
 
-function enableSlots(slotsToEnable) {
+function enableSlots(slotsToEnable: string[] = [], slotsSet: string = 'default'): void {
+	if (slotsSet === 'bfaa-only') {
+		window.ads.runtime.disableSecondCall = true;
+	} else if (slotsSet === 'bfaa-top_boxad') {
+		slotsToEnable = ['top_boxad'];
```
fraszczakszymon: In my opinion Wikia/ad-engine shouldn't be aware of any hardcoded slot name (like `top_boxad` here). It's related to the platform slots context configuration. Also what do you think about giving possibility to create some custom slot presets in the context (e.g. `slotPresets['bfaa-only']` with `[ 'top_leaderboard' ]` in the platform configuration)? Then those presets shouldn't be handled in the `universal-ad-package`, but in the refactored-`btf-blocker-service`.

I'd go further - I'd tear down whole `btf-blocker` in favour of new `ad-layout-manager` which would handle `slotsPresets` properly. Also it could map old `window.ads.runtime.disableSecondCall` to the appropriate preset.

Second part sounds like a big change, but maybe we should start considering that? Let's open discussion on #adeng-tech?

jacekbj:
We decided not to implement the slots-set function.

PR #408, opened 2019-05-31
ADEN-8841 Remove hide(), show() and getContainer() from slotTweaker.

jacekbj:


src/ad-engine/models/ad-slot.ts
```diff
@@ -351,4 +354,47 @@ export class AdSlot extends EventEmitter {
...
+	 *
+	 * Adds class AdSlot.HIDDEN_CLASS to adSlot's element.
+	 */
+	hide() {
+		const added = this.addClass(AdSlot.HIDDEN_CLASS);
+
+		if (added) {
+			logger(AdSlot.LOG_GROUP, 'hide', this.getSlotName());
+		}
+		this.emit(AdSlot.HIDDEN_EVENT);
```
fraszczakszymon: Should we emit event when `!added`?

src/ad-engine/models/ad-slot.ts
```diff
@@ -351,4 +354,47 @@ export class AdSlot extends EventEmitter {
...
+	 *
+	 * Removes class AdSlot.HIDDEN_CLASS from adSlot's element.
+	 */
+	show(): void {
+		const removed = this.removeClass(AdSlot.HIDDEN_CLASS);
+
+		if (removed) {
+			logger(AdSlot.LOG_GROUP, 'show', this.getSlotName());
+		}
+		this.emit(AdSlot.SHOWED_EVENT);
```
fraszczakszymon: Should we emit event when `!removed`?

src/ad-products/styles/_out-of-page.scss
```diff
@@ -38,6 +39,12 @@ $out-of-page-z-index: 5000101;
 	bottom: 0;
 	// needed for Safari 9 - it doesn't position elements from the left by default
 	left: 0;
+	text-align: center;
+
+	.advertisement-label {
```
fraszczakszymon: Are we adding label to floor adhesion?

src/ad-engine/models/ad-slot.ts
jacekbj: IMO returning a boolean from this function means depending on DOM for preserving component state.
I'd rather see code like that:
```
private isVisible = false;
get visible(): boolean { return this.isVisible; }
set visible(value: boolean) {
  if (value !== this.isVisible) {
    this.isVisible = value;
    this.emit(value ? AdSlot.SHOWED_EVENT : AdSlot.HIDDEN_EVENT);
    if (value) {
      this.removeClass('hidden');
    else {
      this.addClass('hidden');
    }
}
```

src/ad-products/styles/_out-of-page.scss
jacekbj: I need @nefretete confirmation on that, good point.

PR #409, opened 2019-06-04
ADEN-8784 Move injecting slots to slotInjector.

jacekbj:
❗️ By mistake I based the code on ADEN-8841 branch.
Once ADEN-8841 is merged, change target of this PR to `dev`.

I add label "Major" because a public function is removed.


src/ad-engine/listeners/scroll-listener.ts
```diff
@@ -33,15 +33,13 @@ export class ScrollListener {
...
 	 *
-	 * @param adStack AdStack to push the slot to.
 	 * @param id ID of the AdSlot to push
 	 * @param threshold slot will be pushed `threshold`px before it appears in viewport
 	 * @param distanceFromTop slot will be pushed after scrolling `distanceFromTop`px
 	 *
 	 * Only one parameter can be supplied: threshold or distanceFromTop
 	 */
 	addSlot(
-		adStack: OldLazyQueue<AdStackPayload>,
```
Bielik20: What about `extensions/wikia/AdEngine3/src/slots.js`?

```
injectFloorAdhesion() {
	scrollListener.addSlot(
			context.get('state.adStack'),
			'floor_adhesion',
			{ distanceFromTop: utils.getViewportHeight() },
	);
},
```

README.md
```diff
@@ -65,7 +67,7 @@ To compile it in your desired application import `getAdEngineLoader` from `confi
 |`slots.{slot_name}.avoidConflictWith`|CSS selector that is going to be checked to prevent loading ad slot in the same viewport|string|✘|
 |`slots.{slot_name}.bidderAlias`|Ad slot name alias for getting bids that are assigned for different ad slot|string|✘|
 |`slots.{slot_name}.insertBelowScrollPosition`|Switch: insert ad slot below scroll|boolean|✘|
-|`slots.{slot_name}.insertBeforeSelector`|CSS selector where to put ad slot when it is going to be created once another slot is created (`events.pushAfterRendered` is required)|string|✘|
+|`slots.{slot_name}.insertBeforeSelector`|CSS selector where to put ad slot when it is going to be created once another slot is created (`events.pushAfterRendered` or `events.pushAfterCreated` is required)|string|✘|
```
Bielik20: Off top, but we could transform it into an interface in ad-engine.

src/ad-engine/ad-engine.ts
```diff
@@ -20,6 +20,10 @@ export interface AdStackPayload {
 	id: string;
 }
 
+export function getAdStack(): OldLazyQueue<AdStackPayload> | undefined {
```
Bielik20: I think `adStack` starts as an empty array at least in app.

src/ad-engine/models/ad-slot.ts
```diff
@@ -397,4 +397,14 @@ export class AdSlot extends EventEmitter {
...
+
+	/**
+	 * Pass all events through eventService before emitting directly from slot.
+	 */
+	emit(event: string | symbol, ...args: any[]): boolean {
+		const result = super.emit(event, ...args);
+
+		eventService.emit(event, this);
+		return result;
+	}
```
Bielik20: I understand the premise, but having the same event emitted in two places, isn't it confusing?

src/ad-engine/services/slot-injector.ts
```diff
@@ -38,7 +41,35 @@ function insertNewSlot(
...
 
 class SlotInjector {
-	inject(slotName: string): HTMLElement | null {
+	constructor() {
+		/**
+		 * I did not create an automated way to build this code from event name
+		 * to config key map because Symbol cannot be iterated over using any for loop
+		 */
+		eventService.on(events.AD_SLOT_CREATED, (adSlot) => {
+			const slotsToPush = context.get(`events.pushAfterCreated.${adSlot.getSlotName()}`) || [];
```
Bielik20: I would put it in private method with a name.

src/ad-engine/services/slot-injector.ts
```diff
@@ -38,7 +41,35 @@ function insertNewSlot(
...
+				if (slotElement) {
+					slotService.pushSlot(slotElement);
+				} else {
+					logger(logGroup, `Could not inject slot ${slotName}.`);
+				}
+			});
+		});
+
+		eventService.on(AdSlot.SLOT_RENDERED_EVENT, (adSlot) => {
+			const slotsToPush = context.get(`events.pushAfterRendered.${adSlot.getSlotName()}`) || [];
```
Bielik20: dito

src/ad-engine/services/slot-injector.ts
```diff
@@ -68,7 +99,9 @@ class SlotInjector {
 			return null;
 		}
 
-		const disablePushOnScroll = config.repeat ? !!config.repeat.disablePushOnScroll : false;
+		if (disablePushOnScroll === undefined) {
```
Bielik20: `typeof disablePushOnScroll !== boolean` - safer

src/ad-engine/listeners/scroll-listener.ts
jacekbj: Fixed https://github.com/Wikia/app/pull/17120/commits/24944f3ce49952a5e7ca2338bc807f6aaf430720

src/ad-engine/models/ad-slot.ts
jacekbj: There is no single point where events from all slots are available.
You can use `slotService.on` but it requires prior knowledge of the slot name.
`slotService` itself does not inherit from `EventEmitter`.

src/ad-engine/services/slot-injector.ts
jacekbj: Good point, will make it easier to refactor later.

jacekbj: Refactored. Please check out the name.

src/ad-engine/services/slot-injector.ts
jacekbj: Refactored!

src/ad-engine/ad-engine.ts
jacekbj: Yes but well, how do I know :|

Bielik20: Yeah, so maybe `unknown` instead of `undefined`?

src/ad-engine/services/slot-injector.ts
```diff
@@ -47,7 +47,10 @@ class SlotInjector {
 		 * to config key map because Symbol cannot be iterated over using any for loop
 		 */
 		eventService.on(events.AD_SLOT_CREATED, (adSlot) => {
```
Bielik20: - adSlot type
- What about putting entire callback into function. To describe what is suppose to happen when this event occur.

src/ad-engine/services/slot-injector.ts
```diff
@@ -61,14 +64,24 @@ class SlotInjector {
 		});
 
 		eventService.on(AdSlot.SLOT_RENDERED_EVENT, (adSlot) => {
```
Bielik20: dito

src/ad-engine/services/slot-injector.ts
jacekbj: I added adSlot type.
Ad. 2: I wanted to avoid method named `pushSlotsAfterCreatedCallback`, that's why I did not name them. I simplified the code a bit by moving logic to `AdSlot`.

src/ad-engine/ad-engine.ts
jacekbj: I removed unknown because it requires casting to lazy-queue everywhere.

PR #411, opened 2019-06-06
ADEN-8768 Change Mobile-Wiki build to use Awesome TypeScript Loader

Bielik20:


src/ad-engine/listeners/listeners.ts
```diff
@@ -37,5 +37,3 @@ export interface VideoEventListener {
 
 	onEvent(eventName: string, params: Dictionary, data: VideoData): void;
 }
-
-export {}; // tslint no-sole-types fix
```
dmnsobczak: 💎

PR #413, opened 2019-06-06
ADEN-8806 Map array from query string to strings.

jacekbj:


src/ad-products/templates/sticky-base.ts
```diff
@@ -25,6 +25,7 @@ export class StickyBase {
 		this.container = this.adSlot.getElement();
 		this.lineId = adSlot.lineItemId.toString() || '';
 		this.lines = context.get(`templates.${this.getName()}.lineItemIds`) || [];
+		this.lines = this.lines.map((el) => el.toString());
```
dmnsobczak: Why split it into two lines? Eg:
```
this.lines = context
    .get(`templates.${this.getName()}.lineItemIds`)
    .map((el) => el.toString()) || [];
```

jacekbj:
```
context.get(`templates.${this.getName()}.lineItemIds`)
``` can have value `null`, which does not have method `map`. Therefore, that would throw an error.

PR #414, opened 2019-06-10
ADEN-8883 AdSlot is now viewed when either AdSlot.SLOT_VIEWED_EVENT o…

jacekbj:
…r AdSlot.VIDEO_VIEWED_EVENT is emitted.

### DIFF
--- a/src/ad-engine/models/ad-slot.ts
+++ b/src/ad-engine/models/ad-slot.ts
@@ -92,6 +92,7 @@ export class AdSlot extends EventEmitter {
 	});
 	viewed = new Promise<void>((resolve) => {
 		this.once(AdSlot.SLOT_VIEWED_EVENT, resolve);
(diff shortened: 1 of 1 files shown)


src/ad-engine/models/ad-slot.ts
```diff
@@ -92,6 +92,7 @@ export class AdSlot extends EventEmitter {
 	});
 	viewed = new Promise<void>((resolve) => {
 		this.once(AdSlot.SLOT_VIEWED_EVENT, resolve);
+		this.once(AdSlot.VIDEO_VIEWED_EVENT, resolve);
```
fraszczakszymon: Doubts: why slot viewed event isn't dispatched properly. Discussed on slack: let's go with this hotfix and then figure out solution for `SLOT_VIEWED_EVENT` event

PR #415, opened 2019-06-10
ADEN-8871 Adding overscrolled event

dmnsobczak:


src/ad-engine/models/ad-slot.ts
```diff
@@ -284,6 +293,23 @@ export class AdSlot extends EventEmitter {
 		}
 	}
 
+	watchForOverscrolled() {
```
fraszczakszymon: I have mixed feelings about putting it in the ad-slot model. What about slot-service/scroll-listener/some new module?

dmnsobczak: I did it at the beginning but after simplifying the code I thought that this is a bot overkill. But maybe you right, I will move it there :)

src/ad-engine/listeners/overscroll-listener.ts
```diff
@@ -0,0 +1,42 @@
+import { AdSlot, Dictionary } from '../models';
+import { viewportObserver } from '../utils';
+
+interface OverscrollListenedElement {
+	wasInViewport: boolean;
+	overscrolledListener: string;
+}
+
+class OverscrollListener {
```
jacekbj: It's already post CR so I will ask for a comment instead of unit tests to describe the logic which decides when `'overscrolled'` event is emitted.

post mortem:
The string `'overscrolled'` could be exported so that other apps can use it.

src/ad-engine/listeners/overscroll-listener.ts
```diff
@@ -0,0 +1,42 @@
...
+					return;
+				}
+
+				if (inViewport) {
+					listenerConfig.wasInViewport = inViewport;
+					return;
+				}
+
+				viewportObserver.removeListener(listenerConfig.overscrolledListener);
+				adSlot.emitEvent('overscrolled');
```
jacekbj: I still wonder what was first:
* `AdSlot` inherits from `EventEmitter`,
* this method is added.
:D

PR #427, opened 2019-06-26
ADEN-8902 | Get geo codes from context

bckatarzyna:


src/ad-engine/utils/geo.ts
```diff
@@ -151,46 +151,25 @@ function containsEarth(countryList: string[], name: string): boolean {
...
-function setGeoData(data: GeoData): void {
-	geoData = data;
-}
-
 /**
  * Return country code based on cookie
  */
 function getCountryCode(): string {
-	return (getGeoData() as GeoData).country;
+	return context.get('geo.country');
```
jacekbj: It can also return `undefined`. Please extend the return type.

spec/ad-engine/utils/geo.spec.ts
```diff
@@ -2,17 +2,15 @@ import { geoService } from '@wikia/ad-engine/utils';
...
 	let sandbox;
 
 	beforeEach(() => {
-		geoService.setGeoData({
-			city: 'Poznan',
-			country: 'PL',
-			continent: 'EU',
-			region: '72',
-		});
+		context.set('geo.continent', 'EU');
```
jacekbj: Please add a tests which test that `getCountryCode`, `getContinentCode`, `getRegionCode ` and `isProperGeo ` work if context `geo` part is not set.

bckatarzyna: Cases added

PR #436, opened 2019-07-04
ADEN-8994 Fix replaying closed UAP

dmnsobczak:


src/ad-products/templates/uap/themes/hivi/hivi-bfaa.ts
```diff
@@ -325,7 +325,9 @@ export class BfaaHiviTheme extends BigFancyAdHiviTheme {
 
 		if (stopVideo) {
 			if (this.video && this.video.ima.getAdsManager()) {
-				this.video.stop();
+				if (this.video.isPlaying()) {
```
nandy-andy: Why are we checking if video is playing within `if (stopVideo)`. Three nested `if`s hurt my eyes as well 🙈 And this feedback also hurts me because I don't have any idea how to make it better :/

nandy-andy: @jacekbj explained me `stopVideo` doesn't mean what I thought it meant ;)

nandy-andy:
Let's move the changes to THE ONE BRANCH: https://github.com/Wikia/app/tree/uap-fixes ;)

PR #441, opened 2019-07-08
ADEN-9021 Add tests for price-helper.ts.

jacekbj:


spec/ad-bidders/prebid/price-helper.spec.ts
```diff
@@ -38,3 +47,130 @@ describe('transformPriceFromCpm', () => {
...
+
+		expect(result).to.equal('20.00');
+	});
+
+	it('should round price to adapter.maxCpm if present', () => {
+		// maxCpm must be greater than 20.
+		let result;
+		adapter.maxCpm = 10;
+
+		result = transformPriceFromBid(PrebidBidFactory.getBid({ cpm: 20.01 }));
```
emadry: Please add case for cpm below 20

spec/ad-bidders/prebid/price-helper.spec.ts
```diff
@@ -38,3 +47,133 @@ describe('transformPriceFromCpm', () => {
...
+		bids = [bid];
+
+		const result = getPrebidBestPrice(bidderName);
+
+		expect(result).to.deep.equal({ [bidderName]: '1.00' });
+	});
+
+	it('should not take rendered bids into consideration', () => {
+		adapters.set(bidderName, { bidderName } as BaseAdapter);
+		const bid = PrebidBidFactory.getBid({ bidderCode: bidderName, status: 'rendered' });
```
Bielik20: I would add cpm value in here just to be sure we are testing rendered, not "no price".

spec/ad-bidders/prebid/price-helper.spec.ts
```diff
@@ -38,3 +47,133 @@ describe('transformPriceFromCpm', () => {
...
+		bids = [bid];
+
+		const result = getPrebidBestPrice(bidderName);
+
+		expect(result).to.deep.equal({ [bidderName]: '' });
+	});
+
```
Bielik20: It is me being paranoid but I would also include case which is mix of 

* should return empty string if there is no price for bidder
* should round to 2 decimal places
* should not take rendered bids into consideration <- with that having higher cpm than previous

So that we are sure that it only drops the ones that it is suppose to drop and leaves the rest.

spec/ad-bidders/prebid/price-helper.spec.ts
```diff
@@ -38,3 +47,133 @@ describe('transformPriceFromCpm', () => {
...
+		sandbox = sinon.createSandbox();
+		adapter = {} as BaseAdapter;
+		sandbox.stub(adaptersRegistry, 'getAdapter').returns(adapter);
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	it(`should round price to ${DEFAULT_MAX_CPM} if maxCpm is not present on adapter`, () => {
```
Bielik20: I am not sure if it wouldn't be better to replace `${DEFAULT_MAX_CPM}` with `DEFAULT_MAX_CPM`. Should some test fail I tend to just search `it` string in IDE. Also it seems more descriptive.

jacekbj: I decided for a compromise:
`should round price to DEFAULT_MAX_CPM (${DEFAULT_MAX_CPM}) if maxCpm is not present on adapter`

spec/ad-bidders/prebid/price-helper.spec.ts
jacekbj: That's handled by a default value but better safe then sorry :+1:

spec/ad-bidders/prebid/price-helper.spec.ts
jacekbj: Added!

spec/ad-bidders/prebid/price-helper.spec.ts
jacekbj: Done

PR #442, opened 2019-07-09
ADEN-9008 Prepare GP infrastructure

Bielik20:


configs/webpack-app.config.js
```diff
@@ -16,7 +16,12 @@ module.exports.getPreLoaders = () => {
 	];
 };
 
-module.exports.getTypeScriptLoader = (include, tsconfig, transpileOnly = false) => {
+module.exports.getTypeScriptLoader = ({
+	include,
+	tsconfig,
+	transpileOnly = false,
+	reportFiles,
```
fraszczakszymon: Variable without default value shouldn't be placed before defaults?

package.json
```diff
@@ -30,6 +34,7 @@
 		"@slack/client": "4.1.0",
 		"@types/chai": "4.1.7",
 		"@types/doubleclick-gpt": "238.0.1",
+		"@types/jquery": "3.3.29",
```
fraszczakszymon: Do we need it?

package.json
```diff
@@ -131,6 +138,7 @@
 		"preversion": "./maintenance/preversion.sh",
 		"version": "./maintenance/version.sh",
 		"test": "mocha-webpack --full-trace './spec/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.spec.js",
+		"test:gp": "mocha-webpack --full-trace './platforms/gamepedia/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.platforms.js --webpack-env.PLATFORM=gamepedia",
```
fraszczakszymon: In my opinion it creates not needed inconsistency now. I'd keep everything in the same structure for now. And **if we decide** to go with specs along with source code then move all suites at the same time.

tsconfig.json
```diff
@@ -2,21 +2,20 @@
...
 		"resolveJsonModule": true,
 		"noImplicitAny": false,
-		"baseUrl": "src",
+		"baseUrl": ".",
 		"paths": {
-			"@wikia/ad-engine": ["ad-engine"],
-			"@wikia/ad-tracking": ["ad-tracking"]
+			"@ad-engine/core": ["src/ad-engine"],
+			"@wikia/ad-engine": ["src/index.ts"],
+			"@wikia/ad-tracking": ["src/ad-tracking"]
```
fraszczakszymon: What about other packages? Why ad-tracking is separated?

Bielik20: It should be `"@ad-engine/core"`. Other packages are not here because they are not imported anywhere.

configs/webpack-app.config.js
Bielik20: They can, we could even set tslint rule to enforce it if you feel like it. What do you think?

package.json
Bielik20: Removed

package.json
Bielik20: Leaving for now (there are no tests in GP) open for further discussion.

package.json
fraszczakszymon: Yay

package.json
fraszczakszymon: Then why leaving it for now? Still, it leaves inconsistency 🙈

configs/webpack-app.config.js
fraszczakszymon: :+1:

PR #450, opened 2019-07-18
ADEN-9025 Adding HideOnViewability template timeout option

dmnsobczak:


src/ad-products/templates/hide-on-viewability.ts
```diff
@@ -47,24 +49,46 @@ export class HideOnViewability {
...
+		}
+
+		Promise.race([this.registerViewabilityListener(this.adSlot), this.registerTimeoutHide()]).then(
+			() => {
+				this.hideSlot(this.adSlot);
+			},
+		);
+	}
+
+	private registerViewabilityListener(adSlot: AdSlot): Promise<void> {
```
fraszczakszymon: s/registerViewabilityListener/getViewabilityPromise

src/ad-products/templates/hide-on-viewability.ts
```diff
@@ -47,24 +49,46 @@ export class HideOnViewability {
...
+				}, this.config.additionalHideTime || 0);
+			});
+		});
 	}
 
-	private registerViewabilityListener(adSlot: AdSlot): void {
-		adSlot.on(AdSlot.SLOT_VIEWED_EVENT, () => {
-			if (!this.config || !this.config.hideAnimation) {
-				return;
+	private registerTimeoutHide(): Promise<void> {
```
fraszczakszymon:
```suggestion
	private getTimeoutHidePromise(): Promise<void> {
```

src/ad-products/templates/hide-on-viewability.ts
```diff
@@ -47,24 +49,46 @@ export class HideOnViewability {
...
+
+		Promise.race([this.registerViewabilityListener(this.adSlot), this.registerTimeoutHide()]).then(
+			() => {
+				this.hideSlot(this.adSlot);
+			},
+		);
+	}
+
+	private registerViewabilityListener(adSlot: AdSlot): Promise<void> {
+		return new Promise((resolve) => {
```
fraszczakszymon:
```suggestion
		return new Promise((resolve) => {
			adSlot.viewed.then(() => {
				setTimeout(() => {
					utils.logger(HideOnViewability.getName(), 'viewability Promise resolved');
					resolve();
				}, this.config.additionalHideTime || 0);
			});
		});
```

Would it work? What do you think? Worth changing? Your call - I'm fine with leaving it as it is too.

src/ad-products/templates/hide-on-viewability.ts
dmnsobczak: Done!

src/ad-products/templates/hide-on-viewability.ts
dmnsobczak: Done!

src/ad-products/templates/hide-on-viewability.ts
dmnsobczak: Yeah, why not. It's still working and it's more readable. Promises FTW!

src/ad-products/templates/hide-on-viewability.ts
```diff
@@ -47,24 +49,46 @@ export class HideOnViewability {
...
+		});
 	}
 
-	private registerViewabilityListener(adSlot: AdSlot): void {
-		adSlot.on(AdSlot.SLOT_VIEWED_EVENT, () => {
-			if (!this.config || !this.config.hideAnimation) {
-				return;
+	private getTimeoutHidePromise(): Promise<void> {
+		return new Promise((resolve) => {
+			if (this.config.timeoutHideTime) {
```
fraszczakszymon: 1. This promise never resolves when timeout is not set. What about returning null or rejecting?
2. It does not work with falsy values (0 timeout).

dmnsobczak: You right. Changed.

PR #452, opened 2019-07-19
[ADEN-9068] GAM lazy loading enabled

Loczi94:


src/ad-engine/services/slot-repeater.ts
```diff
@@ -60,15 +61,25 @@ function repeatSlot(adSlot: AdSlot): boolean {
 class SlotRepeater {
 	init(): void {
 		if (context.get('options.slotRepeater')) {
-			context.push('listeners.slot', {
-				onRenderEnded: (adSlot: AdSlot) => {
+			if (context.get('options.gamLazyLoading.enabled')) {
+				eventService.on(events.AD_SLOT_CREATED, (adSlot: AdSlot) => {
```
fraszczakszymon: What about extracting registered method to `handleSlotRepeating` and reuse it in both places (eventService.on and onRenderEnded)
```suggestion
				eventService.on(events.AD_SLOT_CREATED, (adSlot: AdSlot) => handleSlotRepeating(adSlot));
```

PR #457, opened 2019-07-22
ADEN-9074 Load Porvata in cdm-zone-02

preludek:


platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,35 @@ class AdsSetup {
...
 	private injectIncontentPlayer(): void {
+		const slotName = 'incontent_player';
 		const isVideo =
 			!!document.getElementById('mf-video') ||
 			!!document.getElementById('twitchnet-liveontwitch') ||
 			!!document.getElementById('ds_cpp');
 
-		if (!isVideo) {
-			slotInjector.inject('incontent_player');
+		if (isVideo) {
```
dmnsobczak: Variable `isVideo` is used only once. Maybe it's better to copy those conditions directly into if?

platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,35 @@ class AdsSetup {
...
+			return;
+		}
+
+		if (!document.getElementById('cdm-zone-02')) {
```
dmnsobczak: Let's parametrize name `cdm-zone-02`, eg. `const porvataSlots = ['cdm-zone-02'];` + foeEach?

platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,35 @@ class AdsSetup {
...
+			this.initiateIncontentPlayer(slotName);
 		}
+
+		slotService.on('cdm-zone-02', AdSlot.STATUS_SUCCESS, () => {
+			if (!!document.querySelector('.porvata3')) {
```
dmnsobczak: Looking for `.porvata3` in DOM is inefficient. Let's set some variable in Porvata template, eg. `context.set('options.video.porvataLoaded', true);` and use it as a condition. It's safer.

platforms/gamepedia/slots.ts
```diff
@@ -67,6 +67,9 @@ class SlotsContext {
 			},
 			'cdm-zone-02': {
 				aboveTheFold: true,
+				disabled: true,
```
dmnsobczak: Remove this line.

platforms/gamepedia/setup.ts
preludek: done

platforms/gamepedia/slots.ts
preludek: done

platforms/gamepedia/setup.ts
preludek: done

platforms/gamepedia/setup.ts
preludek: Can you give more arguments about that change? 
I parametrized `cdm-zone-02` within the function.

platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,36 @@ class AdsSetup {
 	}
 
 	private injectIncontentPlayer(): void {
-		const isVideo =
+		const icpSlotName = 'incontent_player';
```
fraszczakszymon: Use clear and descriptive scoped variable names please. E.g. `incontentPlayerSlotName`

platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,36 @@ class AdsSetup {
...
+		const slotCheck = 'cdm-zone-02';
```
fraszczakszymon: Same here. E.g. `porvataAlternativeSlotName`? `porvataDisplaySlotName`?

platforms/gamepedia/setup.ts
```diff
@@ -174,14 +182,36 @@ class AdsSetup {
...
+		});
+
+		slotService.on(slotCheck, AdSlot.STATUS_COLLAPSE, () => {
+			this.initiateIncontentPlayer(icpSlotName);
+		});
+	}
+
+	private initiateIncontentPlayer(slotName: string): void {
+		slotInjector.inject(slotName);
+		slotsContext.setState(slotName, context.get('options.video.isOutstreamEnabled'));
```
fraszczakszymon: This approach might cause race conditions once we decide to add bidders. Bidders are called based on slotsContext states and this change here may happen way too late. There is no action for now, but we'll need to remember about that during bidders integration.

platforms/gamepedia/setup.ts
preludek: done

platforms/gamepedia/setup.ts
preludek: done

PR #461, opened 2019-07-25
ADEN-9078 InstantConfig interpreter

Bielik20:


spec/ad-services/instant-config/instant-config.interpreter.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+		};
+
+		mockResponses([true, true, true, true]);
+		interpreter.getValues(instantConfig);
+
+		expect(browserIsValidStub.firstCall.args[0]).to.equal(instantConfig.babDetection[0].browsers);
+		expect(deviceIsValidStub.firstCall.args[0]).to.equal(instantConfig.babDetection[0].devices);
+		expect(domainIsValidStub.firstCall.args[0]).to.equal(instantConfig.babDetection[0].domains);
+		expect(regionIsValidStub.firstCall.args[0]).to.equal(instantConfig.babDetection[0]);
+		expect(regionIsValidStub.firstCall.args[1]).to.equal('babDetection-0');
```
fraszczakszymon: We need to check tracking and confirm with Martyna/Dorota whether they are fine with new labels (with additional group id)

spec/ad-services/instant-config/instant-config.loader.spec.ts
```diff
@@ -1,9 +1,11 @@
+import { instantConfigLoader } from '@wikia/ad-services/instant-config/instant-config.loader';
```
fraszczakszymon: When did we started this kind of pattern? Why not `instant-config-loader`?

spec/ad-services/instant-config/matchers/device-matcher.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { utils } from '@wikia/ad-engine';
+import { DeviceMatcher } from '@wikia/ad-services/instant-config/matchers/device-matcher';
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+describe('Device Matcher', () => {
```
fraszczakszymon: :gem: Good scenarios

spec/ad-services/instant-config/matchers/browser-matcher.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { utils } from '@wikia/ad-engine';
+import { BrowserMatcher } from '@wikia/ad-services/instant-config/matchers/browser-matcher';
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+describe('Browser Matcher', () => {
```
fraszczakszymon: :gem: Good scenarios

spec/ad-services/instant-config/matchers/domain-matcher.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { DomainMatcher } from '@wikia/ad-services/instant-config/matchers/domain-matcher';
+import { expect } from 'chai';
+
+describe('Domain Matcher', () => {
```
fraszczakszymon: :gem: Good scenarios

spec/ad-services/instant-config/matchers/region-matcher.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { utils } from '@wikia/ad-engine';
+import { RegionMatcher } from '@wikia/ad-services/instant-config/matchers/region-matcher';
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+
+describe('Region Matcher', () => {
```
fraszczakszymon: :gem: Good scenarios

spec/ad-services/instant-config/matchers/region-matcher.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	it('should filter out invalid', () => {
+		regionMatcher.isValid({
+			regions: ['XX/25-cached', 'US-cached', 'PL/25', 'non-PL', 'UK'],
+		});
+
+		expect(isProperGeoStub.getCall(0).args[0]).to.deep.equal(['non-PL', 'UK']);
+	});
+
+	it('should apply sampling', () => {
```
fraszczakszymon: Unfortunately, I believe we have to move sampling from geo module and apply it for whole feature flag groups.

src/ad-services/instant-config/instant-config.service.ts
```diff
@@ -0,0 +1,41 @@
...
+		return defaultValue;
+	}
+
+	/**
+	 * Use only for legacy wgAdDriver keys
+	 * @deprecated
+	 */
+	isGeoEnabled(key: string): boolean {
+		if (!key.startsWith('wgAdDriver')) {
+			throw new Error('This method should be only used for legacy wgAdDriver keys');
```
fraszczakszymon: 💎

src/ad-services/instant-config/matchers/region-matcher.ts
```diff
@@ -0,0 +1,33 @@
+import { utils } from '@ad-engine/core';
+import { cacheSuffix, InstantConfigGroup, samplingSeparator } from '../instant-config.models';
+
+// TODO: Move sampling and cache on top of the group. Make so that cache works per group.
```
fraszczakszymon: :+1:

src/ad-engine/services/local-cache.ts
```diff
@@ -1,13 +1,13 @@
...
+import { UniversalStorage } from './universal-storage';
 
 interface CacheItem<T = any> {
 	expires?: number;
 	data: T;
 }
 
 class LocalCache {
-	private storage = new LocalStorage();
+	private storage = new UniversalStorage();
```
fraszczakszymon: What does `UniversalStorage` mean? Why do we change it while leaving `LocalCache` as "local"?

Bielik20: Damian started using it for session cookie by adding argument to constructor. Thus it stopped being LocalStorage per se. It became something more generic/universal - hence the new name.

Although I think it should have reversed control. Meaning Session storage should use universal one and expose methods through its API, not the other way around. But I am afraid it is beyond scope of this ticket.

spec/ad-services/instant-config/instant-config.loader.spec.ts
Bielik20: I am not very attached to this. I just wanted something to distinguish certain parts of instant-config. That's why it is always `instant-config` followed by `.` and specific responsibility. But I can change it if you feel like it's better to stick to `-` everywhere.

src/ad-engine/services/local-cache.ts
fraszczakszymon: God bless you (for the last sentence) 😂

PR #463, opened 2019-07-25
ADEN-9073 | Add roadblock template to Gamepedia

bckatarzyna:


platforms/gamepedia/templates/roadblock-config.ts
```diff
@@ -0,0 +1,9 @@
+import { RoadblockTemplateConfig } from '@wikia/ad-engine';
+
+export function getRoadblockConfig(): RoadblockTemplateConfig {
+	return {
+		slotsToEnable: ['cdm-zone-01', 'cdm-zone-02'],
+		slotsToDisable: ['cdm-zone-03', 'cdm-zone-04', 'cdm-zone-06', 'incontent_player'],
```
fraszczakszymon: Not sure about this part. Haven't we decide on grooming to keep everything enabled?

bckatarzyna: Everything from roadblock enabled

PR #464, opened 2019-07-29
ADEN-8997 Scroll speed tracking

bckatarzyna:


src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
+import { context, events, eventService, scrollSpeedCalculator } from '@ad-engine/core';
+
+class ScrollTracker {
+	scrollSpeedTrackingStarted: boolean;
+	timer: NodeJS.Timer;
+
+	constructor() {
```
fraszczakszymon: constructor seems not needed. Can we set those values 3 lines above?

src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	initScrollSpeedTracking(applicationAreaClass: string): void {
+		if (!context.get('options.scrollSpeedTracking')) {
+			return;
+		}
+
+		const applicationArea = document.getElementsByClassName(applicationAreaClass)[0];
+
+		this.scrollSpeedTrackingStarted = false;
+		applicationArea.addEventListener('touchstart', () => this.trackScrollSpeedToDW());
```
fraszczakszymon: It may throw an error (`applicationArea` can be undefined)

fraszczakszymon: What about SPA? Would below code solve the problem?
```suggestion
		applicationArea.addEventListener('touchstart', () => this.trackScrollSpeedToDW(), { once: true });
```

src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	/**
+	 * Track scrollY to DW in in three 2s-periods
+	 */
+	trackScrollSpeedToDW(): void {
```
fraszczakszymon: It doesnt really track to DW. What about `dispatchScrollSpeedEvents`?

src/ad-tracking/scroll-tracker.ts
fraszczakszymon: I just found reset method :P. Anyway "once" would be nice here too

src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
...
+			return;
+		}
+
+		const applicationArea = document.getElementsByClassName(applicationAreaClass)[0];
+
+		this.scrollSpeedTrackingStarted = false;
+		applicationArea.addEventListener('touchstart', () => this.trackScrollSpeedToDW());
+	}
+
+	/**
```
fraszczakszymon: Aren't we removing jsdocs? :P (in favour of TS)

src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		const timesToTrack = [0, 2, 4];
+		let startScrollY = 0;
+		this.scrollSpeedTrackingStarted = true;
+
+		timesToTrack.forEach((time) => {
+			this.timer = setTimeout(() => {
+				const scrollY = window.scrollY || window.pageYOffset;
+				eventService.emit(events.TRACK_SCROLL_Y, time, scrollY);
```
fraszczakszymon: TRACK_SCROLL_Y -> SCROLL_POSITION_CHANGED?

src/ad-tracking/scroll-tracker.ts
```diff
@@ -0,0 +1,63 @@
...
+		});
+	}
+
+	/**
+	 * Remove scroll tracking from the page
+	 */
+	resetScrollSpeedTracking(applicationAreaClass: string): void {
+		const applicationArea = document.getElementsByClassName(applicationAreaClass)[0];
+		clearTimeout(this.timer);
+		applicationArea.removeEventListener('touchstart', () => this.trackScrollSpeedToDW());
```
fraszczakszymon: Does it work? It gets a newly created function in the second argument. Should it be the one added in `addEventListener`?

fraszczakszymon: What about assigning applicationArea to variable on init, so we don't need to pass class name again to reset method?

src/ad-tracking/scroll-tracker.ts
bckatarzyna: I've copied them from mobile-wiki where there were previously written ;)

PR #465, opened 2019-07-29
ADEN-8666 | Remove slotId from a9 context (#397)

bckatarzyna:
* ADEN-8666 | Remove slotId from a9 context

* ADEN-8666 | Unifying slot mapping

* ADEN-8666 | A9 tests fix


src/ad-bidders/a9/index.ts
```diff
@@ -258,14 +256,11 @@ export class A9 extends BaseBidder {
 	 */
 	createSlotDefinition(slotAlias: string): A9SlotDefinition | null {
```
dmnsobczak: `slotAlias: string` -> `slotName: string`
and then:
```
slotID: slotName,
slotName,
```
?

PR #466, opened 2019-07-30
ADEN-9100 | Add is_uap param to adengadinfo

bckatarzyna:


src/ad-tracking/slot-tracking-middleware.ts
```diff
@@ -30,6 +30,7 @@ export const slotTrackingMiddleware: utils.Middleware<AdInfoContext> = ({ data,
 			country: (utils.geoService.getCountryCode() || '').toUpperCase(),
 			device: utils.client.getDeviceType(),
 			document_visibility: utils.getDocumentVisibilityStatus(),
+			is_uap: context.get(`slots.${slot.getSlotName()}.targeting.uap`) ? 1 : 0,
```
fraszczakszymon:
```suggestion
			is_uap: slot.getConfigProperty('targeting.uap') ? 1 : 0,
```

fraszczakszymon: It will catch roadblocks too, is it ok?

bckatarzyna: As discussed with @dorotamierzwa it's ok for us

PR #481, opened 2019-08-09
ADEN-9134 Adding BT rec service

dmnsobczak:


src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,121 @@
...
+	uid: string;
+	style?: Dictionary<string>;
+	size: Dictionary<number>;
+	lazy?: boolean;
+}
+
+/**
+ * Injects BT script
+ */
+function loadScript(): Promise<Event> {
```
Bielik20: why this isn't a private method?

@types/window/index.d.ts
```diff
@@ -5,6 +5,9 @@ interface Window {
 	__cmp?: WindowCMP;
 	AdEngine_adType?: any;
 	ads?: Ads;
+	adsQueue?: any;
+	apstag?: Apstag;
+	BT: any;
```
Bielik20: `any`? No can do :P 
But seriously please create types for that 🙏 

Or even better move it to BTrec class as it is used only there. Like that:

```ts
interface Window {
  BT: myAwesomeType
}
```

src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * BT service handler
+ */
+class BTRec {
+	private placementsMap: Dictionary<BTPlacementConfig>;
+
+	/**
+	 * Runs BT rec service and injects code
+	 */
+	run(): Promise<void> {
```
Bielik20: It appears to me that only this method should be public.

src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+			return Promise.resolve();
+		}
+
+		utils.logger(logGroup, 'loading');
+
+		this.markAdSlots();
+
+		if (!isDebug) {
+			loadScript().then(this.triggerScript);
```
Bielik20: please use async await when possible. Making it async would allow people to listen for completion of that method.


(+) please when using `then` prefer arrow function. This `loadScript().then(this.triggerScript);` leaves ambiguity what is this method execution context.

src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,121 @@
...
+	return utils.scriptLoader.loadScript(btLibraryUrl, 'text/javascript', false, document.head
+		.lastChild as HTMLElement);
+}
+
+/**
+ * BT service handler
+ */
+class BTRec {
+	private placementsMap: Dictionary<BTPlacementConfig>;
+
```
Bielik20: Please change order of the methods so that it reflects usage. From what I see, it seems to be:

1. run
2. markAdSlot
3. duplicateSlot
4. loadScript
5. triggerScript

getPlacementId - is declared but never used. Plus perhaps it should be async so that it awaits execution of run. Since it is not used it is hard for me to find a use case for that.

@types/window/index.d.ts
dmnsobczak: One day I'm gonna sit with my children or grandchildren and tell them a story about this Code Review Comment... changed :D

src/ad-services/bt-rec/index.ts
dmnsobczak: I don't have clear reasons. Moved.

src/ad-services/bt-rec/index.ts
dmnsobczak: Moved and removed.

src/ad-services/bt-rec/index.ts
dmnsobczak: It's true for now, bu once we reimplement lazy Blockthrough loading we'll need some of them to be public. But for now we can hide them.

src/ad-services/bt-rec/index.ts
dmnsobczak: Ok, changed to async.

src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,123 @@
...
+interface BTLoader {
+	clearThrough: () => void;
+}
+
+/**
+ * BT service handler
+ */
+class BTRec {
+	private bt: BTLoader = null;
```
Bielik20: unnecessary null assignment.

```ts
private bt: BTLoader;
```

src/ad-services/bt-rec/index.ts
```diff
@@ -0,0 +1,123 @@
+import { context, Dictionary, utils } from '@ad-engine/core';
+
+const logGroup = 'bt-loader';
+const isDebug = utils.queryString.isUrlParamSet('bt-rec-debug');
+const placementClass = 'bt-uid-tg';
+const scriptDomain = 'wikia-inc-com.videoplayerhub.com';
```
Bielik20: I am not sure, but perhaps those should be class members as well. What do you think?

src/ad-services/bt-rec/index.ts
dmnsobczak: 👍

src/ad-services/bt-rec/index.ts
dmnsobczak: Yes and no. I shuffled them a bit, what do you think now?

PR #491, opened 2019-08-22
ADEN-9205 | Remove group number from geo variable name

bckatarzyna:


spec/ad-engine/services/geo-cache-storage.spec.ts
```diff
@@ -111,12 +111,12 @@ describe('Geo Cache Storage', () => {
 		it('should return array', () => {
 			getItemStub.returns({
 				first: {
-					name: 'first',
+					name: 'first-1',
```
Bielik20: I would add one additional scenario with name without a suffix, but other than that ✅

src/ad-engine/services/geo-cache-storage.ts
```diff
@@ -70,10 +70,17 @@ class GeoCacheStorage {
 		return Object.keys(this.cacheStorage).map((id) => this.getResultLog(id));
 	}
 
+	transformName(name: string): string {
```
Bielik20: How about "removeIndexSuffix"?
(+) please move bellow `getRestultLog` and make `private`.

spec/ad-engine/services/geo-cache-storage.spec.ts
bckatarzyna: Added

PR #492, opened 2019-08-23
ADEN-9211 Adjust Muthead ad unit and src

Bielik20:


src/ad-engine/models/targeting.ts
```diff
@@ -0,0 +1,27 @@
+export interface Targeting {
```
Bielik20: Please someone take a look at it. Maybe we don't need some of that stuff anymore.

platforms/mutfut/targeting.ts
```diff
@@ -1,59 +1,33 @@
...
+	const cid = utils.queryString.get('cid');
+	const targeting: Partial<Targeting> = {
+		ae3: '1',
+		skin: 'turf',
+		uap: 'none',
+		uap_c: 'none',
+		s0: 'gaming',
+		s1: domain.name,
+		s2: 'main',
+		dmn: `${domain.name}${domain.tld}`,
```
nandy-andy: What `tld` stands for?

src/ad-engine/models/targeting.ts
```diff
@@ -0,0 +1,27 @@
...
+	ae3: '1';
```
nandy-andy: We can get rid of `ae3` key-val I think. It was needed in the transition phase when for Oasis and mobile-wiki we had also ae2 :)

src/ad-engine/models/targeting.ts
```diff
@@ -0,0 +1,27 @@
...
+	hostpre: string;
+	lang: string;
+	wpage: string;
+	ref: string;
+	esrb: string;
+	geo: string;
+	likho: string[];
+	pv: string;
+	cid: string;
+	top: '1k';
```
nandy-andy: The rest look fine. I'm not sure about the `s2`, `s0v` and `s0c` but if it's consistent with what we did in Gamepedia it's fine :)

platforms/mutfut/targeting.ts
nandy-andy: 🙈 Got it! https://en.wikipedia.org/wiki/Top-level_domain 😊

PR #494, opened 2019-08-23
ADEN-9244 Fix scroll tracker not closing correctly.

bckatarzyna:


spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { utils } from '@wikia/ad-engine';
+import { expect } from 'chai';
+import { createSandbox, SinonSandbox } from 'sinon';
+
+describe('buildPromisedTimeout', () => {
+	let sandbox: SinonSandbox;
+	let clock;
```
Bielik20: type :trollface:

src/ad-tracking/scroll-tracker.ts
```diff
@@ -1,54 +1,87 @@
...
+		return !!context.get('options.scrollSpeedTracking') && !!this.applicationArea;
+	}
+
+	private addTouchStartListener(): void {
+		this.listener = () => this.dispatchScrollSpeedEvents();
+		this.applicationArea.addEventListener('touchstart', this.listener, { once: true });
+	}
 
-		clearTimeout(this.timer);
+	private removeTouchStartListener(): void {
```
Bielik20: could be placed bellow `finishScrollSpeedTracking`.

src/ad-tracking/scroll-tracker.ts
```diff
@@ -1,54 +1,87 @@
...
 		this.applicationArea.removeEventListener('touchstart', this.listener);
-		if (this.speeds.length) {
-			scrollSpeedCalculator.setAverageSessionScrollSpeed(this.speeds);
-		}
 	}
-}
 
-export const scrollTracker = new ScrollTracker();
+	private dispatchScrollSpeedEvents(): void {
+		this.timers = this.timesToTrack.map(utils.buildPromisedTimeout).map((timer) => ({
```
Bielik20: Please use arrow function.
Alternatively you could change `buildPromisedTimeout` into `const` but I would prefer 1st option.

src/ad-tracking/scroll-tracker.ts
```diff
@@ -1,54 +1,87 @@
...
+	private removeTouchStartListener(): void {
 		this.applicationArea.removeEventListener('touchstart', this.listener);
-		if (this.speeds.length) {
-			scrollSpeedCalculator.setAverageSessionScrollSpeed(this.speeds);
-		}
 	}
-}
 
-export const scrollTracker = new ScrollTracker();
+	private dispatchScrollSpeedEvents(): void {
```
Bielik20: I like this method all in all. I have got some suggestions I don't feel so strongly about if you will you can check them out:

```ts
private async dispatchScrollSpeedEvents(): Promise<void> {
	this.timers = this.timesToTrack
		.map((timeToTrack) => utils.buildPromisedTimeout(timeToTrack))
		.map(({ cancel, promise }) => ({
			cancel,
			promise: promise.then((time) => {
				const measurement: SpeedMeasurement = {
					time,
					distance: this.distance,
				};

				eventService.emit(events.SCROLL_TRACKING_TIME_CHANGED, time, this.scrollY);
				this.prevScrollY = this.scrollY;

				return measurement;
			}),
		}));

	const measurements: SpeedMeasurement[] = await Promise.all(
		this.timers.map((timer) => timer.promise),
	);

	this.finishScrollSpeedTracking();
	scrollSpeedCalculator.setAverageSessionScrollSpeed(
		measurements
			.filter((measurement) => measurement.time > 0)
			.map((measurement) => measurement.distance),
	);
}
```

Then you don't even have to keep `measurements` as instance variable.

src/ad-engine/utils/promised-timeout.ts
```diff
@@ -0,0 +1,18 @@
+export interface PromisedTimeout<T> {
+	cancel: () => void;
+	promise: Promise<T>;
+}
+
+export function buildPromisedTimeout(time): PromisedTimeout<number> {
```
Bielik20: type :trollface: 

(+) would you consider moving it to folder "utils-loose" or something like that. We want to remove this utils wrapper so I would avoid adding more to it. Create folder that would be for transition period only. What do you think?

spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	it('should resolve after provided time passes', async () => {
+		let resolved = false;
+		const timeout: utils.PromisedTimeout<number> = utils.buildPromisedTimeout(1000);
+
+		timeout.promise.then(() => {
+			resolved = true;
```
Bielik20: You could have used spy to check if it has been called instead of flag, but it just nitpicking.

spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+		});
+
+		expect(resolved).to.equal(false);
+
+		clock.tick(800);
+		await Promise.resolve();
```
Bielik20: that is interesting, could you explain?

spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+		expect(resolved).to.equal(true);
+	});
+
+	it('should resolve to timeout time', async () => {
+		let result;
+		const time = 1000;
+		const timeout: utils.PromisedTimeout<number> = utils.buildPromisedTimeout(time);
+
+		timeout.promise.then((time) => {
+			result = time;
```
Bielik20: ditto

spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+			resolved = true;
+		});
+
+		expect(resolved).to.equal(false);
+
+		clock.tick(800);
+		await Promise.resolve();
+		expect(resolved).to.equal(false);
+
+		clock.tick(201);
```
Bielik20: why not 200?

spec/ad-engine/utils/promised-timeout.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+		await Promise.resolve();
+		expect(result).to.equal(time);
+	});
+
+	it('should be cancellable', async () => {
+		let result = 0;
+		const timeout: utils.PromisedTimeout<number> = utils.buildPromisedTimeout(1000);
+
+		timeout.promise.then((time) => {
+			result = time;
```
Bielik20: ditto

spec/ad-tracking/scroll-tracker.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * The use of Promise.resolve() is required for testing Promise-based code.
+ * See https://stackoverflow.com/questions/55440400/
+ * testing-that-promise-resolved-not-until-timeout-sinon-chai#
+ * for explanation.
+ */
+
+describe('ScrollTracker', () => {
+	const sandbox = createSandbox();
+	let getElementsStub: SinonStub;
```
Bielik20: you don't need to store this one

spec/ad-tracking/scroll-tracker.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+					fake: [buildFakeElement()],
+				}[className] || []
+			);
+		});
+		clock = sandbox.useFakeTimers();
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+		eventService.removeAllListeners();
```
Bielik20: I feel like this should be stubed - but maybe not?

src/ad-tracking/scroll-tracker.ts
```diff
@@ -1,54 +1,87 @@
...
+				this.prevScrollY = this.scrollY;
+
+				return measurement;
+			}),
+		}));
+
+		Promise.all(this.timers.map((timer) => timer.promise))
+			.then(() => this.finishScrollSpeedTracking())
+			.then(() =>
+				scrollSpeedCalculator.setAverageSessionScrollSpeed(
```
Bielik20: I feels like this won't be called at all if at least one timer is cancelled. Is it by design?

spec/ad-tracking/scroll-tracker.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+
+		eventService.emit(events.BEFORE_PAGE_CHANGE_EVENT);
+
+		clock.tick(700);
+		await Promise.resolve();
+
+		expect(emittedEvents.length).to.equal(1);
+	});
+
+	it('should call scrollSpeedCalculator.getAverageSessionScrollSpeed once timeouts complete', async () => {
```
Bielik20: would like to see test for `scrollSpeedCalculator.getAverageSessionScrollSpeed` with cancelled timer.

src/ad-tracking/scroll-tracker.ts
jacekbj: Yeah I thought of that. Here I decided to keep them together as they're a logic pair.

src/ad-tracking/scroll-tracker.ts
jacekbj: Why use arrow function here - for consistency only?
`utils.buildPromisedTimeout` is safe to use as it's used here.

spec/ad-tracking/scroll-tracker.spec.ts
jacekbj: :+1:

src/ad-tracking/scroll-tracker.ts
jacekbj: Yes.

spec/ad-tracking/scroll-tracker.spec.ts
jacekbj: You got it!

spec/ad-tracking/scroll-tracker.spec.ts
jacekbj: It's fairly easy to test this one without stubbing. Whenever there is a chance I like to use that opportunity - it's easier to spot some unintended dependencies.

spec/ad-engine/utils/promised-timeout.spec.ts
jacekbj: Having come from a different testing framework (Jasmine) I find it better to use this pattern. The reason is that it's explicit and can be used with any testing framework.
Here we could either put `expect` inside callback or program the "expectance" of spy, which I always need to read about before using 😅

src/ad-engine/utils/promised-timeout.ts
jacekbj: What would be the benefit of it? It kind of feels like a separate package (`ad-engine/tools`) rather then a new subdirectory in `ad-engine/core`.

I don't have a clear picture here. If you could share one then we can iterate and/or move this piece.

src/ad-tracking/scroll-tracker.ts
jacekbj: That is brilliant, thank you!
Having tests I'm convenient it works :D Time well spent.

spec/ad-engine/utils/promised-timeout.spec.ts
jacekbj: Added comment in test.
You can read more on it here: https://stackoverflow.com/questions/55440400/testing-that-promise-resolved-not-until-timeout-sinon-chai#

More info avaiable here:
https://stackoverflow.com/questions/52177631/jest-timer-and-promise-dont-work-well-settimeout-and-async-function/52196951#52196951

src/ad-tracking/scroll-tracker.ts
Bielik20: Yes, but it may not in the future. Someone will surely forget about it.

spec/ad-engine/utils/promised-timeout.spec.ts
Bielik20: Ok I think I get it. Promises won't resolve if there is synchronous code to be executed. But what about those two promises next to each other? one releases the other so that it must wait?

jacekbj: Without 2x `Promise.resolve` the `Promise.all` was not resolved. Strange as hell but yes, this solved it.

src/ad-tracking/scroll-tracker.ts
jacekbj: OK

spec/ad-engine/utils/promised-timeout.spec.ts
jacekbj: No reason. Changed it to 200.

src/ad-tracking/scroll-tracker.ts
Bielik20: I had that in mind as well. I would have probably moved it but it's fine.

src/ad-engine/utils/promised-timeout.ts
Bielik20: 1. I would love to reorginize ad-engine/core as I feel like its just a bag for everything. Still I would do it in separate ticket (I even think it exists somewhere).

2. Now I would create just other folder to use in transition period. So that there is no `import as utils` -> `export utils`. The reason for that is I don't think we would be able to move all utils in one move.

jacekbj: I'd rather wrap it in a ticket. There is a question of where this piece should be placed: in ad-engine or in a separate package.

spec/ad-engine/services/scroll-speed-calculator.stub.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type ScrollSpeedCalculatorStub = { [key in keyof ScrollSpeedCalculator]: SinonStub };
+
+export function createScrollSpeedCalculatorStub(sandbox: SinonSandbox): ScrollSpeedCalculatorStub {
+	return {
+		setAverageSessionScrollSpeed: sandbox.stub(),
+	} as any;
+}
+
+export function stubScrollSpeedCalculator(
```
Bielik20: very pretty code :trollface:

PR #495, opened 2019-08-26
ADEN-9243 Fix Cookie

Bielik20:


src/ad-engine/services/instant-config-cache-storage.ts
```diff
@@ -13,17 +12,26 @@ export interface CacheData {
...
+	private readonly sessionCookie = SessionCookie.make();
 	private cacheStorage: CacheDictionary;
 
-	constructor() {
+	private constructor() {
 		this.resetCache();
 	}
 
 	resetCache(): void {
-		sessionCookie.readSessionId();
```
Bielik20: @dmnsobczak please confirm you are sure it is not required here - on all platforms.

dmnsobczak: https://github.com/Wikia/mobile-wiki/pull/1765/files#diff-34a2814942202b24a3dd9801e609c9f3R318
This will break this edge case:
- Open mobile-wiki page
- Hold your phone on this page for 30+ minutes
- Go to next page
Session id won't change in that case. Decide if it's a serious issue or we can ignore it or we can fix it different way (eg. by calling `readSessionId()` from mobile-wiki `triggerBeforePageChangeServices()`).

src/ad-engine/services/scroll-speed-calculator.ts
```diff
@@ -1,22 +1,33 @@
...
+		if (!ScrollSpeedCalculator.instance) {
+			ScrollSpeedCalculator.instance = new ScrollSpeedCalculator();
+		}
+
+		return ScrollSpeedCalculator.instance;
+	}
+
+	private sessionCookie = SessionCookie.make();
+
+	private constructor() {}
```
dmnsobczak: Remove it?

src/ad-engine/services/universal-storage.ts
```diff
@@ -1,10 +1,10 @@
 import { Dictionary } from '../models';
 
 export interface StorageProvider {
-	getItem: (key: string) => string;
```
dmnsobczak: I'm surprised that with all that work you didn't used general Storage type :D

PR #499, opened 2019-08-28
ADEN-8843 Adding UAP template

dmnsobczak:


platforms/shared/slots.ts
```diff
@@ -17,6 +18,20 @@ class SlotsContext {
...
+		const queryParam = utils.queryString.get('uap-pv-restriction');
+
+		if (typeof queryParam !== 'undefined') {
+			uapRestriction = parseInt(queryParam, 10);
+		}
+
+		const isUapAllowed =
+			uapRestriction === window.pvNumber || uapRestriction === 0 || context.get('src') === 'test';
+
+		return isUapAllowed && instantConfig.isGeoEnabled('wgAdDriverUapCountries');
```
jacekbj: Isn't `instantConfig.isGeoEnabled` obsolete?

platforms/shared/slots.ts
```diff
@@ -17,6 +18,20 @@ class SlotsContext {
 		});
 	}
 
+	isUapAllowed(instantConfig: InstantConfigService): boolean {
+		let uapRestriction = instantConfig.get('wgAdDriverUapRestriction');
```
jacekbj: The default value should be `0` so that it works on Muthead (where wgAdDriverUapRestriction is not defined), right?

platforms/shared/slots.ts
```diff
@@ -17,6 +18,20 @@ class SlotsContext {
 		});
 	}
 
+	isUapAllowed(instantConfig: InstantConfigService): boolean {
```
jacekbj: It isn't really related to any slot in particular. Shouldn't that be placed somewhere in `platforms/shared/templates`?

platforms/sports/setup-context.ts
```diff
@@ -103,6 +105,11 @@ class ContextSetup {
 		context.set('targeting', getPageLevelTargeting());
 		context.set('options.maxDelayTimeout', this.instantConfig.get('wgAdDriverDelayTimeout', 2000));
 
+		if (slotsContext.isUapAllowed(this.instantConfig)) {
+			const uapSize: [number, number] = isMobile ? [2, 2] : [3, 3];
```
jacekbj: Aren't these sizes common for GP and Sports? If so that's a good candidate for moving to `platforms/shared`

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const {
+	CSS_CLASSNAME_STICKY_BFAA,
+	CSS_CLASSNAME_THEME_RESOLVED,
+	CSS_TIMING_EASE_IN_CUBIC,
+	FADE_IN_TIME,
+	SLIDE_OUT_TIME,
+} = universalAdPackage;
+
+export function getBfaaConfig(): any {
```
jacekbj: We can use `BigFancyAdAboveConfig` from ad-engine.

platforms/sports/templates/big-fancy-ad-below-config.ts
```diff
@@ -0,0 +1,16 @@
+import { slotsContext } from '@platforms/shared';
+import { AdSlot, UapParams } from '@wikia/ad-engine';
+
+export function getBfabConfig(): any {
```
jacekbj: That's a good moment to add a type for that 💪 
Strong faith I have in you, the defender of the cause!

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+		enabled: true,
+
+		onInit(adSlot: AdSlot, params: UapParams, config: BigFancyAdAboveConfig): void {
+			this.adSlot = adSlot;
+			this.templateParams = params;
+			this.config = config || context.get('templates.bfaa') || {};
+			this.navbarElement = document.querySelector('.header');
+			this.navbarManager = new NavbarManager(this.navbarElement);
+
+			context.set('slots.cdm-zone-04.defaultSizes', [[3, 3]]);
```
jacekbj: I wonder why do this. BFAA is applied once line item has already been selected. Do you happen to know the reasoning?

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+			Object.assign(this.navbarElement.style, {
+				transition:
+					`top ${SLIDE_OUT_TIME}ms ${CSS_TIMING_EASE_IN_CUBIC}, ` +
+					`opacity ${FADE_IN_TIME}ms ${CSS_TIMING_EASE_IN_CUBIC}`,
+				top: '0',
+			});
+		},
+
+		onAfterUnstickBfaaCallback(): void {
+			Object.assign(this.navbarElement.style, {
```
jacekbj: Is it possible to deal with this one using CSS class or is the property set inline in ad-engine?

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+				top: '',
+			});
+
+			this.updateNavbar();
+			this.navbarScrollListener = scrollListener.addCallback(() => this.updateNavbar());
+		},
+
+		onResolvedStateSetCallback(): void {
+			this.updateNavbar();
+			requestAnimationFrame(() => this.updateNavbar());
```
jacekbj: My eyes bleed into my brain (that's why nobody can see it, even me).
Is it still required?

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+		navbarManager: null,
+		navbarScrollListener: null,
+		enabled: true,
+
+		onInit(adSlot: AdSlot, params: UapParams, config: BigFancyAdAboveConfig): void {
+			this.adSlot = adSlot;
+			this.templateParams = params;
+			this.config = config || context.get('templates.bfaa') || {};
+			this.navbarElement = document.querySelector('.header');
+			this.navbarManager = new NavbarManager(this.navbarElement);
```
jacekbj: We have `this.navbarElement` and `this.navbarManager`. If we move `moveNavbar` method to manager then we can live with one of these only.

platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,99 @@
...
+		defaultStateAllowed: true,
+		fullscreenAllowed: true,
+		stickinessAllowed: true,
+		templateParams: null,
+		slotsToDisable: ['cdm-zone-06', 'incontent_player'],
+		slotsToEnable: ['cdm-zone-02', 'cdm-zone-03', 'cdm-zone-04'],
+		navbarElement: null,
+		navbarManager: null,
+		navbarScrollListener: null,
+		enabled: true,
```
jacekbj: Where is it used?

platforms/shared/slots.ts
dmnsobczak: You probably right. Using this variable would disable UAP on Gamepedia too. Removed.

platforms/shared/slots.ts
dmnsobczak: I don't understand. The default value (defined in IC fallback) is 1 and the variable is defined in ICBM

dmnsobczak: Unless you wanted to say, that `window.pvNumber` is undefined? Then it's true, but I would keep it that way. UAP can be tested with URL parameters and everything will be fine once we reimplement tracking on FutMut.

platforms/shared/slots.ts
dmnsobczak: Nice suggestion. Moved.

platforms/sports/setup-context.ts
dmnsobczak: Moved!

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Used.

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Nowhere, I copied too much. Removed.

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Well, not only this but it's a nice idea. Refactored.

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Slot `cdm-zone-04` loads a custom template via 3x3 and this size is not added to the initial context. That's why we're adding it right here.

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Yes, done.

platforms/sports/templates/big-fancy-ad-above-config.ts
dmnsobczak: Actually I can't reproduce the issue, so probably not. Removed.

platforms/sports/templates/big-fancy-ad-below-config.ts
dmnsobczak: I did it 💪

dmnsobczak:
Code review changes applied :)

src/ad-products/templates/uap/big-fancy-ad-above.ts
```diff
@@ -13,26 +13,30 @@ export type StickinessCallback = (
...
+	handleNavbar?: boolean;
+	autoPlayAllowed?: boolean;
+	defaultStateAllowed?: boolean;
+	fullscreenAllowed?: boolean;
+	stickinessAllowed?: boolean;
+	stickyUntilSlotViewed?: boolean;
+	slotSibling?: string;
+	slotsToEnable?: string[];
 	slotsToDisable?: string[];
 	onInit: (adSlot: AdSlot, params: UapParams, config: BigFancyAdAboveConfig) => void;
```
jacekbj: It's so funny that it turns out that only `onInit` is required :D

PR #500, opened 2019-08-28
ADEN-9239 Use icBabDetection instead of wgAdDriverBabDetection

fraszczakszymon:


platforms/gamepedia/setup-context.ts
```diff
@@ -22,7 +22,6 @@ const fallbackInstantConfig = {
 	wgAdDriverA9BidderCountries: ['XX'],
 	wgAdDriverA9DealsCountries: ['XX'],
 	wgAdDriverAppNexusBidderCountries: ['XX'],
-	wgAdDriverBabDetection: ['XX'],
```
jacekbj: What do you think about specifying fallback value? In this case it would be `false`.

PR #501, opened 2019-08-30
ADEN-8646 Get rid of context's listeners.*

fraszczakszymon:
Removes also:
- Twitch player and all related code
- slotListener
- different ways of emitting ad-slot related events (adSlot vs eventService), from now all events going to be dispatched directly on adSlot


platforms/gamepedia/setup-ad-engine.ts
```diff
@@ -62,10 +63,8 @@ function startAdEngine(): void {
 		}
 	});
 
-	context.push('listeners.slot', {
-		onRenderEnded: (slot) => {
-			slot.getElement().classList.remove('default-height');
-		},
+	eventService.on(AdSlot.SLOT_RENDERED_EVENT, (slot) => {
+		slot.getElement().classList.remove('default-height');
```
jacekbj: Please use `slot.removeClass`.

platforms/sports/setup-ad-engine.ts
```diff
@@ -34,10 +34,8 @@ function startAdEngine(): void {
 	engine.init();
 	babDetection.run();
 
-	context.push('listeners.slot', {
-		onRenderEnded: (slot) => {
-			slot.getElement().classList.remove('default-height');
-		},
+	eventService.on(AdSlot.SLOT_RENDERED_EVENT, (slot) => {
+		slot.getElement().classList.remove('default-height');
```
jacekbj: ditto

fraszczakszymon: :+1:

platforms/gamepedia/setup-ad-engine.ts
fraszczakszymon: :+1:

PR #511, opened 2019-09-09
ADEN-9290 | Add iframe type to click position tracker

bckatarzyna:


src/ad-tracking/click-position-tracker.ts
```diff
@@ -52,6 +53,7 @@ class ClickPositionTracker {
 			return;
 		}
 		const iframeBody: HTMLElement = iframeElement.contentWindow.document.body;
+		const frameType = iframeElement.dataset.isSafeframe === 'true' ? 'safe' : 'regular';
```
fraszczakszymon: What do you think about saving this information in the ad-slot module (somewhere in the `gpt-provider`/`ad-slot`/`slot-service`?) and then reading it from `slot` instead of reading it from DOM attributes here?

PR #512, opened 2019-09-09
ADEN-9289 Leave uap restriction on gp only

fraszczakszymon:


platforms/shared/templates/uap-helper.ts
```diff
@@ -1,19 +1,15 @@
-import { context, InstantConfigService, utils } from '@wikia/ad-engine';
+import { context, utils } from '@wikia/ad-engine';
 import { slotsContext } from '../slots';
 
 class UapHelper {
 	firstCallSlotName = 'cdm-zone-01';
```
nandy-andy: We may think of making this configurable in future. But since now we share the same slot names for gamepedia and sports, it's fine, I guess.

PR #525, opened 2019-09-17
ADEN-9327 Add ad-products types

Bielik20:


src/ad-products/templates/uap/themes/theme.ts
```diff
@@ -1,38 +1,18 @@
-import { AdSlot, context, PorvataPlayer, utils } from '@ad-engine/core';
-import { BigFancyAdAboveConfig } from '../big-fancy-ad-above';
+import { AdSlot } from '@ad-engine/core';
+import { PorvataPlayer } from '../../..';
```
nandy-andy: Can we make it more explicit where the class is defined? Or it's generating some unnecessary cost?

src/ad-products/video/player/porvata/index.ts
```diff
@@ -0,0 +1,5 @@
+// export * from './porvata';
```
nandy-andy: Why can't we just remove this file if it's all commented?

Bielik20: I forgot about it, good catch, fixes 👍

src/ad-products/templates/uap/themes/theme.ts
Bielik20: It should have been full path, must have slipped pass me. I have now different webstorm config - it should suggest correct paths by default, but still need some testing. Good catch, fixed 👍

PR #528, opened 2019-09-19
ADEN-9149 Change compilation to enable type checks

Bielik20:


dmnsobczak:
Code looks fine, but let's not forget about things to "fix later". And thank you for quick typescript magic lesson :D

PR #536, opened 2019-09-26
ADEN-9362 Adjust UAP CSS for Futhead

dmnsobczak:


platforms/sports/templates/big-fancy-ad-above-config.ts
```diff
@@ -34,7 +34,11 @@ export function getBfaaConfig(): BigFancyAdAboveConfig {
 		onInit(adSlot: AdSlot, params: UapParams, config: BigFancyAdAboveConfig): void {
 			this.adSlot = adSlot;
 			this.config = config || context.get('templates.bfaa') || {};
-			this.navbarManager = new NavbarManager(document.querySelector('.header'));
+			this.navbarManager = new NavbarManager(
+				document.querySelector(
+					context.get('targeting.s1') === 'muthead' ? '.header' : '.navbar.navbar-futhead',
```
nandy-andy: Yet, another reason to split `sports` into `futhead` and `muthead` ;) Although, once Futhead is on Turf platform it should be similar 🤔

PR #537, opened 2019-09-26
ADEN-9330 | Non lazy loaded incontent boxads

bckatarzyna:


src/ad-bidders/prebid/index.ts
```diff
@@ -229,6 +231,18 @@ export class PrebidProvider extends BidderProvider {
 		const pbjs: Pbjs = await pbjsFactory.init();
 
 		const refreshUsedBid = (winningBid) => {
+			if (winningBid.adUnitCode === 'mobile_in_content') {
```
dmnsobczak: I don't like the idea of using slot names inside pretty general Prebid class. Can we make them more generic and pass what's possible through context?

src/ad-bidders/prebid/index.ts
```diff
@@ -229,6 +231,18 @@ export class PrebidProvider extends BidderProvider {
...
+				this.incontentsCounter += 1;
+			}
+			context.set(
+				`bidders.prebid.bidsRefreshing.incontent_boxad_${this.incontentsCounter}.finished`,
+				new Promise((resolve) => {
```
dmnsobczak: Really nice code. I would do the same :)

src/ad-bidders/prebid/index.ts
```diff
@@ -239,6 +253,10 @@ export class PrebidProvider extends BidderProvider {
 						adUnit.bids[0].bidder === winningBid.bidderCode,
 				);
 
+				utils.logger(
+					'bids-refreshing',
```
dmnsobczak: Usually we're using `logGroup` in logs to easy find logs of similar kind.

src/ad-engine/services/slot-repeater.ts
```diff
@@ -48,7 +48,11 @@ function repeatSlot(adSlot: AdSlot): boolean {
 
 	const container = slotInjector.inject(slotName);
 
-	if (!!container && context.get('options.gamLazyLoading.enabled')) {
+	if (
+		!!container &&
+		(context.get('options.gamLazyLoading.enabled') ||
```
dmnsobczak: Prettier, whyyy?

src/ad-engine/services/slot-repeater.ts
```diff
@@ -72,11 +76,43 @@ function handleSlotRepeating(adSlot: AdSlot): boolean {
 	return false;
 }
 
+async function injectNextSlot(adSlot: AdSlot): Promise<boolean> {
```
dmnsobczak: Shouldn't it be `Promise<void>` if we resolving awaiting promise like that:
`context.get(`bidders.prebid.bidsRefreshing.${callId}.resolve`)();`

src/ad-engine/services/slot-repeater.ts
```diff
@@ -72,11 +76,43 @@ function handleSlotRepeating(adSlot: AdSlot): boolean {
...
+	const adProduct: string = adSlot.config.adProduct;
+
+	if (adProduct === 'top_boxad') {
```
dmnsobczak: Once again we're introducing slot names into `ad-bidders` repository :<

src/ad-engine/services/slot-repeater.ts
```diff
@@ -72,11 +76,43 @@ function handleSlotRepeating(adSlot: AdSlot): boolean {
...
+		await context.get(`bidders.prebid.bidsRefreshing.incontent_boxad_1.finished`);
+		slotInjector.inject('incontent_boxad_1');
+		context.push('state.adStack', { id: 'incontent_boxad_1' });
```
dmnsobczak: Hmm, another idea: move it to mobile wiki and pass as a callback? Looks like it should be different in every app

src/ad-engine/services/slot-repeater.ts
```diff
@@ -72,11 +76,43 @@ function handleSlotRepeating(adSlot: AdSlot): boolean {
...
+	if (adProduct.indexOf('incontent_boxad') !== 0) {
+		return false;
+	}
+
+	return handleSlotRepeatingWhenBidsRefreshed(adSlot);
+}
+
+async function handleSlotRepeatingWhenBidsRefreshed(adSlot: AdSlot): Promise<boolean> {
+	const adProduct: string = adSlot.config.adProduct;
+	const currentBoxadNumber: number = parseInt(adProduct.split('_').pop(), 10);
```
dmnsobczak: We don't need `: number` because parseInt returns number by default.

src/ad-bidders/prebid/index.ts
bckatarzyna: 💎

src/ad-engine/services/slot-repeater.ts
bckatarzyna: I don't feel authorised to answer this question 😞

src/ad-engine/services/slot-repeater.ts
bckatarzyna: The whole function returns boolean afterall

PR #538, opened 2019-09-27
Hivi tests fix

emadry:


tests/common/ad-slot.js
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	get relativeLocationToViewport() {
+		return browser.execute(
+			(givenSelector) => document.querySelector(givenSelector).getBoundingClientRect().y,
+			this.selector,
+		);
+	}
+
+	constructor(config) {
```
dmnsobczak: You can just `constructor(private config) {}` :)

PR #540, opened 2019-09-30
ADEN-9385 cheshire cat 'weird_cat' fix

preludek:


src/ad-services/bill-the-lizard/index.ts
```diff
@@ -400,7 +400,9 @@ export class BillTheLizard {
 	 * @returns {number | undefined}
 	 */
 	getPreviousPrediction(startId, callIdBuilder, modelName) {
-		if (startId <= 1) {
+		const slotName = context.get('options.cheshireCatFirstSlot');
```
dmnsobczak: `slotName` is too general I think... maybe `firstSlotName`?

preludek: done

src/ad-services/bill-the-lizard/index.ts
```diff
@@ -400,7 +400,9 @@ export class BillTheLizard {
 	 * @returns {number | undefined}
 	 */
 	getPreviousPrediction(startId, callIdBuilder, modelName) {
-		if (startId <= 1) {
+		const firstSlotName = context.get('options.cheshireCatFirstSlot');
```
fraszczakszymon: I believe this is not the place for this logic. Bill the lizard module is not aware of any specific model. Also callId might not be related to slot name at all. I would look for fix in the cheshirecat implementation. What do you think?

src/ad-services/bill-the-lizard/index.ts
```diff
@@ -400,19 +400,13 @@ export class BillTheLizard {
 	 * @returns {number | undefined}
 	 */
 	getPreviousPrediction(startId, callIdBuilder, modelName) {
```
fraszczakszymon: What about
```suggestion
	getPreviousAndReusablePrediction(modelName) {
```
First two arguments are not needed. And This method name seems to be more descriptive. It will require bumping "Major" version because of changed API.

src/ad-services/bill-the-lizard/index.ts
preludek: This was changed. Now Bill the lizard is not aware of specific slots.

PR #541, opened 2019-09-30
ADEN-9392 Fix muthead on a devbox

nandy-andy:
A quick-fix for dev.

### DIFF
--- a/platforms/sports/targeting.ts
+++ b/platforms/sports/targeting.ts
@@ -9,7 +9,7 @@ export function getPageLevelTargeting(): Partial<Targeting> {
 		uap: 'none',
 		uap_c: 'none',
 		s0: 'gaming',
-		s1: domain.app,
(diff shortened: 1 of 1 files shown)


platforms/sports/targeting.ts
```diff
@@ -9,7 +9,7 @@ export function getPageLevelTargeting(): Partial<Targeting> {
 		uap: 'none',
 		uap_c: 'none',
 		s0: 'gaming',
-		s1: domain.app,
+		s1: domain.app !== 'fandom-dev' ? domain.app : 'muthead',
```
nandy-andy: Other solutions: A) splitting `sports` platform into two B) passing the information to AdEngine from the bootstrap code within MUT/FUT apps.

nandy-andy:
We'll fix it in cleaner way by creating two platforms.

PR #542, opened 2019-10-01
[ADEN-9371] Browsi clean up

Loczi94:


@types/window/index.d.ts
```diff
@@ -8,7 +8,6 @@ interface Window {
 	ads?: Ads;
 	adsQueue?: any;
 	apstag?: Apstag;
-	buildAdUnitString?: (arg0: string, arg1: any) => string;
```
bckatarzyna: You have to remove buildAdUnitString also from [here](https://github.com/Wikia/ad-engine/blob/2193885582a386f969ef6f636143cef115a357c2/src/ad-services/browsi/index.ts#L32)

Loczi94: I deleted whole file src/ad-services/browsi/index.ts.

bckatarzyna: Whoops, didn't notice, sorry for that 😄

PR #544, opened 2019-10-03
ADEN-9402 Adding VIDEO_AD_IMPRESSION for Porvata player

dmnsobczak:


src/ad-bidders/prebid/index.ts
```diff
@@ -23,6 +23,10 @@ eventService.on(events.VIDEO_AD_IMPRESSION, markWinningVideoBidAsUsed);
 eventService.on(events.VIDEO_AD_ERROR, markWinningVideoBidAsUsed);
 
 async function markWinningVideoBidAsUsed(adSlot: AdSlot): Promise<void> {
+	if (!adSlot) {
```
Bielik20: Why would someone call this function without adSlot, it is not optional param?
I would like to avoid null checks if possible.

dmnsobczak: This function is triggered by `VIDEO_AD_IMPRESSION` event and it's used in two places now. First is JWPlayer Factory which is related with Featured Video slot. Second is Porvata and technically Porvata is not related with any of the slot - you can even load Porvata without initing AdEngine. That's why we can't pass ad slot from event emitting place without other hack.

dmnsobczak: Or you gave me some idea. Changes applied :D

PR #548, opened 2019-10-07
ADEN 9390 IAS Porvata tracking

preludek:


src/ad-products/video/player/porvata/ias/ias-video-tracker.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+class IasVideoTracker {
+	init(
+		google,
+		adsManager: google.ima.AdsManager,
+		videoElement: HTMLElement,
+		config: IasTrackingParams,
+	): void {
+		loadScript().then(() => {
```
jwellborn-ias: Here's what I am pretty sure is happening. This code loads and initializes the googleImaVansAdapter all at once when `google.ima.AdsManagerLoadedEvent` fires. The ad is also being requested at (or soon after) this point so loading the IAS adapter introduces enough delay that the IAS init doesn't get called until after the ad is loaded. This is evidenced by the fact that the error only occurred when reloading the page with the video ad already in view. It doesn't occur for me when loading test page scrolled to the top and then scrolling down to view the ad. In that scenario, the adapter is initialized well before the video ad is loaded. My recommendation is to preload the adapter leaving only initialization for the `AdsManagerLoadedEvent`. Hope this helps.

PR #556, opened 2019-10-11
ADEN-9460 Adding "home" to s2 targeting parameter

dmnsobczak:


platforms/sports/targeting.ts
```diff
@@ -31,3 +31,9 @@ function getDomain(): { name: string; app: string; tld: string } {
 		tld: pieces.slice(-1).join(''),
 	};
 }
+
+function getPageType(): string {
+	const pathName = window.location.pathname;
+
+	return !pathName || pathName === '/' ? 'home' : 'main';
```
fraszczakszymon: What's the difference between home and main? Shouldn't it be home and article?

dmnsobczak: Yes and no. I left `main` just to keep old ad units and add/replace just `home` pages. I'll ask Nef.

PR #564, opened 2019-10-16
ADEN-9474 Disable UAP on old muthead

Bielik20:


platforms/shared/utils/get-page-type.ts
```diff
@@ -1,5 +1,11 @@
 export function getPageType(): string {
 	const pathName = window.location.pathname;
+	const hostname = window.location.hostname.toLowerCase();
+	const pieces = hostname.split('.').filter((piece) => piece !== 'www');
+
+	if (pieces[0] === 'old') {
+		return 'old';
```
jacekbj: Here you're putting Futhead/Muthead specific code (detecting "old" vs "new") in "shared" directory. It's also mnixing page type with platform version. Please separate it.

platforms/shared/setup/dynamic-slots/uap/curse-uap.setup.ts
```diff
@@ -0,0 +1,21 @@
...
+import { UapSetup } from './_uap.setup';
+import { UapRestrictor } from './uap-restrictor/uap-restrictor';
+
+@Injectable()
+export class CurseUapSetup implements UapSetup {
+	private firstCallSlotName = 'cdm-zone-01';
+
+	constructor(private uapRestrictor: UapRestrictor) {}
+
+	configureUap(): void {
```
jacekbj: Could the dependency on global `context` variable be removed by passing `isMobile` parameter to this method?

platforms/gamepedia/setup/dynamic-slots/gamepedia-dynamic-slots.setup.ts
```diff
@@ -1,12 +1,13 @@
...
 import { AdSlot, context, slotInjector, slotService } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
 
 @Injectable()
 export class GamepediaDynamicSlotsSetup implements DynamicSlotsSetup {
-	constructor() {}
+	constructor(private uapSetup: UapSetup) {}
 
 	configureDynamicSlots(): void {
+		this.uapSetup.configureUap();
```
jacekbj: One thing I don't get: "dynamic slots" are those injected into page by AdEngine or those with dynamic configuration? I can't see how `this.uapSetup.configureUap();` and `this.injectIncontentPlayer();` relate.

platforms/shared/utils/get-page-type.ts
Bielik20: It was requested by PO.

platforms/shared/setup/dynamic-slots/uap/curse-uap.setup.ts
Bielik20: It could but I would opt for removing arguments from startup class altogether.

platforms/gamepedia/setup/dynamic-slots/gamepedia-dynamic-slots.setup.ts
Bielik20: We discussed it with Szymon, if you see a better place for configureUap let us know. I feel like it is still better than context setup where it used to be.

platforms/shared/utils/get-page-type.ts
nandy-andy: Correct! Does a request from PO conflict with the idea of separating the logic? 🤔

PR #572, opened 2019-10-21
ADEN-9503 add warner function in utils/ warner in blueprints construtors

preludek:


platforms/shared/modes/ads/_ads.mode.ts
```diff
@@ -1,6 +1,12 @@
+import { utils } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
 
 @Injectable()
 export class AdsMode {
+	constructor() {
+		const warnGroup = 'AdsMode';
+		utils.warner('##### IOC - you are not implementing:', warnGroup);
```
Bielik20: To all:
- warnGroup should be first arguments.

Perhaps it is a good idea to create another method in platforms/shared like `iocDefaultWarning(className: string): void` that would use this warner to print message. What do you think?

preludek: I've created iocDefaultWarning and changed all the blueprints.

platforms/shared/modes/ads/_ads.mode.ts
```diff
@@ -1,6 +1,12 @@
 import { Injectable } from '@wikia/dependency-injection';
+import { iocDefaultWarning } from '../../utils/iocDefaultWarning';
```
fraszczakszymon: We're not using camel case pattern in filenames

preludek: I've changed the file name

PR #578, opened 2019-10-28
ADEN-9537 Fixing calling substr on undefined on JWPlayer no ad error

dmnsobczak:


src/ad-engine/video/vast-parser.ts
```diff
@@ -79,7 +79,7 @@ class VastParser {
 		let creativeId: string;
 		let lineItemId: string;
 		const vastParams: Dictionary<string> = queryString.getValues(
-			vastUrl.substr(1 + vastUrl.indexOf('?')),
+			vastUrl ? vastUrl.substr(1 + vastUrl.indexOf('?')) : '?',
```
nandy-andy: If there is no `?` in VAST url this code will pass the URL to the `queryString.getValues()` call. Is it safe? Example:
> var vastUrl = 'http://abcd.com'
> vastUrl ? vastUrl.substr(1 + vastUrl.indexOf('?')) : '?'
"http://abcd.com"

dmnsobczak: VAST builder basic URL contains `?` so it's impossible to pass URL without `?` there:
https://github.com/Wikia/ad-engine/blob/d75d0a1974f2b100778253fc084fc9799eaa3559/src/ad-engine/video/vast-url-builder.ts#L16

PR #581, opened 2019-11-06
ADEN-9535 Adding Triplelift bidder adapter

dmnsobczak:


platforms/futhead/fallback-config.json
```diff
@@ -27,6 +27,12 @@
 			"value": true
 		}
 	],
+	"icPrebidTriplelift": [
+		{
+			"apps": ["futhead", "gamepedia", "mobile-wiki", "muthead", "oasis"],
+			"value": false
```
preludek: Why set this value to false if every other prebid variable has fallback config set to true?

dmnsobczak: To remember about this fallback in the future. Then we can set it to `XX` once RevOps will pass granular rollout.

PR #585, opened 2019-11-13
ADEN-9581 added ucp to ad-engine platforms

preludek:


platforms/ucp/@types/window/index.d.ts
```diff
@@ -0,0 +1,5 @@
+interface Window {
+	ads?: MediaWikiAds;
+	mw: MediaWiki;
```
Bielik20: not used

platforms/ucp/@types/ucp-mw/index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @see https://doc.wikimedia.org/mediawiki-core/master/js/#!/api/mw.hook
+ */
+interface Hook {
+	add(handler: () => void): void;
+
+	fire(data: any): void;
+
+	remove(handler: () => void): void;
+}
```
Bielik20: It is not used anywhere.

platforms/ucp/@types/ucp-mw/index.d.ts
```diff
@@ -0,0 +1,39 @@
...
+		pageType: string;
+		wikiDbName: string;
+		wikiLanguage: string;
+	};
+}
+
+interface MediaWikiAds {
+	context?: MediaWikiAdsContext;
+	runtime: Runtime;
+}
```
Bielik20: Problem I with defining new types is that it is not inline with our architecture tenets:
- https://wikia-inc.atlassian.net/wiki/spaces/ADEN/pages/588776018/Architecture+Tenets

I don't know what to make of it right now but it is something to consider.

platforms/ucp/index.ts
```diff
@@ -0,0 +1,20 @@
...
+window.RLQ.push(async () => {
+	context.extend(basicContext);
+
+	const [consent, container]: [boolean, Container] = await Promise.all([
+		bootstrapAndGetCmpConsent(),
+		setupUcpIoc(),
+	]);
+	const platformStartup = container.get(PlatformStartup);
+
+	platformStartup.configure({ isOptedIn: consent, isMobile: getDeviceMode() === 'mobile' });
```
Bielik20: Is this correct way of getting device mode for ucp?

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		if (cid !== undefined) {
+			targeting.cid = cid;
+		}
+
+		return targeting;
+	}
+
+	private getDomain(): string {
```
Bielik20: What about `getDomain` from `shared`?

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,17 @@
...
+	configureDynamicSlots(): void {
+		this.injectTopLeaderboardPlaceholder();
+	}
+
+	private injectTopLeaderboardPlaceholder(): void {
+		const container = document.querySelector('.WikiaPage');
+		const topLeadeboardWrapper = document.createElement('div');
+		topLeadeboardWrapper.id = 'top_leaderboard';
+		const parentElement = container.parentNode;
+		parentElement.insertBefore(topLeadeboardWrapper, container);
```
Bielik20: consider placing `const` before followed by empty line and then operations.

platforms/ucp/@types/ucp-mw/index.d.ts
preludek: 🔪

platforms/ucp/@types/window/index.d.ts
preludek: 🔪

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
preludek: considered it and moved it

platforms/ucp/index.ts
preludek: we are doing it that way on other platforms

preludek: Anyway since there is no mobile yet I just set that to false.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
fraszczakszymon: Added `base` to getDomain method :+1:

platforms/ucp/@types/ucp-mw/index.d.ts
fraszczakszymon: Currently, context is exposed by backend and I don't see a point in requesting it separately on the frontend side

PR #586, opened 2019-11-14
ADEN-9380 Add geo setup to AdEngine

Bielik20:


platforms/shared/ensure-geo-cookie.ts
```diff
@@ -0,0 +1,65 @@
+import { CookieStorageAdapter } from '@wikia/ad-engine';
+
+interface GeoResponse {
+	continent_code: string;
+	country_code: string;
+	region: string;
+}
+
+interface GeoData {
```
dmnsobczak: `GeoData` interface is now duplicated?

PR #588, opened 2019-11-18
[ADEN-9528] Send comics=1 page level key-val to GAM

Loczi94:


src/ad-services/taxonomy/taxonomy-service.ts
```diff
@@ -33,6 +33,28 @@ export class TaxonomyService implements DelayModule {
...
+			this.resolveDelayPromise();
+			return {};
+		}
+
+		const isComicsRelated: string = await taxonomyServiceLoader.getComicsTag();
+		const comicsTag: AdTags = { txn_comics: [isComicsRelated] };
+
+		utils.logger(logGroup, 'taxonomy comics tag', comicsTag);
+
+		context.set('targeting.txn_comics', comicsTag['txn_comics']);
```
preludek: maybe simply [isComicsRelated] instead of comicsTag['txn_comics']?

spec/ad-services/taxonomy/taxonomy-service.spec.ts
```diff
@@ -63,3 +63,56 @@ describe('Taxonomy service', () => {
...
+		expect(context.get('targeting.txn_comics')).to.deep.equal(['[1]']);
+	});
+
+	it('does not fetch comics tag when service is disabled', async () => {
+		context.set('services.taxonomy.comics.enabled', false);
+
+		const fetchedComicsTag = await taxonomyService.configureComicsTargeting();
+
+		expect(geComicsTagStub.called).to.be.false;
+		expect(fetchedComicsTag).to.deep.equal({});
```
preludek: Why not use .to.be.empty instead of deep.equal?

src/ad-services/taxonomy/taxonomy-service.loader.ts
```diff
@@ -47,6 +48,47 @@ export class TaxonomyServiceLoader {
...
+				(response: Response) => {
+					if (response.status === 200) {
+						utils.logger(logGroup, 'successful response');
+
+						return response.json();
+					}
+
+					return {};
+				},
+				() => {
```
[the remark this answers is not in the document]
Loczi94: Good idea. 👍

src/ad-services/taxonomy/taxonomy-service.ts
Loczi94: Is there any benefit of changing it?

preludek: Just felt it would be more clear.

src/ad-services/taxonomy/taxonomy-service.loader.ts
```diff
@@ -96,6 +96,11 @@ export class TaxonomyServiceLoader {
 				return comicsTag;
 			});
 	}
+
+	reset(): void {
+		this.adTagsPromise = null;
+		this.comicsTagPromise = null;
```
preludek: We should keep adTags and need to reset data from targeting.

PR #591, opened 2019-11-22
ADEN-9633 Create top_boxad and bottom_leaderboard in UCP

preludek:


platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -1,18 +1,20 @@
...
-		const container = document.querySelector('.WikiaPage');
-		const topLeadeboardWrapper = document.createElement('div');
-		const parentElement = container.parentNode;
-
-		topLeadeboardWrapper.id = 'top_leaderboard';
-		parentElement.insertBefore(topLeadeboardWrapper, container);
+	private injectSlots(): void {
+		const slots = context.get('slots');
+		Object.values(slots).forEach((slot) => {
+			const { nextSibling, adProduct }: any = slot;
```
fraszczakszymon: Don't use `any` please. I'd rename this property to `nextSiblingSelector`. I'm not sure  what`adProduct` actually is. Use `slot.getSlotName()` please:
```suggestion
			const { nextSiblingSelector }: string = slot;
			
			if (nextSiblingSelector) {
				insertNewSlot(slot.getSlotName(), document.querySelector(nextSiblingSelector), true);
			}
```

preludek: I've changed the function so it doesn't use adProduct now. Also I've fixed the types for slots

platforms/ucp/modes/ucp-ads.mode.ts
```diff
@@ -21,7 +21,7 @@ export class UcpAdsMode implements AdsMode {
 		context.push('state.adStack', { id: 'top_leaderboard' });
 		context.push('state.adStack', { id: 'top_boxad' });
 		context.push('state.adStack', { id: 'incontent_boxad_1' });
-		context.push('state.adStack', { id: 'bottom_leaderboard' });
+		context.push('events.pushOnScroll.ids', 'bottom_leaderboard');
```
Loczi94: I pull all changes from your branch and it's still here.

Loczi94: It is still not removed. (I pull ale the changes from your branch.)

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -1,18 +1,19 @@
...
-		const topLeadeboardWrapper = document.createElement('div');
-		const parentElement = container.parentNode;
-
-		topLeadeboardWrapper.id = 'top_leaderboard';
-		parentElement.insertBefore(topLeadeboardWrapper, container);
+	private injectSlots(): void {
+		const slots: { [key: string]: SlotConfig } = context.get('slots');
+		Object.keys(slots).forEach((slotName) => {
+			if (slots[slotName].nextSiblingSelector) {
+				insertNewSlot(slotName, document.querySelector(slots[slotName].nextSiblingSelector), true);
```
Loczi94: Why you use bracket notation and then dot notation?
`slots[slotName].nextSiblingSelector`
I think it would be more consistent if you use only one notation type.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
Loczi94: I think you did not push your changes since I can't see them. I mean for example: using `slot.getSlotName()`.

fraszczakszymon: @Loczi94 it's my bad. `getSlotName` is not available here. It's not an `AdSlot` object, it's slot definition only.

platforms/ucp/modes/ucp-ads.mode.ts
preludek: It's because I went with the shorter way with all the lazy loaded slots shenaningans

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
preludek: as discussed, let's leave this as it is

src/ad-engine/services/slot-injector.ts
```diff
@@ -22,7 +22,7 @@ function findNextSuitablePlace(
 	return null;
 }
 
-function insertNewSlot(
+export function insertNewSlot(
```
Bielik20: https://github.com/Wikia/ad-engine/pull/602#discussion_r352049111

PR #596, opened 2019-11-26
[ADEN-9569] OneVideo bidder integration

Loczi94:


src/ad-bidders/prebid/adapters/one-video.ts
```diff
@@ -0,0 +1,38 @@
...
+			bids: [
+				{
+					bidder: this.bidderName,
+					params: {
+						site,
+						pubId,
+						video: {
+							playerWidth: 640,
+							playerHeight: 480,
+							mimes: ['video/mp4', 'application/javascript'],
```
bckatarzyna:
```suggestion
							mimes: ['video/mp4', 'video/x-flv', 'video/webm', 'video/ogg'],
```
We want only videos here, `application/javascript` would allow VPAIDs which could have specific behaviour or change things and we don't want it to happen.

src/ad-bidders/prebid/adapters/one-video.ts
```diff
@@ -0,0 +1,38 @@
...
+							protocols: [2, 5],
```
bckatarzyna:
```suggestion
							protocols: [2, 3, 5, 6],
```
It allows us to support VAST 3.0 protocol.
(numbers are  given in [pubmatic docs](http://prebid.org/dev-docs/bidders/pubmatic.html), search for `video.protocols`)

src/ad-bidders/prebid/adapters/one-video.ts
```diff
@@ -0,0 +1,38 @@
...
+							api: [1, 2],
```
bckatarzyna:
```suggestion
							api: [2],
```
I think it should be an empty array as we don't support VPAIDs, but let's keep it consistent with our other bidder adapters.
(numbers are  given in [pubmatic docs](http://prebid.org/dev-docs/bidders/pubmatic.html), search for `video.api`)

PR #598, opened 2019-11-27
ADEN-9577 Adding Criteo adapter

preludek:


src/ad-bidders/prebid/adapters/criteo.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	networkId: string;
+
+	get bidderName(): string {
+		return Criteo.bidderName;
+	}
+
+	prepareConfigForAdUnit(code, { sizes, networkId, zoneId }): PrebidAdUnit {
+		switch (code.toLowerCase()) {
+			case 'featured':
```
dmnsobczak: We need to rethink in all adapters how to mark video slots

platforms/futhead/bidders/prebid/criteo.ts
```diff
@@ -0,0 +1,55 @@
+import { DeviceMode } from '@platforms/shared';
+
+export function getCriteoContext(device: DeviceMode): any {
+	const context = {
+		desktop: {
+			enabled: false,
+			slots: {
+				'01_LB': {
+					sizes: [[728, 90], [970, 250]],
+					networkId: '3306',
```
dmnsobczak: If that's pretty much constant then maybe we can move it to the adapter like for A9 or Index Exchange and leave only sizes here?

src/ad-bidders/prebid/adapters/criteo.ts
fraszczakszymon: We could have some `isVideoSlot` flag in the slots-context

PR #599, opened 2019-11-28
ADEN-9415 Refactor JWPlayer

Bielik20:


src/ad-products/video/jwplayer-ads-factory.ts
```diff
@@ -35,6 +39,12 @@ const vastUrls = {
 const EMPTY_VAST_CODE = 21009;
 const log = (...args) => utils.logger('jwplayer-ads-factory', ...args);
 
+function setAttributes(element: HTMLElement, attributes: VastAttributes): void {
+	Object.keys(attributes)
+		.map((key) => ({ key, value: attributes[key] }))
```
jacekbj: why not only: `.forEach((key ) => element.setAttribute(key, attributes[key]));` :D?

src/ad-products/video/jwplayer-ads-factory.ts
```diff
@@ -176,88 +180,40 @@ function create(
...
+					playVideoAd('preroll');
 				}
 			}
 
 			prerollPositionReached = true;
 		});
 
+		function playVideoAd(position: 'midroll' | 'postroll' | 'preroll') {
+			tracker.adProduct = `${adProduct}-${position}`;
+			adSlot.setConfigProperty('audio', !player.getMute());
```
jacekbj: This was previously set even if ad was not played. Now it's hidden inside of `playVideoAd`. Should it stay here?

jacekbj: BTW why do we really need this piece of information on ad slot? 😅

src/ad-products/video/new/jwplayer-handler.ts
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	private beforePlay(): Observable<any> {
+		return this.streams.beforePlay$.pipe(
+			tap(({ depth }) => {
+				this.helper.updateVideoId();
+				this.helper.updateVideoDepth(depth);
+			}),
+			filter(({ depth }) => this.helper.shouldPlayPreroll(depth)),
+			mergeMap((payload) => this.helper.awaitIasTracking(payload)),
```
jacekbj: I will ask a lot of stupid questions so:
1. what if `awaitIasTracking` fails?

src/ad-products/video/new/jwplayer-helper.ts
```diff
@@ -0,0 +1,169 @@
...
+		this.adSlot.creativeSize = vastParams.size;
+	}
+
+	setSlotElementAttributes(vastParams: VastParams): void {
+		const attributes = vastDebugger.getVastAttributesFromVastParams('success', vastParams);
+		const element = this.adSlot.element;
+
+		Object.keys(attributes)
+			.map((key) => ({ key, value: attributes[key] }))
+			.forEach(({ key, value }) => element.setAttribute(key, value));
```
jacekbj: I've seen that somewhere :D

src/ad-products/video/new/jwplayer-helper.ts
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	emitVideoAdError(adErrorCode: number): void {
+		if (adErrorCode === EMPTY_VAST_CODE) {
+			this.adSlot.setStatus(AdSlot.STATUS_COLLAPSE);
+		} else {
+			this.adSlot.setStatus(AdSlot.STATUS_ERROR);
+		}
+
+		eventService.emit(events.VIDEO_AD_ERROR, this.adSlot);
```
jacekbj: You can use `this.adSlot.emit` instead: it calls `eventService.emit`.

src/ad-products/video/new/jwplayer-streams.ts
```diff
@@ -0,0 +1,128 @@
...
+		complete$,
+	};
+}
+
+function createJwpStream<TEvent extends keyof JWPlayerEventParams | JWPlayerNoParamEvent>(
+	jwplayer: JWPlayer,
+	event: TEvent,
+): Observable<TEvent extends keyof JWPlayerEventParams ? JWPlayerEventParams[TEvent] : void> {
+	return new Observable((observer) => {
+		jwplayer.on(event as any, (param) => observer.next(param));
```
jacekbj: No need for unsubscribing, right?

src/ad-products/video/new/jwplayer-streams.ts
```diff
@@ -0,0 +1,128 @@
...
+	});
+}
+
+function ensureEventTag<T>(adRequest$: JWPlayerStreams['adRequest$']): RxJsOperator<T, T> {
```
jacekbj: This and following functions require some comments to prep the reader for what they're about to read. It's very concise and elegant, yet demanding.

src/ad-products/video/new/jwplayer-streams.ts
```diff
@@ -0,0 +1,128 @@
...
+		source.pipe(
+			withLatestFrom(base$),
+			map(([adError, adRequest]) => _merge(adRequest, adError)),
+		);
+}
+
+function scanCorrelatorDepth<T>(): RxJsOperator<T, { depth: number; correlator: number }> {
+	return (source: Observable<T>) =>
+		source.pipe(
+			scan(
```
jacekbj: cool stuff

src/ad-products/video/new/jwplayer-streams.ts
```diff
@@ -0,0 +1,128 @@
...
+					depth: depth + 1,
+				}),
+				{ correlator: 0, depth: 0 },
+			),
+		);
+}
+
+function supplementCorrelatorDepth<T>(
+	beforePlay$: JWPlayerStreams['beforePlay$'],
+): RxJsOperator<T, { depth: number; correlator: number }> {
```
jacekbj: Do you think `{ depth: number; correlator: number }` could get a type?

src/ad-products/video/new/jwplayer-streams.ts
```diff
@@ -0,0 +1,128 @@
+import { VastParams, vastParser } from '@ad-engine/core';
```
jacekbj: This file is plain beauty ❤️

src/ad-products/video/player/porvata/ima/google-ima-player.ts
```diff
@@ -52,8 +52,11 @@ export class GoogleImaPlayer {
 
 	setVastAttributes(status: string, currentAd?: google.ima.Ad): void {
 		const playerElement: HTMLVideoElement = this.params.container.querySelector('.video-player');
+		const attributes = vastDebugger.getVastAttributes(this.vastUrl, status, currentAd);
 
-		vastDebugger.setVastAttributes(playerElement, this.vastUrl, status, currentAd);
+		Object.keys(attributes)
+			.map((key) => ({ key, value: attributes[key] }))
+			.forEach(({ key, value }) => playerElement.setAttribute(key, value));
```
jacekbj: Ditto

src/ad-products/video/new/jwplayer-manager.ts
```diff
@@ -0,0 +1,107 @@
...
+	private onPlayerReady(): Observable<PlayerReadyResult> {
+		const playerReady$ = this.communicator.actions$.pipe(ofType(jwpReady));
+
+		const loadPlugin$: Observable<any> = playerReady$.pipe(
+			take(1),
+			tap(() => {
+				this.loadMoatPlugin();
+				this.loadIasTrackerIfEnabled();
+			}),
+			filter(() => false),
```
jacekbj: Is it disabled by default?

src/ad-products/video/new/jwplayer-manager.ts
```diff
@@ -0,0 +1,107 @@
...
+		);
+
+		const createValues$ = playerReady$.pipe(
+			map(({ options, targeting, playerKey }) => {
+				const player: JWPlayer = window[playerKey];
+				const adSlot = this.createAdSlot(options, player);
+				const tracker = this.createTracker(options, adSlot);
+
+				tracker.register(player); // TODO: need to handle it separately
```
jacekbj: Is there a ticket for that TODO?

src/ad-products/video/new/jwplayer-manager.ts
jacekbj: Oh I see - `filter(() => false),` is to not return anything from `loadPlugin$` but it still needs to be triggered when returning `return merge(loadPlugin$, createValues$);`.

src/ad-products/video/jwplayer-ads-factory.ts
Bielik20: Yeah, you are right, it could be that way, I feel like it was copied and transformed along the way, thats way this syntax.

src/ad-products/video/jwplayer-ads-factory.ts
Bielik20: I don't know if we need it. I tried to _preserve_ everything I could. That `playVideoAd` we came up with Szymon. I was considering splitting it into separate methods but I am not sure. Maybe leave it as it is, and come back to it when we do JWPlayer tracking?

src/ad-products/video/new/jwplayer-handler.ts
Bielik20: Bad things shall happen. This stream will become useless (broken) but it shouldn't be different from what we have now. I feel like we could handle it somehow - perhaps just return payload - it should be fine. I would like someone else's input on that as well.

src/ad-products/video/new/jwplayer-helper.ts
Bielik20: Yeah but that previous code is suppose to be removed (at some point).

src/ad-products/video/new/jwplayer-helper.ts
Bielik20: Yeah, forgot about that. What is the actual way that we agreed on?

src/ad-products/video/new/jwplayer-streams.ts
Bielik20: There was no need before, we could figure something out should a need arise.

src/ad-products/video/new/jwplayer-streams.ts
Bielik20: I agree, I feel like this file is most demanding file of all, and took most of the writing time. I don't know about the comments though. I am open for discussion what kind of comments would you suggest?

src/ad-products/video/new/jwplayer-streams.ts
Bielik20: Yeah but I struggled how to name it :/
Do you have a suggestion?

src/ad-products/video/new/jwplayer-manager.ts
Bielik20: Yeah, maybe I am missing something. I basically want "tapOnce".

src/ad-products/video/new/jwplayer-manager.ts
Bielik20: I just made one https://wikia-inc.atlassian.net/browse/ADEN-9698

src/ad-products/video/jwplayer-ads-factory.ts
fraszczakszymon: But it introduces additional, not needed iteration? :)

src/ad-products/video/jwplayer-ads-factory.ts
fraszczakszymon: It is needed in order to calculate ad unit and key-vals properly. I'm also not sure if `audio` flag is needed when there are no ads. It would require additional QA of tracking when ads are disabled. @Bielik20 was it tested?

src/ad-products/video/new/jwplayer-helper.ts
jacekbj: It's better to go through ad slot. This way `AdSlot` becomes the central point, not `eventService`. Plus it takes only one argument :)

src/ad-products/video/player/porvata/ima/google-ima-player.ts
Bielik20: done

src/ad-products/video/new/jwplayer-helper.ts
Bielik20: done

src/ad-products/video/new/jwplayer-streams.ts
jacekbj: Good point, I have mixed feelings about types like `DepthAndCorrelator`. 
However, here I'd suggest just `interface VideoDepth`. @fraszczakszymon do you agree?

src/ad-products/video/jwplayer-ads-factory.ts
Bielik20: done

src/ad-products/video/new/jwplayer-handler.ts
fraszczakszymon: In my opinion, it is different from what we have now. Having this change, as you said, broken IAS tracking breaks the whole stream. On the other hand, in the current implementation, failed IAS does not impact playing preroll ad.

src/ad-products/video/jwplayer-ads-factory.ts
Bielik20: If automated tests do cover that, then yes. If not... well then no.

src/ad-products/video/new/jwplayer-handler.ts
fraszczakszymon: True, we didn't notice that. In my opinion, we should play the preroll ad also on failure.

src/ad-products/video/new/jwplayer-streams.ts
fraszczakszymon: I'd go for VideoDepth :+1:

src/ad-products/video/new/jwplayer-handler.ts
Bielik20: done

src/ad-products/video/new/jwplayer-helper.ts
Bielik20: I did change that, but I am not super positive about having this giant god class.

src/ad-products/video/new/jwplayer-manager.ts
Bielik20: done tapOnce

PR #606, opened 2019-12-12
ADEN-9627 Upgrade Prebid.js and enable CCPA module

dmnsobczak:


webpack.dev.js
```diff
@@ -1,5 +1,4 @@
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
```
fraszczakszymon: It seems that we no longer need this dependency in `package.json`

src/ad-engine/services/pbjs-factory.ts
```diff
@@ -13,17 +15,14 @@ class PbjsFactory {
...
 
-			if (libraryUrl) {
-				scriptLoader.loadScript(libraryUrl, 'text/javascript', true, 'first');
-			} else {
-				logger(
-					logGroup,
-					'Prebid library URL not defined. Assuming that window.pbjs will be loaded.',
-				);
-			}
+			scriptLoader.loadScript(libraryUrl || prebidLibraryUrl, 'text/javascript', true, 'first');
```
fraszczakszymon: What about `icPrebidVersion`?

webpack.dev.js
dmnsobczak: You're right! Removed.

src/ad-engine/services/pbjs-factory.ts
dmnsobczak: It will be added under `libraryUrl` in application.

PR #612, opened 2019-12-18
ADEN-9578 Add NoBid bidder

nandy-andy:


src/ad-engine/services/pbjs-factory.ts
```diff
@@ -2,7 +2,7 @@ import { logger, scriptLoader } from '../utils';
 import { context } from './context-service';
 
 const prebidLibraryUrl =
-	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/v2.44.0/20191213.min.js';
+	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/2.44.0/20191223.min.js';
```
nandy-andy: Missing `v` 🤦‍♂

platforms/futhead/bidders/prebid/nobid.ts
```diff
@@ -0,0 +1,55 @@
+import { DeviceMode } from '@platforms/shared';
+
+export function getNoBidContext(device: DeviceMode): any {
+	const context = {
+		desktop: {
+			enabled: false,
+			slots: {
+				'01_LB': {
+					sizes: [[728, 90], [970, 250]],
+					siteId: '21872987314',
```
dmnsobczak: For Futhead `siteId` should be `21874833814`

platforms/futhead/setup/context/prebid/futhead-prebid-config.setup.ts
```diff
@@ -17,6 +18,7 @@ export class FutheadPrebidConfigSetup implements PrebidConfigSetup {
 		context.set('bidders.prebid.appnexus', getAppNexusContext(mode));
 		context.set('bidders.prebid.criteo', getCriteoContext(mode));
 		context.set('bidders.prebid.indexExchange', getIndexExchangeContext(mode));
+		context.set('bidders.prebid.noBid', getNoBidContext(mode));
```
dmnsobczak: I guess Nobid is a one word so it should be `bidders.prebid.nobid` but who cares ¯\_(ツ)_/¯

src/ad-engine/services/pbjs-factory.ts
```diff
@@ -2,7 +2,7 @@ import { logger, scriptLoader } from '../utils';
 import { context } from './context-service';
 
 const prebidLibraryUrl =
-	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/v2.44.0/20191213.min.js';
+	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/v2.44.0/20191223.min.js';
```
dmnsobczak: We need to clean this up with release in January.

platforms/futhead/setup/context/prebid/futhead-prebid-config.setup.ts
nandy-andy: The adapter in Prebid.js is called `nobidBidAdapter`, so you're right, I guess. I'll think about how impactful the change is and if it's small I'll change it 🙇

src/platforms-shared/bidders-state-setup.ts
```diff
@@ -22,6 +22,7 @@ export function setupBidders(context: Context, instantConfig: InstantConfigServi
 		context.set('bidders.prebid.indexExchange.enabled', instantConfig.get('icPrebidIndexExchange'));
 		context.set('bidders.prebid.kargo.enabled', instantConfig.get('icPrebidKargo'));
 		context.set('bidders.prebid.lkqd.enabled', instantConfig.get('icPrebidLkqd'));
+		context.set('bidders.prebid.nobid.enabled', instantConfig.get('icPrebidNoBid'));
```
dmnsobczak: `icPrebidNobid`? 🙈

src/ad-bidders/prebid/adapters/nobid.ts
```diff
@@ -0,0 +1,29 @@
+import { PrebidAdapter } from '../prebid-adapter';
+import { PrebidAdSlotConfig } from '../prebid-models';
+
+export class NoBid extends PrebidAdapter {
```
dmnsobczak: `export class Nobid`? 🙈

src/platforms-shared/bidders-state-setup.ts
nandy-andy: I'd love to but it doesn't seem so easy. There is no EDIT button in ICBM I'd need to connect to MySQL and update the data directly in the database, wouldn't I?

dmnsobczak: Oh, you're right. Let's ignore it then 🤷‍♂

nandy-andy: Or... I can create a new ICBM var and we can remove the NoBid one in January ;) Better to do it now than later 👍

nandy-andy: Oh, it's not that easy... ICBM isn't case-sensitive 🙊

PR #616, opened 2019-12-24
ADEN-9699 JWPlayer ads on UCP

fraszczakszymon:


platforms/ucp/index.ts
```diff
@@ -16,5 +21,8 @@ window.RLQ.push(async () => {
 	const platformStartup = container.get(PlatformStartup);
 
 	platformStartup.configure({ isMobile: false });
+
+	communicator.dispatch({ type: '[AdEngine] Configured' });
```
Bielik20: I feel like we could move it `platformStartup.run`. `Communicator` could be injected with DI.

fraszczakszymon: I just moved it from `platfromStartup` because UCP is the only platform that uses post-quecast for now and I'd like to avoid blueprint class for other platforms. I'll add an appropriate comment.

Bielik20: I would place it at the end of `platformStartup.configure` because it says '[AdEngine] Configured' I feel like the source of the message should be the same for all the platforms (even though it is not always used). What do you think?

`setupPostQuecast` should stay where it is.

fraszczakszymon: Discussed offline

PR #618, opened 2020-01-08
ADEN-9601 Adding bidders DW tracking

dmnsobczak:


src/ad-engine/models/ad-slot.ts
```diff
@@ -84,6 +86,17 @@ export class AdSlot extends EventEmitter {
 	static AD_CLASS = 'gpt-ad';
 	static HIDDEN_CLASS = 'hide';
 
+	static generateSlotId(): string {
```
fraszczakszymon: What do you think about moving it to utils?

src/ad-engine/models/ad-slot.ts
```diff
@@ -136,6 +149,13 @@ export class AdSlot extends EventEmitter {
 			this.on(event.name, event.callback);
 		});
 
+		if (!this.config.uid) {
+			const uid = AdSlot.generateSlotId();
+
+			this.config.uid = uid;
+			context.set(`slots.${ad.id}.uid`, uid);
```
fraszczakszymon: What about slot repeating? Does it work properly?

src/ad-engine/models/ad-slot.ts
```diff
@@ -136,6 +149,13 @@ export class AdSlot extends EventEmitter {
 			this.on(event.name, event.callback);
 		});
 
+		if (!this.config.uid) {
+			const uid = AdSlot.generateSlotId();
+
+			this.config.uid = uid;
```
fraszczakszymon: This line seems not needed if we set uid on context below

src/ad-engine/services/slot-service.ts
```diff
@@ -235,6 +259,15 @@ class SlotService {
 			.map(([name, config]) => name);
 	}
 
+	/**
+	 * Returns slots with selected bidderAlias value.
```
fraszczakszymon:
```suggestion
	 * Returns slot names with selected bidderAlias value.
```

src/ad-engine/services/slot-service.ts
```diff
@@ -113,6 +113,30 @@ class SlotService {
 		return slotByPos;
 	}
 
+	/**
+	 * Get slot id by its name
+	 */
+	getId(id: string): string {
```
fraszczakszymon:
```suggestion
	getSlotId(id: string): string {
```
and jsdoc seems not needed

src/ad-engine/services/slot-service.ts
```diff
@@ -113,6 +113,30 @@ class SlotService {
...
+		let slotName = id;
+
+		if (!context.get(`slots.${slotName}`)) {
+			slotName = this.getSlotsByBidderAlias(id).shift();
```
fraszczakszymon: Why do we take slotName based on bidder alias? I believe all slots should have it's own slot uid

src/ad-engine/services/slot-service.ts
```diff
@@ -235,6 +259,15 @@ class SlotService {
 			.map(([name, config]) => name);
 	}
 
+	/**
+	 * Returns slots with selected bidderAlias value.
+	 */
+	getSlotsByBidderAlias(alias: string): string[] {
```
fraszczakszymon: I have mixed feelings about this method in this file. Bidders may not be used on some platforms.

src/ad-engine/services/slot-service.ts
fraszczakszymon: Actually, jsdoc seems not needed

src/ad-engine/services/slot-service.ts
fraszczakszymon:
```suggestion
	getSlotNamesByBidderAlias(alias: string): string[] {
```

src/ad-engine/services/slot-service.ts
```diff
@@ -113,6 +113,30 @@ class SlotService {
...
+			if (!slotName) {
+				return '';
+			}
+		}
+
+		let uid = context.get(`slots.${slotName}.uid`);
+
+		if (!uid) {
+			uid = AdSlot.generateSlotId();
+			context.set(`slots.${slotName}.uid`, uid);
```
fraszczakszymon: I have mixed feelings about setting uid in the getter. But I can't think of a better solution right now

src/ad-engine/services/slot-service.ts
fraszczakszymon: I'd move it to `bidder-tracking-middleware.ts`

src/ad-tracking/bidder-tracking-middleware.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	return next({
+		bid,
+		data: {
+			...data,
+			timestamp,
+			slot_id: slotId,
+			name: bid.bidderCode,
+			size: bid.size,
+			price: bid.cpm,
```
fraszczakszymon: Is this a real cpm or the value that we're passing to GAM? What do we want to track?

src/ad-tracking/bidder-tracking-middleware.ts
```diff
@@ -0,0 +1,32 @@
+import { context, slotService, utils } from '@ad-engine/core';
+import { AdBidderContext } from './bidder-tracker';
+
+function isBidOnTime(responseTime): string {
+	const biddersTimeout = context.get('bidders.timeout') || 2000;
+
+	return responseTime > biddersTimeout ? 'too_late' : 'on_time';
```
fraszczakszymon: This is not really true. Bidders timeout isn't proper for lazy loaded slots. In my opinion, we have to store the information whether a certain slot has been requested and this is the key flag for information whether the bid was on time or not.

src/ad-engine/models/ad-slot.ts
dmnsobczak: Cool idea! Moved.

src/ad-engine/models/ad-slot.ts
dmnsobczak: 👍

src/ad-engine/services/slot-service.ts
dmnsobczak: I removed this method completely for now. It's not used and it won't be until `adengadinfo` refactor.

src/ad-engine/services/slot-service.ts
dmnsobczak: I could extract it to separate method but this won't really make a difference since we can't call actual AdSlot object at this moment. Let's rethink that once we'll start using `slot_id` in `adengadinfo` tracking too.

src/ad-engine/services/slot-service.ts
dmnsobczak: No, that's not the part I removed :D Renamed!

src/ad-engine/services/slot-service.ts
dmnsobczak: Sorry... a habit. Removed.

src/ad-engine/services/slot-service.ts
dmnsobczak: Great suggestion of splitting `getSlotId` into two methods. Refactored.

src/ad-tracking/bidder-tracking-middleware.ts
dmnsobczak: For now it's the real CPM. I discussed this with @martynaut and for MVP this is fine.

src/ad-tracking/bidder-tracking-middleware.ts
dmnsobczak: I forgot about lazy loading. Logic changed.

src/ad-engine/models/ad-slot.ts
dmnsobczak: Oh yeah, you right. It will be a problem because cloned slots will have the same uid... but not really. I fixed this in `slotRepeater` to not miss this in the future but as a MVP FMR ad slots will report all bid events under first slot `uid`. In the future we'll decide if and how we want to link the received bid with ad slot, because this causes a lot of "cheshire-cat-type" problems.

src/ad-engine/models/ad-slot.ts
```diff
@@ -129,13 +132,18 @@ export class AdSlot extends EventEmitter {
 	constructor(ad: AdStackPayload) {
 		super();
 
+		this.pushTime = new Date().getTime();
```
fraszczakszymon: This is not "push" time for slots postponed by btfBlocker. I believe that here (link below) is the last point in the code before making the request to GAM for an ad:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/providers/gpt-provider.ts#L236

fraszczakszymon: Actually, there is another approach. We can test/subscribe to a new gpt event:
https://developers.google.com/doubleclick-gpt/reference#googletag.events.slotrequestedevent

fraszczakszymon: I'd go with second option. We can declare a new promise (`requested`) in set proper time once slotRequestedEvent is dispatched.

src/ad-engine/utils/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './iframe-builder';
 export * from './lazy-queue';
 export * from './logger';
 export * from './middleware';
+export * from './misc';
```
fraszczakszymon: What about s/misc/id-generator

fraszczakszymon:
One more thing left -- push time calculation. Also, it would be cool if we could have some unit tests for newly created functions. Anyway, LGTM

src/ad-engine/utils/index.ts
dmnsobczak: Renamed to `generate-unique-id.ts` thanks to Bielik 👍

src/ad-engine/models/ad-slot.ts
dmnsobczak: Great idea! Changed.

PR #622, opened 2020-01-16
ADEN-9785 Add A9 CCPA compliance

dmnsobczak:


src/ad-bidders/a9/index.ts
```diff
@@ -94,6 +97,18 @@ export class A9Provider extends BidderProvider {
 		return {};
 	}
 
+	private getCcpaIfApplicable(signalData: SignalData): Partial<A9CCPA> {
```
jacekbj: Mark argument `signalData` as optional.

PR #623, opened 2020-01-16
ADEN-9786 | Permutive MVP

bckatarzyna:


src/ad-services/permutive/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/* tslint:disable */
+const PROJECT_ID = '88ca3150-0f6f-482a-bbc1-2aa3276b3cab';
+const PUBLIC_API_KEY = '0006e595-b3f2-4dfc-b3a4-657eb42a74cf';
+// @ts-ignore
+const NAMESPACE = 'fandom';
+const logGroup = 'permutive';
+
+class Permutive {
+	call() {
```
jacekbj: return type missing

src/ad-services/permutive/index.ts
```diff
@@ -0,0 +1,51 @@
...
+		if (!context.get('services.permutive.enabled') || !context.get('options.trackingOptIn')) {
```
jacekbj: Do we need to check trackingOptIn? They're GDPR and CCPA compliant. We can leave this for now and ask them in e-mail/call.

src/ad-services/permutive/index.ts
```diff
@@ -0,0 +1,51 @@
...
+		this.setAddon('web', {});
+	}
+
+	configure(): void {
+		// @ts-ignore
+		!function(n,e,o,r,i){if(!e){e=e||{},window.permutive=e,e.q=[],e.config=i||{},e.config.projectId=o,e.config.apiKey=r,e.config.environment=e.config.environment||"production";for(var t=["addon","identify","track","trigger","query","segment","segments","ready","on","once","user","consent"],c=0;c<t.length;c++){var f=t[c];e[f]=function(n){return function(){var o=Array.prototype.slice.call(arguments,0);e.q.push({functionName:n,arguments:o})}}(f)}}}(document,window.permutive,PROJECT_ID,PUBLIC_API_KEY,{});
+	}
+
+	setTargeting(): void {
+		const g = window.localStorage.getItem('_pdfps');
```
jacekbj: We're changing their code so we may also change the variable name to a bit more meaningful.

src/ad-services/permutive/index.ts
bckatarzyna: According to Nef: Waiting for official statement they are safe

PR #628, opened 2020-01-21
ADEN-9789 Do not insert slot if sibling does not exist.

jacekbj:


platforms/shared/setup/_bidders-state.setup.ts
```diff
@@ -2,7 +2,7 @@ import { iocDefaultWarning } from '../utils/ioc-default-warning';
 
 export class BiddersStateSetup {
 	constructor() {
-		iocDefaultWarning('DynamicSlotsSetup');
+		iocDefaultWarning('BiddersStateSetup');
```
fraszczakszymon: 💎

PR #637, opened 2020-01-27
Remove top padding from top leaderboard

latata:
The 10px "space" above the ad will be added in UCP so it will be visible no matter if ads are loaded.

https://wikia-inc.atlassian.net/browse/IW-2882
The space between the global nav and the content or the top leaderboard is added in UCP (https://github.com/Wikia/unified-platform/pull/838). I left only top padding to have a space between the ad and the content.


nandy-andy:
Welcome in the AdEngine repository, @latata ! 🎉

platforms/ucp/styles.scss
```diff
@@ -4,7 +4,7 @@
 }
 
 #top_leaderboard {
-	padding: 10px 0;
+	padding: 0 0 10px;
```
nandy-andy: The change looks good and does exactly what the description of the PR says 👌 

Where does it come from, though? Do we have some designs or tickets we can refer to?

latata: Sorry @nandy-andy I should have marked this PR as "Work In Progress". Let me describe everything in the description

PR #647, opened 2020-01-30
ADEN-9822 JWPlayer Tracking Refactor

Bielik20:


src/ad-products/video/jwplayer/streams/jwplayer-stream-state.ts
```diff
@@ -0,0 +1,106 @@
...
+	mute: boolean;
+	adInVideo: 'bootstrap' | 'none' | 'preroll' | 'midroll' | 'postroll';
+}
+
+interface VideoDepth {
+	depth: number;
+	correlator: number;
+}
+
+export function createJwpStateStream(
```
jacekbj: If there is `stateless stream` then this could be `stateful stream`.

jacekbj:
Great job!
What I would like to ask you is to add some comments to each class briefly describing its duties.
It's well written and kind of obvious when you have the full picture during PR. However, it can be useful when coming back to that code later.

PR #648, opened 2020-01-31
ADEN-9817 Enable tracking on UCP

Bielik20:


@types/ads/index.d.ts
```diff
@@ -1,9 +1,48 @@
-interface Ads {
+interface MediaWikiAds {
```
dmnsobczak: This object is also available on Futhead on Muthead and there is no Media Wiki there. FYI.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
```diff
@@ -28,7 +28,7 @@ export class UcpTargetingSetup implements TargetingSetup {
 			s0: wiki.targeting.mappedVerticalName,
 			s1: this.getRawDbName(wiki),
 			s2: this.getAdLayout(wiki),
-			skin: 'ucp',
+			skin: 'oasis',
```
dmnsobczak: If parameter is defined in `ad-context.ts` file then why not remove them from here?

@types/ads/index.d.ts
```diff
@@ -1,9 +1,48 @@
...
+	targeting: MediaWikiAdsTargeting;
+}
+
+interface MediaWikiAdsOpts {
+	isAdTestWiki: boolean;
+	pageType: string;
+	showAds: boolean;
+}
+
+interface MediaWikiAdsTargeting {
```
dmnsobczak: 💎

@types/ads/index.d.ts
Bielik20: I know... those types are a mess. I will have to refactor it later.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
Bielik20: We are aiming to have a single source of that information - aka remove it from ad-context.

PR #651, opened 2020-02-06
ADEN-9859 Add rollout_tracking parameter to slot tracking

fraszczakszymon:


platforms/ucp/ad-context.ts
```diff
@@ -344,6 +344,7 @@ export const basicContext = {
 		rubiconDfp: true,
 		serverPrefix: 'wka1b',
 		wikiIdentifier: '_not_a_top1k_wiki',
+		rolloutTracking: 'ucp',
```
nandy-andy: Will we be able to pass more information here? Something like a string with coma-separated values like: `'ucp,oasis2,ae3.5'`?

PR #652, opened 2020-02-06
ADEN-9899 | Set permutive targeting value later

bckatarzyna:


src/ad-engine/video/vast-url-builder.ts
```diff
@@ -17,12 +17,27 @@ const baseUrl = 'https://pubads.g.doubleclick.net/gampad/ads?';
...
+		if (typeof value === 'function') {
+			params[key] = value();
+		} else if (value !== 'undefined' && value !== null) {
+			params[key] = value;
+		}
+	}
+
+	Object.keys(targeting).forEach((key) => {
+		setTargetingValue(key, targeting[key]);
+	});
```
fraszczakszymon: It may override slot-level/extra targeting. Can we keep the same order of passed values?

fraszczakszymon:
```suggestion
	});
	
	params = {
		...params,
		...slot.getTargeting(),
		...extraTargeting,
	};
```

src/ad-services/permutive/index.ts
```diff
@@ -34,12 +34,22 @@ class Permutive {
...
 		let permutiveTargeting = segments ? JSON.parse(segments) : [];
-
 		permutiveTargeting.push('_test');
-		context.set('targeting.permutive', permutiveTargeting.toString());
+
+		return permutiveTargeting;
+	}
+
+	setTargeting(): void {
+		context.set('targeting.permutive', this.getTargeting);
```
fraszczakszymon: It is possible that `googletag.pubads()` won't be available at this point. We should wrap it with callback pushed to `google.cmd` (then it's going to be executed as soon as gpt is ready).

fraszczakszymon:
```suggestion
		googletag.cmd.push(() => {
			context.set('targeting.permutive', this.getTargeting);
		});
```

PR #663, opened 2020-02-17
[ADEN-9862] handle showAds=false on UCP

Loczi94:


platforms/ucp/setup/context/wiki/ucp-wiki-context.setup.ts
```diff
@@ -7,6 +7,6 @@ export class UcpWikiContextSetup extends WikiContextSetup {
 	configureWikiContext(): void {
 		super.configureWikiContext();
 		// Make sure showAds is set to false in case of any falsy value
-		context.set('wiki.opts.showAds', !!context.get('wiki.opts.showAds'));
+		context.set('wiki.opts.showAds', context.get('wiki.opts.pageType') !== 'no_ads');
```
fraszczakszymon: Is this change needed? This is exactly what UCP `showAds`flag represents:
https://github.com/Wikia/unified-platform/blob/master/extensions/fandom/AdEngine/AdEngine.class.php#L111
https://github.com/Wikia/unified-platform/blob/master/extensions/fandom/AdEngine/AdPageType.php#L70
Having that we don't need to use hardcoded strings. What do you think?

platforms/ucp/modes/ucp-no-ads.mode.ts
```diff
@@ -0,0 +1,5 @@
+import { NoAdsMode } from '@platforms/shared';
+
+export class UcpNoAdsMode implements NoAdsMode {
```
Bielik20: This class doesn't do anything, it should be removed.

PR #664, opened 2020-02-19
ADEN-9761 Cleanup Prebid 3.0 transition

dmnsobczak:


dmnsobczak:
> Looks good but I guess we no longer need `icPrebidVersion` and setting it in context:
> 
> https://github.com/Wikia/ad-engine/blob/3d23b5126ff7cb5e5dbbbbce32e9e03842160f23/src/platforms-shared/bidders-state-setup.ts#L14

Fine, removed :D

src/ad-engine/services/pbjs-factory.ts
```diff
@@ -2,7 +2,7 @@ import { logger, scriptLoader } from '../utils';
 import { context } from './context-service';
 
 const prebidLibraryUrl =
-	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/v2.4.0/20191216.min.js';
+	'//origin-images.wikia.com/fandom-ae-assets/prebid.js/v3.2.0/20200114.min.js';
```
Loczi94: The AC is not fullfilled - 'Hardcoded version numbers are removed'. It's also hardcoded here: https://github.com/Wikia/ad-engine/blob/dev/examples/bidders/prebid/script.ts#L64

dmnsobczak: Nice! Removed.

PR #666, opened 2020-02-21
ADEN-9902 Fix code prio with wikia video

dmnsobczak:


@types/ads/index.d.ts
```diff
@@ -50,4 +50,9 @@ interface Runtime {
 	bab?: any;
 }
 
+interface VideoStatus {
+	wikiaVideo?: boolean;
+	videoPlayed?: boolean;
```
fraszczakszymon: What about `hasVideoOnPage`?

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
```diff
@@ -70,22 +70,34 @@ export class UcpTargetingSetup implements TargetingSetup {
...
-		let layout = wikiContext.targeting.pageType || 'article';
+	private getVideoStatus(): VideoStatus {
+		if (context.get('wiki.targeting.hasFeaturedVideo')) {
+			// Comparing with false in order to make sure that API already responds with "isDedicatedForArticle" flag
+			const isWikiaVideo =
+				context.get('wiki.targeting.featuredVideo.isDedicatedForArticle') === false;
+			const wikiaVideoPlayed = isWikiaVideo && window.canPlayVideo && window.canPlayVideo();
+
+			return {
+				wikiaVideo: isWikiaVideo,
```
fraszczakszymon: I'd love to avoid naming of our custom 'wikia video' thing, especially in the code. It's used only by us. What about `isDedicatedForArticle`?

src/ad-engine/utils/client.ts
```diff
@@ -142,6 +144,11 @@ class Client {
 
 		return userAgent.toLowerCase().indexOf('steam') > -1;
 	}
+
+	canPlayVideo(): boolean {
```
fraszczakszymon: Seems not needed, 🤞

@types/ads/index.d.ts
dmnsobczak: Renamed.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
dmnsobczak: Done!

src/ad-engine/utils/client.ts
dmnsobczak: Yep, removed.

PR #667, opened 2020-02-21
ADEN-9920 Add UAP to UCP

Bielik20:


platforms/ucp/setup/context/slots/ucp-slots-context.setup.ts
```diff
@@ -56,7 +60,7 @@ export class UcpSlotsContextSetup implements SlotsContextSetup {
 						],
 					},
 				],
-				defaultSizes: [[728, 90]],
+				defaultSizes: [[728, 90], [2, 2]],
```
dmnsobczak: Why 2x2? It's a mobile UAP size.

platforms/ucp/setup/context/slots/ucp-slots-context.setup.ts
```diff
@@ -222,5 +226,10 @@ export class UcpSlotsContextSetup implements SlotsContextSetup {
 		};
 
 		context.set('slots', slots);
+
+		// context.set('slots', slots) must be called first!
+		if (!this.targetingSetup.getVideoStatus().hasVideoOnPage) {
+			slotsContext.addSlotSize('hivi_leaderboard', [3, 3]);
```
dmnsobczak: What if HiVi leaderboard will be disabled? If we're switching to global as default let's remove the icVariable.
`slots.addSlotSize(context.get('custom.hiviLeaderboard') ? 'hivi_leaderboard' : 'top_leaderboard', [3, 3]);`

platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
```diff
@@ -1,11 +1,11 @@
 import { slotsContext, SlotsStateSetup } from '@platforms/shared';
+import { context } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
 
 @Injectable()
 export class UcpSlotsStateSetup implements SlotsStateSetup {
 	configureSlotsState(): void {
-		slotsContext.setState('hivi_leaderboard', true);
-		slotsContext.setState('top_leaderboard', true);
+		slotsContext.setState('hivi_leaderboard', !!context.get('options.hiviLeaderboard'));
```
dmnsobczak: `!!` is unnecessary.

dmnsobczak: Nevermind, it is, because context can return undefined, right?

platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
```diff
@@ -1,11 +1,11 @@
 import { slotsContext, SlotsStateSetup } from '@platforms/shared';
+import { context } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
 
 @Injectable()
 export class UcpSlotsStateSetup implements SlotsStateSetup {
 	configureSlotsState(): void {
-		slotsContext.setState('hivi_leaderboard', true);
-		slotsContext.setState('top_leaderboard', true);
```
dmnsobczak: Can we move setting `hivi_leaderboard` to ucp-dynamic-slots?

dmnsobczak: Also because slot state is a 3-state boolean it might be hard to understand why TLB is enabled if HiViTLB will be disabled:
`return this.slotStates[slotName] !== false;`

platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
Bielik20: Right

platforms/ucp/styles.scss
```diff
@@ -1,3 +1,43 @@
+@import 'src/ad-products/styles/mixin';
+@import 'src/ad-products/styles/breakpoints';
+@import 'src/ad-products/styles/uap-variables';
+@import './styles/navbar';
+@import './styles/button-close';
+@import './styles/advertisement-label';
+
+// TODO: Change CSS selector
```
fraszczakszymon: ?

platforms/ucp/templates/handlers/debug-transition-handler.ts
```diff
@@ -0,0 +1,15 @@
+import { TEMPLATE, TemplateStateHandler, TemplateTransition } from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+
+@Injectable()
+export class DebugTransitionHandler implements TemplateStateHandler {
```
fraszczakszymon: What about adding this handler when `?adengine_debug=1` is set?

platforms/ucp/styles.scss
jacekbj: Frankly I don't remember anymore :(

platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
jacekbj: @fraszczakszymon I think you're the best person to answer that ;)

fraszczakszymon: I'm not sure about moving it to dynamic-slots. Why? In my opinion, this is kind of initial state.

`top_leaderboard` is enabled because it is required to create DOM element for that slot (no matter if we're going to use it or not):
https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts#L20
However, I made it more explicit by setting hardcoded `true` in this place.

Then, `hivi_leaderboard` disables `top_leaderboard` in certain conditions:
https://github.com/Wikia/ad-engine/blob/13d98414e8094e5234765c2ced90935e40969c3c/platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts#L106-L116

PR #668, opened 2020-02-24
ADEN-9922 Add UCP Bfaa tracking

Bielik20:


src/ad-products/templates/uap/constants.ts
```diff
@@ -18,3 +18,6 @@ export const FAN_TAKEOVER_TYPES = ['uap', 'vuap'];
 
 export const BFAA_UNSTICK_DELAY = 3000;
 export const TLB_UNSTICK_DELAY = 2000;
+export const SLOT_FORCE_UNSTICK = 'force-unstick';
```
jacekbj: These variables may be also used in stickyTLB template, which is not an UAP. Could you move them on dir level higher?

PR #669, opened 2020-02-25
ADEN-9921

jacekbj:


platforms/ucp/templates/handlers/bfab-bootstrap-handler.ts
```diff
@@ -0,0 +1,41 @@
...
+	constructor(
+		@Inject(TEMPLATE.SLOT) private adSlot: AdSlot,
+		@Inject(TEMPLATE.PARAMS) private params: UapParams,
+	) {}
+
+	async onEnter(transition: TemplateTransition<'resolved' | 'impact'>): Promise<void> {
+		await slotTweaker.onReady(this.adSlot);
+		await this.awaitVisibleDOM();
+
+		this.adSlot.addClass('expanded-slot'); // TODO: Consider moving to a separate handler
```
Bielik20: Consider moving it above those awaits, so that people do not think it needs to await that stuff.

PR #672, opened 2020-02-26
ADEN-9758 | Add video config to IndexExchange adapter

bckatarzyna:


src/ad-bidders/prebid/adapters/index-exchange.ts
```diff
@@ -12,6 +12,42 @@ export class IndexExchange extends PrebidAdapter {
 	}
 
 	prepareConfigForAdUnit(code, { sizes, siteId }): PrebidAdUnit {
+		switch (code.toLowerCase()) {
```
fraszczakszymon: I know that we already did it in a few adapters, but in my opinion, we shouldn't use hardcoded (non-general) values here. What about setting an appropriate flag in the slot context and then:
```suggestion
		if (context.get(`slots.${code}.isVideo`)) {
			return this.getVideoConfig(code, siteId);
		}

		return this.getStandardConfig(code, { sizes, siteId });
```

src/platforms-shared/bidders-state-setup.ts
```diff
@@ -21,6 +21,10 @@ export function setupBidders(context: Context, instantConfig: InstantConfigServi
 		context.set('bidders.prebid.criteo.enabled', instantConfig.get('icPrebidCriteo'));
 		context.set('bidders.prebid.gumgum.enabled', instantConfig.get('icPrebidGumGum'));
 		context.set('bidders.prebid.indexExchange.enabled', instantConfig.get('icPrebidIndexExchange'));
+		context.set(
+			'bidders.prebid.indexExchangeFeatured.enabled',
```
fraszczakszymon: It seems not used anywhere

PR #695, opened 2020-03-05
ADEN-9983 Don't go to UAP sticked if viewability+timeout is counted

dmnsobczak:


platforms/ucp/templates/helpers/bfaa-helper.ts
```diff
@@ -90,4 +92,18 @@ export class BfaaHelper {
 
 		return () => window.scrollBy(0, elementOfReference.getBoundingClientRect().top - startValue);
 	}
+
+	isViewedAndDelayed(): Observable<unknown> {
+		const slotViewed: Promise<void> = this.adSlot.loaded.then(() => this.adSlot.viewed);
+
+		return from(slotViewed.then(() => utils.wait(this.getAdditionalStickinessTime())));
```
jacekbj: Consider using [concat](https://www.learnrxjs.io/learn-rxjs/operators/combination/concat) and [timer](https://www.learnrxjs.io/learn-rxjs/operators/creation/timer):
```
return concat(
 from(this.adSlot.loaded),
 from(this.adSlot.viewed),
 timer(this.getAdditionalStickinessTime()),
);
```

platforms/ucp/templates/handlers/bfaa-impact-handler.ts
```diff
@@ -30,7 +30,19 @@ export class BfaaImpactHandler implements TemplateStateHandler {
 		this.helper = new BfaaHelper(this.manipulator, this.params, this.adSlot, navbar);
 	}
 
-	async onEnter(transition: TemplateTransition<'sticky'>): Promise<void> {
+	async onEnter(transition: TemplateTransition<'sticky' | 'resolved'>): Promise<void> {
+		let shouldStick = true;
```
jacekbj: It's possible to remove the local variable `shouldStick`:
```
const isViewedAndDelayed$ = this.helper.isViewedAndDelayed().pipe(
  takeUntil(this.unsubscribe$),
  shareReplay(1),
);
```
and later.... (see next comment)

platforms/ucp/templates/handlers/bfaa-impact-handler.ts
```diff
@@ -58,7 +70,11 @@ export class BfaaImpactHandler implements TemplateStateHandler {
 				tap(() => {
 					const correction = this.helper.usePositionCorrection(this.footer);
 
-					transition('sticky').then(correction);
+					if (shouldStick) {
```
jacekbj: ...and:
```
filter(() => this.reachedResolvedSize()),
withLatestFrom(isViewedAndDelayed$),
tap(([_, shouldStick]) => {
  const correction = this.helper.usePositionCorrection(this.footer);
  if (shouldStick) {
    transition('sticky').then(correction);
  } else {
    transition('resolved').then(correction);
  }
})
```

platforms/ucp/templates/helpers/bfaa-helper.ts
```diff
@@ -90,4 +92,18 @@ export class BfaaHelper {
...
+	}
+
+	private getAdditionalStickinessTime(): number {
+		if (!isUndefined(this.params.stickyAdditionalTime)) {
+			return this.params.stickyAdditionalTime;
```
jacekbj: Can't wait for TS upgrade! `return return this.params?.stickyAdditionalTime ?? universalAdPackage.BFAA_UNSTICK_DELAY`

dmnsobczak: Yeah. I was joking about that at the beginning but this is very useful change :)

platforms/ucp/templates/helpers/bfaa-helper.ts
dmnsobczak: I actually moved this method without thinking about improvements. Thanks for the input!
...but it's not working. The tap is executed three times after every element of the concat :/

platforms/ucp/templates/handlers/bfaa-impact-handler.ts
dmnsobczak: Nice idea! Changed.

platforms/ucp/templates/handlers/bfaa-impact-handler.ts
dmnsobczak: 👍

platforms/ucp/templates/helpers/bfaa-helper.ts
jacekbj: Lets just leave it as is. I don't know if there is a better solution than 3x`switchMap`

PR #699, opened 2020-03-09
ADEN-9924 Add support for video in UAP HiVi BFAB.

jacekbj:


platforms/ucp/templates/handlers/bfab-bootstrap-handler.ts
```diff
@@ -20,8 +23,10 @@ export class BfabBootstrapHandler implements TemplateStateHandler {
 	async onEnter(transition: TemplateTransition<'resolved' | 'impact'>): Promise<void> {
 		this.adSlot.addClass('expanded-slot');
 		this.adSlot.getAdContainer().classList.add('iframe-container');
+		this.manipulator.element(this.adSlot.getElement()).setProperty('visibility', 'hidden');
```
Bielik20: Do we need manipulator in bootstrap?

platforms/ucp/templates/helpers/video-helper.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class VideoHelper {
+	constructor(
+		private manipulator: DomManipulator,
+		private params: UapParams,
+		private adSlot: AdSlot,
+	) {}
+
+	/**
+	 */
```
Bielik20: remove

platforms/ucp/templates/helpers/video-helper.ts
```diff
@@ -0,0 +1,69 @@
...
+	getImpactProgress(): number {
+		const slotWidth = this.adSlot.getElement().offsetWidth;
+		const slotResolvedHeight = slotWidth / this.params.config.aspectRatio.resolved;
+		const slotDefaultHeight = slotWidth / this.params.config.aspectRatio.default;
+
+		/* changes between 0 (impact, full height) to 1 (resolved size);
+		 * used to make video height transition smooth between
+		 * this.params.config.state.height.default
+		 * and this.params.config.state.height.resolved
+		 */
```
Bielik20: shouldn't it be a method description?

platforms/ucp/templates/handlers/bfab-bootstrap-handler.ts
jacekbj: It is supposed to remove the `visibility: hidden` when exiting the state. Is the overhead big enough that it justifies just doing it in `onLeave`?

Bielik20: It is fine but I remember in bfaa in initial state I wasn't using dom manipulator in a sense that `AdSlot` has a hide method, and every state called show method. But perhaps this is a better take on that problem.

PR #702, opened 2020-03-10
[ADEN-9988] Header bidding on UCP

Loczi94:


platforms/ucp/bidders/prebid/audience-network.ts
```diff
@@ -0,0 +1,6 @@
+export function getAudienceNetworkContext(): any {
```
bckatarzyna: It's not supported anymore, please 🔪 
You can also remove it from the app context at the same time :)

platforms/ucp/modes/ucp-no-ads.mode.ts
```diff
@@ -0,0 +1,20 @@
+import { NoAdsMode } from '@platforms/shared';
+import { Injectable } from '@wikia/dependency-injection';
+import { Communicator } from '@wikia/post-quecast';
+
+@Injectable()
+export class UcpNoAdsMode implements NoAdsMode {
+	handleNoAds(): void {
```
bckatarzyna: What about the other slots in no ads mode?

platforms/ucp/setup/context/prebid/ucp-prebid-config.setup.ts
```diff
@@ -0,0 +1,45 @@
...
+import { getWikiaContext } from '../../../bidders/prebid/wikia';
+import { getWikiaVideoContext } from '../../../bidders/prebid/wikia-video';
+
+@Injectable()
+export class UcpPrebidConfigSetup implements PrebidConfigSetup {
+	configurePrebidContext(): void {
+		context.set('bidders.prebid.aol', getAOLContext());
+		context.set('bidders.prebid.appnexus', getAppnexusContext());
+		context.set('bidders.prebid.appnexusAst', getAppnexusAstContext());
+		context.set('bidders.prebid.audienceNetwork', getAudienceNetworkContext());
```
bckatarzyna: 🔪

platforms/ucp/modes/ucp-no-ads.mode.ts
Loczi94: What do you mean? There is no ad slot in no ads mode. We set up only JWPlayer since it's needed to display no-ads-related videos.

bckatarzyna: [discussed offline] ✔️

PR #705, opened 2020-03-13
ADEN-9954 | Adjust permutive to SPA

bckatarzyna:


src/ad-services/permutive/index.ts
```diff
@@ -8,33 +8,49 @@ const NAMESPACE = 'fandom';
 const logGroup = 'permutive';
 
 class Permutive {
+	setup: () => void;
```
jacekbj: What do you thing about storing a boolean flag which indicates if service has already been initiated?
```
private isSetUp = false;

setup() {
  if (!isInitiated) {
    this.configure();
    this.loadScript();
    this.setTargeting();
    this.setNextSetup();
    this.isSetUp = true;
  }
}
```
To me it'm much more clear this way.

PR #706, opened 2020-03-13
ADEN-10001 Adding Outstream template to UCP

dmnsobczak:


platforms/shared/platform-startup.ts
```diff
@@ -44,8 +44,8 @@ export class PlatformStartup {
 	configure(args: PlatformStartupArgs): void {
 		this.wikiContextSetup.configureWikiContext();
 		this.baseContextSetup.configureBaseContext(args.isMobile);
-		this.targetingSetup.configureTargetingContext();
 		this.slotsContextSetup.configureSlotsContext();
+		this.targetingSetup.configureTargetingContext();
```
jacekbj: Is the order important? If yes then please add a note in a comment.

platforms/ucp/setup/context/slots/ucp-slots-context.setup.ts
```diff
@@ -226,10 +220,7 @@ export class UcpSlotsContextSetup implements SlotsContextSetup {
 		};
 
 		context.set('slots', slots);
-
-		// context.set('slots', slots) must be called first!
-		if (!this.targetingSetup.getVideoStatus().hasVideoOnPage) {
```
jacekbj: Shouldn't that disable UAP in video pages?

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
```diff
@@ -89,16 +89,21 @@ export class UcpTargetingSetup implements TargetingSetup {
 
 	private getAdLayout(targeting: MediaWikiAdsTargeting): string {
 		let layout = targeting.pageType || 'article';
+		const videoStatus = this.getVideoStatus();
+
+		context.set('custom.hasFeaturedVideo', !!videoStatus.hasVideoOnPage);
+		context.set(
+			'custom.hasIncontentPlayer',
+			!context.get('custom.hasFeaturedVideo') &&
+				document.querySelector(context.get('slots.incontent_player.insertBeforeSelector')) !== null,
```
jacekbj: Comparison to `null` is not necessary but ok.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
```diff
@@ -89,16 +89,21 @@ export class UcpTargetingSetup implements TargetingSetup {
 
 	private getAdLayout(targeting: MediaWikiAdsTargeting): string {
 		let layout = targeting.pageType || 'article';
+		const videoStatus = this.getVideoStatus();
+
```
jacekbj: Please use a single variable `const hasFeaturedVideo: boolean = !!videoStatus.hasVideoOnPage`.
`context.get('custom.hasFeaturedVideo')` appears a couple of times and it's easier to lookup a single variable (i.e. with IDE) than look for a place where `custom.hasFeaturedVideo` is set in context.

Same applies for `hasIncontentPlayer`.

This would not occur if this method was split into two: setting variables in context and building the actual ad layout variable.

platforms/ucp/templates/porvata-template.ts
```diff
@@ -0,0 +1,19 @@
+import { slotsContext } from '@platforms/shared';
+import { NavbarManager, PorvataTemplate, templateService } from '@wikia/ad-engine';
+
+const getPorvataConfig = () => {
```
jacekbj: Return type of this function should be `PorvataTemplateConfig`.

platforms/ucp/templates/porvata-template.ts
```diff
@@ -0,0 +1,19 @@
...
+	const navbarManager = new NavbarManager(document.getElementById('globalNavigation'));
+
+	return {
+		inViewportOffsetTop: navbarManager.getHeight(),
+		isFloatingEnabled: true,
+		onInit: (adSlot, params) => {
```
jacekbj: If typing the return type does not help, please add also types for this function.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -27,34 +28,46 @@ export class UcpDynamicSlotsSetup implements DynamicSlotsSetup {
...
 				);
 			});
 	}
 
+	private appendIncontentPlayer(): void {
+		const icpSlotName = 'incontent_player';
+
+		context.set(`slots.${icpSlotName}.customFiller`, 'porvata');
+		context.set(`slots.${icpSlotName}.customFillerOptions`, {
+			enableInContentFloating: true,
```
jacekbj: Is there a type for `customFillerOptions` that we could use here? That would make it easier to reason what this variable is later on.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -27,34 +28,46 @@ export class UcpDynamicSlotsSetup implements DynamicSlotsSetup {
...
 				this.appendRotatingSlot(
-					'incontent_boxad_1',
+					icbSlotName,
 					slotConfig.repeat.slotNamePattern,
 					document.querySelector(slotConfig.parentContainerSelector),
 				);
 			});
 	}
 
+	private appendIncontentPlayer(): void {
```
jacekbj: Does it really append incontent_player slot?
It's more like configureIncontentPlayerPorvataFiller.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
```diff
@@ -27,34 +28,46 @@ export class UcpDynamicSlotsSetup implements DynamicSlotsSetup {
...
 		this.appendIncontentBoxad(slots['incontent_boxad_1']);
+		this.appendIncontentPlayer();
 	}
 
 	private appendIncontentBoxad(slotConfig: SlotConfig): void {
+		const icbSlotName = 'incontent_boxad_1';
 		const communicator = new Communicator();
 
+		if (context.get('custom.hasFeaturedVideo')) {
+			context.set(`slots.${icbSlotName}.defaultSizes`, [300, 250]);
```
jacekbj: Is this mentioned in https://wikia-inc.atlassian.net/wiki/spaces/ADEN/pages/146246794/Code+level+priorities?

platforms/ucp/templates/porvata-template.ts
```diff
@@ -0,0 +1,19 @@
...
+			slotsContext.setupSlotVideoAdUnit(adSlot, params);
+			params.viewportHookElement = document.getElementById('INCONTENT_WRAPPER');
+		},
+	};
+};
+
+export function registerPorvataTemplate(): void {
+	templateService.register(PorvataTemplate, getPorvataConfig());
```
jacekbj: Why not move this line to `ucp-templates-setup`? Then this file could be called `outstream-config` of `incontent-player-config`.

platforms/shared/platform-startup.ts
dmnsobczak: Yes, I'm using `context.get('slots.incontent_player.insertBeforeSelector')` in the targeting context setup so I need this filled up before it inside slots context setup

platforms/ucp/setup/context/slots/ucp-slots-context.setup.ts
dmnsobczak: It should. I moved it here:
https://github.com/Wikia/ad-engine/pull/706/files#diff-7ad31548b20a2fe42aecad801b0a6889R108
Pretty static slots context is I think not the best place for that. Also I need some information which are not available yet.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
dmnsobczak: It's not but I found this code in old MediaWiki and I decided to copy it.
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/setup.js#L87
Let's ask dinosaurs: @fraszczakszymon, @nandy-andy any comment on that?

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
dmnsobczak: It was before I refactored it and forgot to change its name 🙈 Changed!

platforms/ucp/templates/porvata-template.ts
dmnsobczak: Right!

platforms/ucp/templates/porvata-template.ts
dmnsobczak: It helped :)

platforms/ucp/templates/porvata-template.ts
dmnsobczak: Yeah, it makes sense. Moved.

platforms/ucp/setup/dynamic-slots/ucp-dynamic-slots.setup.ts
dmnsobczak: It's actually a `Partial<PorvataGamParams>`. Type added!

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
dmnsobczak: Right.

platforms/ucp/setup/context/targeting/ucp-targeting.setup.ts
dmnsobczak: I always fight with myself if two usages is enough for a new variable or if I should wait for at least three 🙃 
Modified!

PR #715, opened 2020-03-18
ADEN-10061 Templates 2.0 Additional Dependencies

Bielik20:


platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+		private domListener: DomListener,
+		private manipulator: DomManipulator,
+		private manager: VideoDomManager,
+	) {}
+
+	async onEnter(): Promise<void> {
+		this.playerRegistry.video$
+			.pipe(
+				switchMap(({ player }) =>
+					merge(this.domListener.resize$).pipe(
```
jacekbj: Why `merge`?

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+						startWith({}),
+						map(() => player),
+					),
+				),
+				tap((player) => this.manager.setDynamicVideoImpactSize(player, 0)),
```
jacekbj: Please remind me why we're using `tap` instead of putting the final action in `subscribe` callback?

platforms/ucp/templates/handlers/video-ctp-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+export class VideoCtpHandler implements TemplateStateHandler {
+	constructor(private playerRegistry: PlayerRegistry) {}
+
+	async onEnter(transition: TemplateTransition<'impact'>): Promise<void> {
+		this.playerRegistry.video$
+			.pipe(
+				take(1),
+				filter(({ params }) => !params.autoPlay),
+				switchMap(({ player }) => fromEvent(player, 'wikiaAdStarted').pipe(take(1))),
+				tap(() => transition('impact', { allowMulticast: true })),
```
jacekbj: Handler name suggests it's generic. However, with the value of transition defined here it's strictly bound to UAP (or in case of new naming: bfaX) template.
I want us to have that in mind. The same goes for `VideoCompletedHandler`.

platforms/ucp/templates/handlers/video-resolved-handler.ts
```diff
@@ -0,0 +1,39 @@
+import { DomListener, TemplateStateHandler } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { merge, Subject } from 'rxjs';
+import { map, startWith, switchMap, takeUntil, tap } from 'rxjs/operators';
+import { DomManipulator } from '../helpers/manipulators/dom-manipulator';
+import { PlayerRegistry } from '../helpers/player-registry';
+import { VideoDomManager } from '../helpers/video-dom-manager';
+
+@Injectable()
+export class VideoResolvedHandler implements TemplateStateHandler {
```
jacekbj: The only difference between`BfaaVideoStickyHandler`, `BfaaVideoImpactHandler` and `VideoResolvedHandler` is what size is set for player.

What if in such case we could create a handler like `createVideoResizeHandler({mode: 'dynamic' | 'resolved', progress?: 0})`. What do you think about it?

platforms/ucp/templates/handlers/video-resolved-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+	private unsubscribe$ = new Subject<void>();
+
+	constructor(
+		private playerRegistry: PlayerRegistry,
+		private domListener: DomListener,
+		private manipulator: DomManipulator,
```
jacekbj: `manipulator` is not used in any of the 3 video handlers mentioned in previous comment.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+		private manipulator: DomManipulator,
+		private manager: VideoDomManager,
+	) {}
+
+	async onEnter(): Promise<void> {
+		this.playerRegistry.video$
+			.pipe(
+				switchMap(({ player }) =>
+					merge(this.domListener.resize$).pipe(
+						startWith({}),
```
jacekbj: I don't like the idea of using `startWith` to trigger one initial execution. For me it would make more sense to move the `tap` part a bit higher up (before `switchMap`) to indicate that it's a one-off event.

Yes, that would result in `(player) => this.manager.setDynamicVideoImpactSize(player, 0)` included twice (one in `tap`, one in `subscribe` callback).

The advantage is removing the question why `this.domListener.resize$` is initialized with an empty object when it should be an `Event` instance.

platforms/ucp/templates/handlers/resolved-handler.ts
```diff
@@ -0,0 +1,35 @@
+import { DomListener, TemplateStateHandler } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { Subject } from 'rxjs';
+import { startWith, takeUntil, tap } from 'rxjs/operators';
+import { DomManipulator } from '../helpers/manipulators/dom-manipulator';
+import { UapDomManager } from '../helpers/uap-dom-manager';
+
+@Injectable()
+export class ResolvedHandler implements TemplateStateHandler {
+	private unsubscribe$ = new Subject<void>();
```
jacekbj: I wish at some point we could just inherit that from a base class :D

platforms/ucp/templates/helpers/uap-dom-reader.ts
```diff
@@ -0,0 +1,31 @@
+import { AdSlot, TEMPLATE, UapParams } from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+
+@Injectable()
+export class UapDomReader {
```
jacekbj: Why `UapDomManager` and `UapDomReader` are two separate classes? They're very much related and IMO this split to "getters" and "setters" feels artificial.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
Bielik20: Remnant from bfaa video impact handler - good catch, will fix it.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
Bielik20: Because there are scenarios where actions are not at the end of the stream and I wanted it to be unified across all streams. To not confuse people why sometimes it is subscribe and sometimes tap. But I can see I failed with avoiding confusion :(

platforms/ucp/templates/handlers/video-ctp-handler.ts
Bielik20: Since there are not other templates as of right now. I would leave it as it is. But I agree it may require renaming after new templates are added. Still I would opt against premature optimization.

platforms/ucp/templates/handlers/video-resolved-handler.ts
Bielik20: You would have a function what returns a class? Well I am not sure, feels like it would obscure some things. I would sooner to something like that in template dependencies
```ts
[
...,
{ bind: MODE_SYMBOL, value: 'dynamic'
]
```

And then inject symbol. Still not 100%, I fear it would grow in complexity fast if we do that. Although I would ask other because they are my primary concern.

platforms/ucp/templates/handlers/video-resolved-handler.ts
Bielik20: Yeah, it only does `restore`, in 10066 I moved restoring to yet another handler.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
Bielik20: > The advantage is removing the question why this.domListener.resize$ is initialized with an empty object when it should be an Event instance.

We are doing it all the time with scroll as well. I feel like it is becoming convention.

I am in a tight spot here. I have unified it to this form because I find it more readable, but now I can see you prefer previous one. I don't like the same operation being called twice.

```ts
this.domListener.resize$.pipe(startWith({}))
```

basically means "do it once and then on resize", and this is what we want, don't we?

platforms/ucp/templates/handlers/resolved-handler.ts
Bielik20: I you really want we could make in injectable.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
jacekbj: Well it's exactly what we want: do it on enter and on resize.
Putting it in one stream makes it more succinct  but actually obscures what is intended.

platforms/ucp/templates/handlers/resolved-handler.ts
jacekbj: Aren't we going to far with injection? It feels like a perfect use case for inheritance.

platforms/ucp/templates/helpers/uap-dom-reader.ts
Bielik20: It speaks more when you look at the Handler you know what it really requires.
But really change comes from times before `template dependencies` and I wanted to reduce number of things components that just want to read need to inject, or create like DomManipulator. Now those arguments are obsolete. I would leave it for now, but should you have one day a strong desire to change it, I am ok with that.

platforms/ucp/templates/bfaa-template.ts
```diff
@@ -23,20 +32,28 @@ export function registerBfaaTemplate(registry: TemplateRegistry): Observable<Tem
...
 				BfaaStickyHandler,
-				BfaaStickyDurationHandler,
+				BfaaStickyDecisionHandler,
 				CloseButtonHandler,
-				BfaaStickyVideoHandler,
+				BfaaVideoStickyHandler,
 			],
-			transition: [BfaaTransitionHandler, BfaaTransitionVideoHandler],
-			resolved: [BfaaResolvedHandler, BfaaResolvedVideoHandler],
+			transition: [BfaaStickyHandler, BfaaTransitionHandler, BfaaVideoTransitionHandler],
```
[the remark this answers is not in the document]
Bielik20: Yes, because transition state shared stickiness with sticky. No code duplication for the win.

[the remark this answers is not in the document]
Bielik20: That's why I feel like this is our wrong that we called (some) Handlers with the name of the state they are in.

platforms/ucp/templates/handlers/bfaa/bfaa-config-handler.ts
```diff
@@ -49,5 +48,7 @@ export class BfaaConfigHandler implements TemplateStateHandler {
 		}
 	}
 
-	async onLeave(): Promise<void> {}
+	async onLeave(): Promise<void> {
+		this.adSlot.show();
```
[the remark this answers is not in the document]
Bielik20: I was in almost all Hanlders, and that almost actually worries me. It is safer that way. `BfaaConfigHandler` hides it and so it also unhides it.

platforms/ucp/templates/handlers/bfaa/bfaa-sticky-decision-handler.ts
```diff
@@ -47,6 +50,5 @@ export class BfaaStickyDurationHandler implements TemplateStateHandler {
 
 	async onLeave(): Promise<void> {
 		this.unsubscribe$.next();
-		this.manipulator.restore();
```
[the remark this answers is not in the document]
Bielik20: Perhaps this handler doesn't use it anymore.

platforms/ucp/templates/bfaa-template.ts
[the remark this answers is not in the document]
jacekbj: It's all a result of the choice to go with a modified finite state machine implementation.
FSM expects code to emit actions and makes decision which state to transition to (if any). Our implementation emits transitions directly. This makes handlers bound to specific machine.

platforms/ucp/templates/handlers/resolved-handler.ts
Bielik20: I would actually wouldn't inject nor inherit it. I like it is explicitly defined in here. It is a a one line after all.

platforms/ucp/templates/bfaa-template.ts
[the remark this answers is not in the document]
Bielik20: It only bounds them so that machine they are in have states defined in that Handlers as transitionable. In other words the states defines for transition method. 

I kind of regret now not calling them Plugins instead of Handlers. Handlers are not states. States are defined in a template, and they only now that are is an array Handlers that have `onEnter` and `onLeave` methods, and cause state change by calling `transition` method. Handlers itself provide some functionality to the state they are in. Functionalities like "Stick", "Go to impact on replay", "Go to transition after time" etc.

platforms/ucp/templates/handlers/bfab/bfab-video-impact-handler.ts
Bielik20: Hmm, I can see what you are saying be I am not necessary agreeing that your version is more readable. I would say they both are kind of hard. So let me propose an alternative solution. Lets wrap those kind of operations:
```ts
({ player }) =>
					merge(this.domListener.scroll$, this.domListener.resize$).pipe(
						startWith({}),
						map(() => player),
					),
```
within a method that describes what it does, like... `fireOnceAndOnSrollOrResize`. What do you think. I could even write an operator to do it, since it is such a common use case. How would you call such operator? It would replace switch map. This:
```ts
				switchMap(({ player }) =>
					merge(this.domListener.scroll$, this.domListener.resize$).pipe(
						startWith({}),
						map(() => player),
					),
				),
```
would become:
```ts
nameOfOperator(merge(this.domListener.scroll$, this.domListener.resize$)),
```

What do you think?

Bielik20: I have created operator `startAndRespondTo`, it should be more readable now.

platforms/ucp/templates/handlers/advertisement-label-handler.ts
```diff
@@ -1,7 +1,7 @@
 import { AdSlot, AdvertisementLabel, TEMPLATE, TemplateStateHandler } from '@wikia/ad-engine';
 import { Inject, Injectable } from '@wikia/dependency-injection';
 
-@Injectable()
+@Injectable({ autobind: false })
```
jacekbj: Could you explain how `autobind` works? I skimmed through the dependency-injection library but could not get a grasp of it.

platforms/ucp/templates/handlers/dom-cleanup-handler.ts
```diff
@@ -0,0 +1,17 @@
+import { TemplateStateHandler } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { DomManipulator } from '../helpers/manipulators/dom-manipulator';
+
+/**
+ * Resets DomManipulator on leave
+ */
+@Injectable({ autobind: false })
+export class DomCleanupHandler implements TemplateStateHandler {
```
jacekbj: I have very mixed feelings about this plugin. It's going to be easy to forget about it.

platforms/ucp/templates/handlers/resolved-handler.ts
jacekbj: It's actually more than one line: it's 
```
private unsubscribe$ = new Subject<void>();

and

async onLeave(): Promise<void> {
    this.unsubscribe$.next();
}
```
Repeated over multiple files.

package.json
```diff
@@ -89,7 +89,7 @@
 		"tslint-config-airbnb": "5.11.1",
 		"tslint-config-prettier": "1.18.0",
 		"tslint-eslint-rules": "5.4.0",
-		"typescript": "3.6.3",
+		"typescript": "3.8.3",
```
jacekbj: That just kind of sneaked in, didn't it :)?

platforms/ucp/templates/bfaa-template.ts
```diff
@@ -43,17 +46,33 @@ export function registerBfaaTemplate(registry: TemplateRegistry): Observable<Tem
...
-		[PlayerRegistry, DomManipulator, UapDomManager, UapDomReader, VideoDomManager],
+		[
+			ScrollCorrector,
+			PlayerRegistry,
+			DomManipulator,
+			UapDomManager,
+			UapDomReader,
+			VideoDomManager,
+			StickinessTimeout,
+			{ bind: STICKINESS_TIMEOUT_DEFAULT, value: universalAdPackage.BFAA_UNSTICK_DELAY },
```
jacekbj: Wow how does that work?

platforms/ucp/templates/handlers/dom-cleanup-handler.ts
jacekbj: However I can see the problem that it is solving: calling `this.manipulator.restore()` multiple times in the same state from it's handlers.

platforms/ucp/templates/bfaa-template.ts
jacekbj: I see now. I'm really sorry for being a pain in the neck but don't you feel we removed the context in favor of a much more complicated solution?

jacekbj:
I am very impressed with the level of code expertise you achieved with DI. It definitely solves many problems we had.

On the other hand, it's increasing the code complexity. I don't know where is the tipping point at which we should stop.

IMO we can proceed with QA for this ticket. I just don't like the idea of having TS upgraded. If you want to keep this part here then please add info to ticket that a more thorough testing has to be executed.

platforms/ucp/templates/handlers/resolved-handler.ts
Bielik20: Yes, but you cannot do anything about onLeave, you could call super, but what seems even worst.

jacekbj: Out of curiosity, why is calling `super` bad?

platforms/ucp/templates/handlers/advertisement-label-handler.ts
Bielik20: Sure ;)
```ts
const container = new Container();

container.get(ClassWithAutoBind) // works
container.get(ClassWithoutAutoBind) // throws

container.bind(ClassWithoutAutoBind)
container.get(ClassWithoutAutoBind) // works
```

When creating new templates I don't want people to forget to provide dependencies, or inject handler outside of template.

platforms/ucp/templates/handlers/dom-cleanup-handler.ts
Bielik20: Yeah, I had mixed fillings as well, but at the end of the day I feel like this is better. There is nothing more confusing than class that injects manipulator only to call restore. If you are not hard into all the helpers chances are you are not sure whether to call it or not. And should someone forget, it would be quite visible in terms of UI behaviour.

package.json
Bielik20: Actually I used `asserts` in DI library so now I need TS 3.8 support 😂

platforms/ucp/templates/bfaa-template.ts
Bielik20: Added alternate solution, tell me what do you think?

src/ad-engine/rxjs/start-and-respond-to.ts
```diff
@@ -0,0 +1,7 @@
+import { Observable, of } from 'rxjs';
+import { respondTo } from './respond-to';
+import { RxJsOperator } from './rxjs-operator';
+
+export function startAndRespondTo<T>(...observables: Observable<any>[]): RxJsOperator<T, T> {
+	return respondTo(...observables, of({}));
```
jacekbj: This operator mixes the type of the observables it receives with `object`, right? This breaks TS checks as you have to guard agains an empty object.

Bielik20: No, type is:
```ts
RxJsOperator<T, T>
```
Where T is type of original stream. Streams that you provide as arguments are meaningless in therms of their types.

platforms/ucp/templates/handlers/resolved-handler.ts
Bielik20: I would say it is more ambiguous than calling `this.unsubscribe$.next();`. You would have to travel to another class to inspect what is happening. Had it been more than one line, I would be more eager to follow this path, but as of right now I fail to see benefits.

jacekbj: It's two lines :D 
https://github.com/Wikia/ad-engine/pull/715#discussion_r399065862
I'm not going to push it any further

PR #723, opened 2020-03-25
[ADEN-10081] f2 platform with basic context

Loczi94:


platforms/f2/ad-context.ts
```diff
@@ -0,0 +1,56 @@
+import { utils } from '@wikia/ad-engine';
+
+const serverPrefix = utils.geoService.isProperCountry(['AU', 'NZ']) ? 'vm' : 'wka';
```
fraszczakszymon, line 3: Correct me if I'm wrong, but `context.get('geo.country')` is not set at this point -> `vm` won't work.

https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/utils/geo.ts#L97

platforms/f2/ad-context.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const basicContext = {
+	adUnitId:
+		`/{networkId}/${serverPrefix}1b.{slotConfig.group}/{slotConfig.slotName}` +
+		'/{state.deviceType}/ns-{custom.adLayout}/_fandom-all',
+	custom: {
+		dfpid: '5441',
```
fraszczakszymon, line 10: Where is `custom.dfpid` used?

platforms/f2/ad-context.ts
```diff
@@ -0,0 +1,56 @@
+import { utils } from '@wikia/ad-engine';
+
+const serverPrefix = utils.geoService.isProperCountry(['AU', 'NZ']) ? 'vm' : 'wka';
+
+export const basicContext = {
+	adUnitId:
+		`/{networkId}/${serverPrefix}1b.{slotConfig.group}/{slotConfig.slotName}` +
```
fraszczakszymon, line 7: `networkId` is not defined

platforms/f2/ad-context.ts
```diff
@@ -0,0 +1,56 @@
...
+		pushOnScroll: {
+			ids: [],
+			threshold: 100,
+		},
+	},
+	options: {
+		contentLanguage: 'en',
+		porvata: {
+			audio: {
+				exposeToSlot: 'true',
```
fraszczakszymon, line 22: I believe this is no longer supported

platforms/shared/setup/_base-context.setup.ts
```diff
@@ -85,6 +88,7 @@ export class BaseContextSetup {
 			this.instantConfig.get('icPermutive') && !context.get('wiki.targeting.directedAtChildren'),
 		);
 		context.set('services.nielsen.enabled', this.instantConfig.get('icNielsen'));
+		context.set('services.moatYi.enabled', this.instantConfig.get('icMoatYieldIntelligence'));
```
fraszczakszymon, line 91: MOAT Yi is already removed from ad-engine code:
https://github.com/Wikia/ad-engine/pull/734

platforms/f2/ad-context.ts
Loczi94: Good question. (: It wasn't used anywhere but after I read your comment I changed `networkId` in `adUnitId` and `vast.adUnitId` to `custom.dfpId` because this is how it's done for other platforms.

platforms/f2/ad-context.ts
Loczi94: 👍 I changed `networkId` to `custom.dfpId`.

platforms/f2/ad-context.ts
Loczi94: You mean only `exposeToSlot` or something else?

platforms/shared/setup/_base-context.setup.ts
Loczi94: Right, good catch. 👍

platforms/f2/ad-context.ts
Loczi94: Yes, of course, you're right.

platforms/f2/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { take } from 'rxjs/operators';
+import { action } from 'ts-action';
+import { ofType } from 'ts-action-operators';
+import { basicContext } from './ad-context';
+import { setupF2Ioc } from './setup-f2-ioc';
+import './styles.scss';
+
+setupPostQuecast();
+
+export const f2ready = action('[F2] Configured');
```
Bielik20, line 14: Consider creating config type, and include it like:

```suggestion
export const f2Ready = action('[F2] Configured', props<F2Config>());
```

platforms/f2/index.ts
```diff
@@ -0,0 +1,46 @@
...
+	const platformStartup = container.get(PlatformStartup);
+
+	platformStartup.configure({ isMobile: f2Payload.isMobile });
+
+	communicator.dispatch(adEngineConfigured());
+
+	platformStartup.run();
+};
+
+load();
```
Bielik20, line 46: I might have made mistake here, but this is how I would see it:

```suggestion

const communicator = new Communicator();

async function load(f2Config: F2Config) {
	context.set('src', f2Config.src);
	context.set('custom.namespace', f2Config.namespace);
	context.set('custom.adLayout', f2Config.adLayout);

	context.extend(basicContext);

	const [container]: [Container, ...any[]] = await Promise.all([
		setupF2Ioc(),
		bootstrapAndGetConsent(),
	]);
	const platformStartup = container.get(PlatformStartup);

	platformStartup.configure({ isMobile: f2Config.isMobile });

	communicator.dispatch(adEngineConfigured());

	platformStartup.run();
};

communicator.actions$.pipe(ofType(f2Ready), take(1)).subscribe(load);
```

Plus I am not a fan of that:
```ts
	context.set('src', f2Config.src);
	context.set('custom.namespace', f2Config.namespace);
	context.set('custom.adLayout', f2Config.adLayout);
```
In here. It should be somewhere in platform configure. Please change it. Perhaps in `F2BaseContextSetup`.

platforms/f2/setup-f2-ioc.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Container } from '@wikia/dependency-injection';
+import { set } from 'lodash';
+import * as fallbackInstantConfig from './fallback-config.json';
+import { F2BaseContextSetup } from './setup/f2-base-context.setup';
+
+export async function setupF2Ioc(): Promise<Container> {
+	const container = new Container();
+
+	set(window, context.get('services.instantConfig.fallbackConfigKey'), fallbackInstantConfig);
+	container.bind(InstantConfigService as any).value(await InstantConfigService.init());
```
Bielik20, line 12: I think this is no longer necessary.
```suggestion
	container.bind(InstantConfigService).value(await InstantConfigService.init());
```

platforms/shared/setup/_base-context.setup.ts
```diff
@@ -31,9 +31,12 @@ export class BaseContextSetup {
...
 	private setOptionsContext(): void {
-		context.set('options.tracking.kikimora.player', true);
-		context.set('options.tracking.slot.status', true);
-		context.set('options.tracking.slot.viewability', true);
+		context.set('options.tracking.kikimora.player', this.instantConfig.get('icPlayerTracking'));
+		context.set('options.tracking.slot.status', this.instantConfig.get('icSlotTracking'));
+		context.set(
+			'options.tracking.slot.viewability',
+			this.instantConfig.get('icViewabilityTracking'),
+		);
```
Bielik20, line 39: 💎

platforms/f2/index.ts
Loczi94: Where you think it should be exported?

platforms/f2/setup-f2-ioc.ts
Loczi94: Why you think so? What has changed?

platforms/f2/index.ts
Bielik20: Discussed on slack.

platforms/f2/setup-f2-ioc.ts
Bielik20: In the early version of DI lib, there was problem with types, that caused us to add this `as any`. It has been resolved for some time now.

platforms/f2/index.ts
```diff
@@ -0,0 +1,39 @@
...
+		bootstrapAndGetConsent(),
+	]);
+	const platformStartup = container.get(PlatformStartup);
+
+	platformStartup.configure({ isMobile: f2Payload.isMobile });
+
+	communicator.dispatch(adEngineConfigured());
+
+	platformStartup.run();
+}
```
Bielik20, line 37: You are awaiting this action twice, completely unnecessary.
Hint: you know you can commit suggestions from GitHub? ;)

```suggestion
async function load(f2config: F2Config): Promise<any> {
	context.extend(basicContext);

	const [container]: [Container, ...any[]] = await Promise.all([
		setupF2Ioc(f2config),
		bootstrapAndGetConsent(),
	]);
	const platformStartup = container.get(PlatformStartup);

	platformStartup.configure({ isMobile: f2config.isMobile });
	communicator.dispatch(adEngineConfigured());
	platformStartup.run();
}
```

platforms/f2/setup-f2-ioc.ts
```diff
@@ -0,0 +1,17 @@
+import { BaseContextSetup } from '@platforms/shared';
+import { context, InstantConfigService } from '@wikia/ad-engine';
+import { Container } from '@wikia/dependency-injection';
+import { set } from 'lodash';
+import * as fallbackInstantConfig from './fallback-config.json';
+import { F2_CONFIG, F2BaseContextSetup } from './setup/f2-base-context.setup';
+
+export async function setupF2Ioc(f2Config): Promise<Container> {
```
Bielik20, line 8: Missing type

platforms/f2/setup/f2-base-context.setup.ts
```diff
@@ -0,0 +1,25 @@
+import { BaseContextSetup } from '@platforms/shared';
+import { context, InstantConfigService, utils } from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+import { F2Config } from '../setup-f2';
+
+export const F2_CONFIG = Symbol('f2 config');
```
Bielik20, line 6: I would keep that in `platforms/f2/setup-f2.ts`, because it is not directly connected to this class. You may need to use that config in different places in the future. You don't want it to be bound to `F2BaseContextSetup` so tightly.

PR #736, opened 2020-04-06
ADEN-10107 Fix FV rv tracking

Bielik20:


src/ad-products/video/jwplayer/streams/jwplayer-stream-state.ts
```diff
@@ -18,6 +18,7 @@ export interface JwpState extends VideoDepth {
 interface VideoDepth {
 	depth: number;
 	correlator: number;
+	rv: number;
```
Bielik20, line 21: I am not sure if this is a right place for that.

src/ad-products/video/jwplayer/streams/jwplayer-stream-state.ts
```diff
@@ -87,18 +88,25 @@ export function createJwpStateStream(
 }
 
 function scanCorrelatorDepth<T>(): RxJsOperator<T, VideoDepth> {
+	const capping = context.get('options.video.adsOnNextVideoFrequency');
```
Bielik20, line 91: I don't like we start using context in `jwplayer-stream-state`.

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -62,8 +62,8 @@ export class JWPlayerHelper {
 		this.adSlot.creativeSize = vastParams.size;
 	}
 
-	setSlotElementAttributes(vastParams: VastParams): void {
-		const attributes = vastDebugger.getVastAttributesFromVastParams('success', vastParams);
+	setSlotElementAttributes(status: 'success' | 'error', vastParams: VastParams): void {
+		const attributes = vastDebugger.getVastAttributesFromVastParams(status, vastParams);
```
Bielik20, line 66: Sneaky fix.

src/ad-products/video/jwplayer/streams/jwplayer-stream-state.ts
nandy-andy: I can't think of anything better 🤷‍♂️Maybe if we keep something information of the ad slot itself (`FEATURED_VIDEO`) 🤔But this will complicate the code changes, won't it?

src/ad-products/video/jwplayer/streams/jwplayer-stream-state.ts
nandy-andy: Are there any other places which we can use here instead? Like extending `JWPlayerConfig` with it. `JWPlayerConfig` has already something like `WikiaSettings` and `options.video.adsOnNextVideoFrequency` wouldn't change with every video -- it's configuration taken from `WikiConfig`/`WikiFactory`.

Bielik20: Hmm, I don't know. It is taken from icbm. It was done in handler before that, but since it is used more than once I thought I would just include it in state. It is just if we decide to go full DI, it may be inconvenient to inject context in here. But it is very distant scenario...

PR #739, opened 2020-04-08
ADEN-10096 Create external and a9 loggers

fraszczakszymon:


src/ad-products/video/jwplayer/jwplayer-a9-logger.ts
```diff
@@ -0,0 +1,34 @@
...
+						errorCode: payload.code,
+						vast,
+						vastUrl,
+					});
+				});
+		}
+	}
+}
+
+export const jwplayerA9Logger = new JWPlayerA9Logger();
```
Bielik20, line 34: I feel like it doesn't need to be exported like that. It would be instantiated in jwplayer-manager.

PR #745, opened 2020-04-15
ADEN-10122 Adding disabled recirculation experiment

dmnsobczak:


examples/styles.scss
```diff
@@ -177,6 +183,66 @@ pre {
 	position: static;
 }
 
+.recirculation-rail {
+	h2 {
+		text-transform: capitalize;
```
nandy-andy, line 188: Consider ordering properties alphabetically -- it used to be a guideline in our company. I'm not sure if it still is 🤔

src/ad-products/utils/fmr-rotator.ts
```diff
@@ -49,17 +54,22 @@ export class FmrRotator {
...
+					if (!onlyFirstSlot) {
 						this.slotStatusChanged(AdSlot.STATUS_COLLAPSE);
 						this.scheduleNextSlotPush();
 					}
 				});
 			}
 		});
 
-		setTimeout(() => {
+		if (context.get('options.floatingMedrecRecirculationDisabled')) {
```
nandy-andy, line 65: I'm not sure how important it is but since it's the 2nd time you're calling `context.get()` for this option maybe it's worth assigning it to a variable and reusing the variable here 🤔

src/ad-products/utils/fmr-rotator.ts
```diff
@@ -33,13 +33,18 @@ export class FmrRotator {
 		this.recirculationElement = document.getElementById('recirculation-rail');
 		this.refreshInfo.startPosition =
 			utils.getTopOffset(this.recirculationElement) - this.navbarManager.getHeight();
-		this.btRecStatus = this.btRec.isEnabled();
+		this.btRecStatus = this.btRec && this.btRec.isEnabled();
```
jacekbj, line 36:
```suggestion
		this.btRecStatus = this.btRec?.isEnabled();
```

nandy-andy: 💎 fancy "new" features 👍

dmnsobczak: Oh YES! I forgot we can finally use it 💎

src/ad-products/utils/fmr-rotator.ts
dmnsobczak: I don't want to create the function-global variable. I think two is not enough for that, but we can debate :D

examples/styles.scss
dmnsobczak: Me likey! Reordered.

PR #746, opened 2020-04-15
[ADEN-10133] IAB ads on UCP Gamepedia desktop

Loczi94:


platforms/ucp-hydra/setup/dynamic-slots/hydra-dynamic-slots-setup.service.ts
```diff
@@ -0,0 +1,112 @@
...
+			this.appendRotatingSlot(
+				icbSlotName,
+				slotConfig.repeat.slotNamePattern,
+				document.querySelector(slotConfig.parentContainerSelector),
+			);
+		});
+	}
+
+	private configureIncontentPlayerFiller(): void {
+		const icpSlotName = 'incontent_player';
```
preludek, line 56: 😢

Loczi94: Yo miss Porvata? :D

platforms/ucp-hydra/ad-context.ts
```diff
@@ -0,0 +1,112 @@
+export const basicContext = {
+	adUnitId:
+		'/{custom.dfpId}/{custom.serverPrefix}.{slotConfig.group}/{slotConfig.adProduct}/{state.deviceType}/' +
+		'{targeting.skin}-{targeting.s2}/_gp_wiki-gamepedia',
+	custom: {
+		dfpId: '5441',
+		serverPrefix: 'wka1b',
+		rolloutTracking: 'ucp',
```
nandy-andy, line 8: that's interesting we've got it in two places 🤔

platforms/ucp-hydra/modes/hydra-ads-mode.service.ts
```diff
@@ -0,0 +1,43 @@
...
+		const inhibitors: Promise<any>[] = [];
+		const targeting = context.get('targeting');
+
+		inhibitors.push(bidders.requestBids());
+		inhibitors.push(wadRunner.call());
+
+		permutive.call();
+		confiant.call();
+		durationMedia.call();
+		nielsen.call({
```
nandy-andy, line 30: Do we have Nielsen on Hydra? I assume the Nielsen service's `call()` won't work here because it's disabled in ICBM but that's a good question about all those trackings on Fandom and Gamepedia pages and its state after moving to UCP 🤔

PR #747, opened 2020-04-16
ADEN-10152 UCP Skin Template

Bielik20:


platforms/ucp/styles.scss
```diff
@@ -117,6 +117,16 @@
...
+// #### ROADBLOCK STYLES ####
+
+#hivi_leaderboard,
+#top_leaderboard,
+.WikiaPage,
+.wds-global-footer {
+	position: relative;
+	z-index: 1;
+}
+
```
Bielik20, line 129: It is without any special selector so that it works always. It is slightly easier to catch regression if someone is not using roadblock often.

platforms/ucp/templates/handlers/skin/skin-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+		this.params.pixels = this.params.pixels || [];
+		this.params.backgroundColor = this.removeColorHash(this.params.backgroundColor);
+		this.params.middleColor = this.removeColorHash(
+			this.params.middleColor ?? this.params.backgroundColor,
+		);
+	}
+
+	private removeColorHash(color: string = ''): string {
+		return color.replace('#', '');
+	}
```
Bielik20, line 26: Do we need that? I wrote it because there was similar code on oasis. But it is weird we check if someone configured it right. We should expect correct configuration in order for the feature to work correctly.

PR #749, opened 2020-04-16
ADEN-10050 | Add IAS PO integration

bckatarzyna:


src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
+import { context, utils } from '@ad-engine/core';
+
+const logGroup = 'ias-publisher-optimization';
+const scriptUrl = '//cdn.adsafeprotected.com/iasPET.1.js';
+const SLOT_LIST = [
```
fraszczakszymon, line 5: It should not be defined in the service. What about context `services. iasPublisherOptimization.slots.*`?

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+interface IasPetQueueElement {
+	adSlots: number[][];
+	dataHandler: (string) => void;
+}
+
+interface IasTargetingData {
+	fr?: string;
+	slots?: any[];
+}
```
fraszczakszymon, line 26:
```suggestion
interface IasTargetingSlotData {
  id?: string;
  vm?: string[];
}

interface IasTargetingData {
	fr?: string;
	slots?: IasTargetingSlotData[];
}
```

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+				.then(() => this.setup());
+			this.isLoaded = true;
+		}
+	}
+
+	private setup(): void {
+		// @ts-ignore
+		this.iasPET = __iasPET || {};
+		this.iasPET.queue = this.iasPET.queue || [];
+		this.iasPET.pubId = '930616';
```
fraszczakszymon, line 55: What about moving it to context? (i.e. `services.iasPublisherOptimization.publisherId`)

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+				adSlots: iasPETSlots,
+				dataHandler: this.iasDataHandler,
+			});
+		});
+	}
+
+	private setInitialTargeting(): void {
+		context.set('targeting.fr', -1);
+		SLOT_LIST.forEach((slotName) => {
+			context.set(`slots.${slotName}.targeting.vw`, -1);
```
fraszczakszymon, line 87: I'm not sure if we don't need it as a string?
```suggestion
		context.set('targeting.fr', '-1');
		SLOT_LIST.forEach((slotName) => {
			context.set(`slots.${slotName}.targeting.vw`, '-1');
```

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+		});
+	}
+
+	private iasDataHandler(adSlotData: string): void {
+		const iasTargetingData: IasTargetingData = JSON.parse(adSlotData);
+		context.set('targeting.fr', iasTargetingData.fr);
```
fraszczakszymon, line 93: What about mapping it to 0/1?
```suggestion
		context.set('targeting.fr', iasTargetingData.fr ? '1' : '0');
```

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	private setup(): void {
+		// @ts-ignore
+		this.iasPET = __iasPET || {};
+		this.iasPET.queue = this.iasPET.queue || [];
+		this.iasPET.pubId = '930616';
+		this.setInitialTargeting();
+
+		window.googletag.cmd.push(() => {
```
fraszczakszymon, line 58: I believe we could export `postponeExecutionUntilGptLoads` from `gpt-provider` and reuse it here

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+			const slots = [];
+			const iasPETSlots = [];
+			SLOT_LIST.forEach((slotName) => {
+				slots.push(context.get(`slots.${slotName}`));
+			});
+
+			slots.forEach((slot) => {
```
fraszczakszymon, line 65: It seems that doubled iteration is not needed here. What about iterating over SLOT_LIST once and taking slot definition in the loop?

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -0,0 +1,101 @@
...
+				iasPETSlots.push({
+					adSlotId: slot['adProduct'],
+					size: sizes,
+					adUnitPath,
+				});
+			});
+
+			this.iasPET.queue.push({
+				adSlots: iasPETSlots,
+				dataHandler: this.iasDataHandler,
```
fraszczakszymon, line 79:
```suggestion
				dataHandler: () => this.iasDataHandler(),
```
please, so we're able to use `this` in the function in future

PR #754, opened 2020-04-22
ADEN-10162 Unify UCP

Bielik20:


platforms/shared/context/base/ucp-base-context.setup.ts
```diff
@@ -1,6 +1,6 @@
-import { BaseContextSetup } from '@platforms/shared';
 import { context, Dictionary } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
+import { BaseContextSetup } from '../../setup/_base-context.setup';
```
nandy-andy, line 3: This looks suspicious -- the relative path, is it correct?

platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -1,9 +1,21 @@
-import { getDomain, TargetingSetup } from '@platforms/shared';
-import { context, Targeting, utils } from '@wikia/ad-engine';
-import { Injectable } from '@wikia/dependency-injection';
+import { Binder, context, Targeting, utils } from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+import { TargetingSetup } from '../../setup/_targeting.setup';
+import { getDomain } from '../../utils/get-domain';
+
+const SKIN = Symbol('targeting skin');
```
nandy-andy, line 6: 🤔

platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -70,46 +82,51 @@ export class UcpTargetingSetup implements TargetingSetup {
...
+			return {
+				isDedicatedForArticle,
+				hasVideoOnPage: isDedicatedForArticle || bridgeVideoPlayed,
+			};
+		}
+
+		return {};
 	}
 
+	// TODO: This should not be here. It is a side effect that is unpredictable.
```
nandy-andy, line 129: Where do you think it should be? It seems to being used only in the `getAdLayout()` and the two lines can be moved there 🤔Although, the issue is wider as I understand since with each `getAdLayout()` we re-set the context.

platforms/shared/index.ts
```diff
@@ -27,10 +26,14 @@ export * from './templates/sports-templates.setup';
 export * from './setup/_tracking.setup';
 export * from './tracking/common-tracking.setup';
 export * from './modes/no-ads/_no-ads.mode';
+export * from './modes/no-ads/ucp-no-ads.mode';
```
nandy-andy, line 29: What is the naming convention we should stick to? Some files are prefixed with an underscore and some not 🤔

platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -70,46 +82,51 @@ export class UcpTargetingSetup implements TargetingSetup {
...
 
 		this.updateVideoContext(hasFeaturedVideo, hasIncontentPlayer);
 
-		return layout;
+		return adLayout;
+	}
+
+	private getPageType(targeting: MediaWikiAdsTargeting): string {
+		return targeting.pageType || 'article';
+	}
```
Loczi94, line 110: Why you moved it to separate function? It's only one line of code.

platforms/shared/context/targeting/ucp-targeting.setup.ts
Loczi94: Why do you think that leaving it here can have unpredictable side effects?

nandy-andy: I don't think @Bielik20 meant leaving it here can have unpredictable side effects. I think it can have unpredictable side effects no matter if it's in this private methods or if we put these two lines within `getAdLayout()`. I'm guessing it's because every time we're calling `getAdLayout()` we may change the context which is the unpredictable side effect 🤔But we need to wait for Bielik to present his point of view :)

platforms/shared/context/base/ucp-base-context.setup.ts
Bielik20: Yes because they are both in shared.

platforms/shared/context/targeting/ucp-targeting.setup.ts
Bielik20: Because I feel like it is conceptually separate. In GP for example there is method for getting PageType, and this seems to be its equivalent.

platforms/shared/index.ts
Bielik20: Idea is that blueprint classes have prefixes. The ones used in platform startup.

platforms/shared/context/targeting/ucp-targeting.setup.ts
Bielik20: Side effect is that invoking `getAdLayout` method sets something on the context. I would have never thought that invoking method prefixed with `get` would introduce some changes to the broadly understood state.

platforms/shared/context/targeting/ucp-targeting.setup.ts
Bielik20: With DI we can inject symbols like
```ts
constructor(@Inject(SKIN) private skin: string) {}
```

For the DI container to be able to deliver `SKIN`, one needs to provide it.
We can provide skin by calling `value` method on the binding:
```ts
container.bind(SKIN).value('oasis');
```

Or by delivering binder object:
```ts
container.bind({ bind: SKIN, value: 'oasis' });
```

In our exemple we are choosing second option:
```ts
container.bind(TargetingSetup).to(UcpTargetingSetup);
container.bind(UcpTargetingSetup.skin('oasis'));
```

First we bind `TargetingSetup` to `UcpTargetingSetup`.
Second we bind `SKIN` to value `oasis`.

Now in `UcpTargetingSetup` class we create static method that return binder object for us:
```ts
	static skin(skin: string): Binder {
		return {
			bind: SKIN,
			value: skin,
		};
	}
```

Because of that, we don't even need to export SKIN symbol, because we export binder that includes it.

PR #764, opened 2020-04-28
ADEN-10169 UCP-GP UAP

Bielik20:


platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
```diff
@@ -18,6 +18,13 @@ export class BfaaBootstrapHandler implements TemplateStateHandler {
 	) {}
 
 	async onEnter(transition: TemplateTransition<'sticky' | 'impact'>): Promise<void> {
+		this.adSlot.getElement().style.setProperty('backgroundColor', '#000');
```
fraszczakszymon, line 21: Do we need to set it in JS?

platforms/shared/templates/handlers/video/video-learn-more-handler.ts
```diff
@@ -0,0 +1,37 @@
...
+				takeUntil(this.unsubscribe$),
+			)
+			.subscribe();
+	}
+
+	private setLearnMoreBlockDisplay(player: Porvata4Player): void {
+		const playerContainer: HTMLElement = player.settings.getPlayerContainer();
+		const learnMore: HTMLElement = playerContainer.querySelector('.learn-more');
+
+		this.manipulator.element(learnMore).setProperty('display', 'block');
```
fraszczakszymon, line 31: Imo, we should avoid setting inline styles. What about adding/removing `.hide` class?

platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -9,20 +9,21 @@ export class UapDomManager {
...
-	setBodyOffsetResolved(): void {
-		this.setBodyOffset(this.reader.getBodyOffsetResolved());
+	setPageOffsetResolved(): void {
+		this.setPageOffset(this.reader.getPageOffsetResolved());
 	}
 
-	private setBodyOffset(value: number): void {
-		this.manipulator.element(document.body).setProperty('paddingTop', `${value}px`);
+	private setPageOffset(value: number): void {
+		this.manipulator.element(this.page).setProperty('marginTop', `${value}px`);
```
fraszczakszymon, line 26: Is it fine for desktop bfaa on ucp? (margin vs padding)

platforms/ucp-hydra/styles.scss
```diff
@@ -1,8 +1,51 @@
...
+@include expanded-slot();
+
+// #######################################
+
 .gpt-ad {
 	&#top_leaderboard,
 	&#bottom_leaderboard {
 		text-align: center;
-		padding: 5px 0;
+		margin: 5px 0;
```
fraszczakszymon, line 48: It will create additional empty space, when there is no ad (collapse response). Am I wrong?

platforms/ucp-minerva/setup/context/slots/minerva-slots-context-setup.ts
```diff
@@ -109,6 +109,22 @@ export class MinervaSlotsContextSetup implements SlotsContextSetup {
...
+				options: {},
+				slotShortcut: 'b',
+				sizes: [],
+				disabled: true,
+				defaultSizes: [[2, 2]],
+				insertAfterSelector: '#bodyContent',
+				targeting: {
+					loc: 'footer',
+					rv: 1,
+					xna: 1,
```
fraszczakszymon, line 125: Is this needed?

platforms/ucp/styles.scss
```diff
@@ -1,14 +1,13 @@
...
-@import './styles/progress-bar';
-@import './styles/porvata-icons';
+@import 'platforms/shared/styles/index';
+
+.wds-global-navigation-wrapper {
+	@include uap-navbar();
+}
+
+#globalNavigation {
+	position: static;
```
fraszczakszymon, line 9: I have mixed feelings about setting so general styles that impact the platform. In my opinion, they should be applied on the platform side or when it's necessary (like UAP scenarios). But for now, it turns out that loaded ad-engine changes global navigation styles, which may conflict with platform styles/behavior 🤷‍♂️

platforms/shared/templates/helpers/uap-dom-manager.ts
Bielik20: Yes, I left padding initially because of legacy reasons, but both work fine.

platforms/ucp-hydra/styles.scss
Bielik20: Not sure.

platforms/ucp-minerva/setup/context/slots/minerva-slots-context-setup.ts
Bielik20: Don't know. I copied it from ucp-oasis.

platforms/ucp/styles.scss
Bielik20: This is not a change, I just moved it from another place.

I actually confronted ... someone, I don't remember who. And according to that person changing those styles on the platform side is very problematic.

I intentionally change it even if there is no UAP, so that it is easier to catch it should it conflict with something.

platforms/shared/templates/handlers/video/video-learn-more-handler.ts
Bielik20: I am actually more in favour of using inline styles as we use them for almost everything in UAP. Also, not very important, but `DomManipulator` is smarter with inline styles than with classes.

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
```diff
@@ -29,11 +36,20 @@ export class BfaaBootstrapHandler implements TemplateStateHandler {
 		}
 	}
 
+	private ensureImage(): void {
+		if (!(this.params.image2 && this.params.image2.background)) {
```
dmnsobczak, line 40: `if (!this.params.image2 || !this.params.image2.background) {`?

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
```diff
@@ -29,11 +36,20 @@ export class BfaaBootstrapHandler implements TemplateStateHandler {
 		}
 	}
 
+	private ensureImage(): void {
```
dmnsobczak, line 39: Maybe I'm missing some context but why `ensure`, not eg. `provideImage` or `setImages`?

platforms/shared/templates/handlers/video/video-learn-more-handler.ts
dmnsobczak: What you mean `smarter`?

I'm also giving my vote to classes. It's easier to read what code does, especially if classes are called like `hideWhenActive` rather than bunch of different inline styles.

platforms/shared/templates/helpers/video-dom-manager.ts
```diff
@@ -35,19 +36,30 @@ export class VideoDomManager {
...
+		this.setProperties(thumbnail, props);
+	}
+
+	private setProperties(
+		element: HTMLElement,
+		{ width, height, top, right, bottom }: UapVideoSize,
+	): void {
+		this.manipulator.element(element).setProperty('width', `${width}px`);
+		this.manipulator.element(element).setProperty('height', `${height}px`);
+		if (!isUndefined(top)) {
```
dmnsobczak, line 55: I would suggest `isNumeric` instead but are we 100% sure that this is gonna be number, not string?

platforms/shared/templates/helpers/video-dom-reader.ts
```diff
@@ -0,0 +1,64 @@
...
+		const { height, width } = this.getSize(slotHeight, progress);
+		const top = this.getPercentage(progress, this.params.config.state.top);
+		const right = this.getPercentage(progress, this.params.config.state.right);
+		const bottom = this.getPercentage(progress, this.params.config.state.bottom);
+
+		return {
+			top,
+			right,
+			bottom,
+			height: Math.ceil(height),
```
dmnsobczak, line 42: Please add to QA notes a sanity check if using ceils won't reintroduce "one blank pixel" regressions.

dmnsobczak: Also: can't it be ceiled inside `getSize` method?

platforms/ucp-minerva/modes/minerva-ads-mode.ts
```diff
@@ -43,6 +43,7 @@ export class MinervaAdsMode implements AdsMode {
 	private setAdStack(): void {
 		context.push('state.adStack', { id: 'top_leaderboard' });
 		context.push('state.adStack', { id: 'top_boxad' });
+		context.push('events.pushOnScroll.ids', 'bottom_leaderboard');
```
dmnsobczak, line 46: Order ಠ_ಠ

platforms/ucp-minerva/setup/context/slots/minerva-slots-context-setup.ts
dmnsobczak: I think it is, but maybe we're missing this logic on platforms?
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/slots.js#L340

platforms/ucp/styles.scss
dmnsobczak: I'm not sure what's the reason for that change, but can't we at least move it inside `.has-uap #globalNavigation` body class? Do we need `position: static;` always?

platforms/ucp/styles.scss
```diff
@@ -17,77 +16,33 @@
...
-				position: static;
-				text-align: center;
-			}
-		}
 	}
 }
 
+@include expanded-slot();
+
+// #######################################
```
dmnsobczak, line 44: Can we name those separators and what they're separating?

platforms/ucp/styles.scss
Bielik20: I feel like it is safer to have it always. Should change occur, I want someone to notice it immediately rather than at some later date.

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
Bielik20: It's been written some time ago by myself and Jacek. I don't remember why it is like that, but why change now?

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
Bielik20: Because there are two images, and you ensure that at least one is visible.

platforms/shared/templates/helpers/video-dom-manager.ts
Bielik20: Its type is `number | undefined`. So there is really no need to check if it is a number.

platforms/shared/templates/helpers/video-dom-reader.ts
Bielik20: > Also: can't it be ceiled inside getSize method?

There were reasons. I had it both ways, and end up using this. Don't remember why.

> Please add to QA notes a sanity check if using ceils won't reintroduce "one blank pixel" regressions.

There is always problem with blank pixel, but my testing showed I had least problem with ceil :/ Still it will happen on some sizes.
We (meaning me and Szymon) decided that if there is a problem that bothers someone we will fix it later.

platforms/ucp/styles.scss
Bielik20: I guess. They separate UAP styles from some special site styles.

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
dmnsobczak: No reason. I'm just a fan of De Morgan's law :D

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
dmnsobczak: Got it!

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
Bielik20: Done

platforms/shared/templates/helpers/video-dom-manager.ts
Bielik20: Done.

platforms/ucp-minerva/modes/minerva-ads-mode.ts
Bielik20: It is the same order as in `ucp` and `ucp-hydra`. I don't know what you want me to do :P

platforms/ucp/styles.scss
Bielik20: Moved into a mixin, aka function.

platforms/ucp-minerva/modes/minerva-ads-mode.ts
dmnsobczak: Push order doesn't matter, I'm talking about alphabetical :D Nevermind, I'm just complaining :P

platforms/shared/templates/handlers/video/video-learn-more-handler.ts
Bielik20: Replaced with `show-learn-more` css class.

PR #765, opened 2020-04-29
ADEN-10166 | Add incontent_boxad to ucp-minerva.

jacekbj:
Slot is injected before fifth header of size h1..h6.

Signed-off-by: Jacek Bronski-Jankowski <jjankowski@fandom.com>


platforms/ucp-minerva/setup/context/slots/minerva-slots-context-setup.ts
```diff
@@ -68,14 +68,33 @@ export class MinervaSlotsContextSetup implements SlotsContextSetup {
 					[300, 600],
 				],
 				bidderAlias: '02_MR',
-				insertBeforeSelector: '#mw-content-text h2:nth-child(2)',
+				// before second header of size h1..h6
```
nandy-andy, line 71: Am I reading the code below correctly and it just goes from h1 to h2 which is not what this comment claims?

jacekbj: It checks if there are two h1 headers, then h2 and so on up to h6.
I copied this directly from GP site.

jacekbj: Good point! Fixed.

PR #767, opened 2020-04-30
[ADEN-10164] ucp-hydra: top boxad

Loczi94:


platforms/ucp-hydra/setup/dynamic-slots/hydra-dynamic-slots-setup.ts
```diff
@@ -43,6 +44,16 @@ export class HydraDynamicSlotsSetup implements DynamicSlotsSetup {
 		}
 	}
 
+	private injectTB(): void {
+		const siderail = document.getElementById('siderail_ucpinternalgptestproject43');
```
fraszczakszymon, line 48: What about
```suggestion
		const dbName = context.get('wiki.adsContext.targeting.wikiDbName');
		const siderail = document.getElementById(`siderail_${dbName}`);
```

PR #777, opened 2020-05-07
ADEN-10239 Adding Labrador keyvals to GAM queries on platforms

dmnsobczak:


platforms/shared/context/base/ucp-base-context.setup.ts
```diff
@@ -34,5 +34,7 @@ export class UcpBaseContextSetup extends BaseContextSetup {
 				);
 			}
 		}
+
+		super.setLabradorContext();
```
jacekbj, line 38: If it has to be at the end then please add a comment.

jacekbj: Isn't it already called in `super.configureBaseContext` (line 8)?

dmnsobczak: Discussed on Slack. Refactored.

PR #780, opened 2020-05-08
ADEN-9797 Add option in MessageBus to filter messages by origin.

jacekbj:
Require messages intercepted by PostMessageTracker in platforms/shared/tracking to be from the same origin or from https://tpc.googlesyndication.com.


spec/ad-engine/services/message-bus.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+		});
+
+		it('should not execute callback when event origin does not match origin in match object', () => {
+			messageBus.register({ origin: [mockOrigin] }, cbStub);
+			mockWindow.dispatchMessage({ data: { AdEngine: {} }, origin: 'https://foo.com' });
+
+			expect(cbStub.callCount).to.equal(0);
+		});
+
+		it.skip('should execute callback when payload includes all keys specified in match object', () => {
```
dmnsobczak, line 96: Is this some kind of placeholder?

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -18,6 +18,7 @@ import { Provider } from './provider';
 const logGroup = 'gpt-provider';
 
 export const ADX = 'AdX';
+export const GAMOrigin = 'https://tpc.googlesyndication.com';
```
dmnsobczak, line 21: Google is also using `googleads.g.doubleclick.net` domain. I don't know if they fully transitioned to new domain - I can't find any clear news about that. I found an article which list both sources in content:
https://spaceads.digital/how-to-fix-googleads-g-doubleclick-net-sources-and-tpc-googlesyndication-com-in-google-analytics/
And a bit older Google reply:
https://support.google.com/google-ads/thread/4757309?hl=en
Maybe it's safe to add both values here?

jacekbj: Very good catch. I added the `https://googleads.g.doubleclick.net` to GAMOrigins array.

spec/ad-engine/services/message-bus.spec.ts
jacekbj: I left it to indicate that it's not fully covered. I did not have enough time to provide full coverage.

PR #781, opened 2020-05-11
ADEN-10178 Add IHI Products

Bielik20:


platforms/shared/templates/handlers/slot/slot-decision-on-viewability-handler.ts
```diff
@@ -0,0 +1,37 @@
+import { AdSlot, TEMPLATE, TemplateStateHandler, TemplateTransition } from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+import { EMPTY, from, merge, Observable, Subject, timer } from 'rxjs';
+import { mergeMap, takeUntil, tap } from 'rxjs/operators';
+
+@Injectable({ autobind: false })
+export class SlotDecisionOnViewabilityHandler implements TemplateStateHandler {
+	private unsubscribe$ = new Subject<void>();
+	private additionalHideTime?: number;
+	private timeoutHideTime?: number;
```
Bielik20, line 10: It used to be delivered in the config of old HideOnViewability template.
It could be injected in templates 2.0, that's why I left it here.
I would like to hear a feedback do we wan't to implement method for injecting those params, or we want to get rid of them.

Bielik20: Discussed on slack. It is delivered though ICBM.

platforms/shared/templates/handlers/slot/slot-decision-on-viewability-handler.ts
```diff
@@ -0,0 +1,35 @@
...
+			.pipe(
+				take(1),
+				tap(() => transition('transition')),
+				takeUntil(this.unsubscribe$),
+			)
+			.subscribe();
+	}
+
+	private getViewabilityStream(): Observable<unknown> {
+		return from(this.adSlot.viewed).pipe(mergeMap(() => timer(this.additionalHideTime)));
```
jacekbj, line 34: If you don't care about value returned by `this.adSlot.viewed` then you can use `merge` operator.

jacekbj: Or would it fail because the timer would be created too fast?

Bielik20: Yeah, it would be too fast.
We want `this.adSlot.viewed` to complete, and then start timer.
`mergeMap` is used to flatten observable.

PR #782, opened 2020-05-12
ADEN-10241 A9 video bids refreshing

fraszczakszymon:


src/ad-bidders/a9/index.ts
```diff
@@ -29,14 +29,14 @@ const logGroup = 'A9Provider';
 
 export class A9Provider extends BidderProvider {
 	static A9_CLASS = 'a9-ad';
+	private static isApstagConfigured = false;
```
Bielik20, line 32: Why this is static?
We do not create multiple instances of A9Provider, do we?

fraszczakszymon: Discussed on slack. We're creating multiple instances.

PR #784, opened 2020-05-13
ADEN-10224 | Add Facebook pixel integration

bckatarzyna:


src/ad-services/facebook-pixel/index.ts
```diff
@@ -0,0 +1,33 @@
+import { context, utils } from '@ad-engine/core';
+import { insertFacebookPixel } from './facebook-pixel-script';
+
+const logGroup = 'facebook-pixel';
+
+class FacebookPixel {
+	isSetUp = false;
```
fraszczakszymon, line 7: Where is this used?

bckatarzyna: Usage added

PR #786, opened 2020-05-14
ADEN-10257 Add IAS Brand Safety key-vals

fraszczakszymon:


src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -3,13 +3,27 @@ import { decorate } from 'core-decorators';
...
+
+interface BrandSafetyData {
+	adt?: BrandSafetyValue;
+	alc?: BrandSafetyValue;
+	dlm?: BrandSafetyValue;
+	drg?: BrandSafetyValue;
+	hat?: BrandSafetyValue;
+	off?: BrandSafetyValue;
+	vio?: BrandSafetyValue;
+}
```
Bielik20, line 18: Just a suggestion it could be handled like so:

```suggestion
const brandSafetyKeys = ['adt', 'alc', 'dlm', 'drg', 'hat', 'off', 'vio'] as const;

type BrandSafetyValue = 'veryLow' | 'low' | 'medium' | 'high';
type BrandSafetyKey = typeof brandSafetyKeys[number]
type BrandSafetyData = Partial<Collection<BrandSafetyKey, BrandSafetyValue>>
```

Although I would lie saying that I had no problems with that `as const` types in the past. They work well until you want to use fraction of that type. It is your call, in this case it would be fine.

fraszczakszymon: Looks nice 💎

PR #789, opened 2020-05-14
[ADEN-10259] Logo Replacement Template

Loczi94:


src/ad-products/templates/uap/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './big-fancy-ad-above';
 export * from './big-fancy-ad-below';
 export * from './big-fancy-ad-in-player';
+export * from './logo-replacement';
```
fraszczakszymon, line 4: The template itself isn't really connected to UAP. What do you think about putting it one level higher --  in the `templates` directory?

src/ad-products/templates/uap/logo-replacement.ts
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	init(params: LogoReplacementTemplateConfig): void {
+		utils.logger(LogoReplacement.getName(), 'init');
+
+		this.replaceLogo(params);
+	}
+
+	private replaceLogo(params: LogoReplacementTemplateConfig): void {
+		const parentElement = document.querySelector('.wds-global-navigation__content-bar-left');
```
fraszczakszymon, line 24: Are those selectors valid for all platforms? What do you think about putting string selectors in the template config?

PR #794, opened 2020-05-19
ADEN-10270 Lazy load slots when UAP

Bielik20:


platforms/ucp-minerva/modes/minerva-ads-mode.ts
```diff
@@ -44,9 +47,18 @@ export class MinervaAdsMode implements AdsMode {
...
+				context.push('events.pushOnScroll.ids', 'top_boxad');
+				context.push('events.pushOnScroll.ids', 'incontent_boxad_1');
+				context.push('events.pushOnScroll.ids', 'footer_boxad');
+			} else {
+				context.push('state.adStack', { id: 'top_boxad' });
+				context.push('state.adStack', { id: 'incontent_boxad_1' });
+				context.push('state.adStack', { id: 'footer_boxad' });
+			}
+			context.push('events.pushOnScroll.ids', 'bottom_leaderboard');
+		});
```
Bielik20, line 62: I don't like that we introduce yet another place where we make decision about slots.
That is to be addressed in:
* https://wikia-inc.atlassian.net/browse/ADEN-10279
* https://wikia-inc.atlassian.net/browse/ADEN-10280

PR #803, opened 2020-05-25
[ADEN-10276] ucp: center ad in hivi leaderboard and bottom leaderboard

Loczi94:


Loczi94:
By adding it to .gpt-ad we center all ads and I'm not sure if we want all of them to be always centered. What about for example incontent player? Is it always centered? Bug fix refers to BLB and HiVi only and I'd rather stick to it.

Bielik20:
Well, this is a question I asked here: https://fandom.slack.com/archives/G03AC6KLY/p1586329152032800

PR #809, opened 2020-05-28
[ADEN-10273] LogoReplacement on all platforms

Loczi94:


platforms/gamepedia/styles.scss
```diff
@@ -57,3 +57,11 @@ html.disable-ads {
...
 	margin-top: 0 !important;
 }
+
+.new-logo {
+	max-height: 30px;
+}
+
+.tracking-pixel {
+	display: none;
+}
```
Bielik20, line 67: Consider moving it somewhere to ad-products as mixin so that platform needs only to import it and invoke.

platforms/muthead/templates/big-fancy-ad-above-config.ts
```diff
@@ -0,0 +1,92 @@
+import { slotsContext } from '@platforms/shared';
```
Bielik20, line 1: This should be in shared for sports (mut + fut).

platforms/muthead/templates/big-fancy-ad-below-config.ts
```diff
@@ -0,0 +1,16 @@
+import { slotsContext } from '@platforms/shared';
```
Bielik20, line 1: This should be in shared for sports (mut + fut).

PR #817, opened 2020-06-08
ADEN-10311 Sync BFT with SafeFanTakeover

Bielik20:


src/ad-products/templates/safe-uap/safe-big-fancy-ad-proxy.ts
```diff
@@ -14,6 +14,7 @@ export class SafeBigFancyAdProxy {
 
 	loadTemplate(): void {
 		this.createNewAdSlotIframe();
+		this.finishFirstCallIfBfaa();
```
Bielik20, line 17: Is this the right place?
Is this the right name?

PR #818, opened 2020-06-09
ADEN-10305 | Remove amazon from targeting when bid expired

bckatarzyna:


src/ad-engine/video/vast-url-builder.ts
```diff
@@ -41,6 +41,8 @@ function getCustomParameters(slot: AdSlot, extraTargeting: Dictionary = {}): str
 		...extraTargeting,
 	};
 
+	validateAmazonBid(params);
```
fraszczakszymon, line 44: I have mixed feelings about coupling `vast-url-builder` with A9/amzn parameters. What do you think about dispatching custom event here? For example `INVALIDATE_TARGETING`. Then in the A9 module, we could listen for this event and filter out expired bids.

src/ad-bidders/a9/index.ts
```diff
@@ -328,14 +345,23 @@ export class A9Provider extends BidderProvider {
...
 		this.bids[slotName] = {};
 		keys.forEach((key) => {
 			if (this.targetingKeys.indexOf(key) === -1) {
 				this.targetingKeys.push(key);
 			}
 			this.bids[slotName][key] = bidTargeting[key];
 		});
+
+		if (context.get(`slots.${slotName}.isVideo`)) {
+			this.bids[slotName]['amznExpirationDate'] = expirationDate.toString();
```
fraszczakszymon, line 363:
```suggestion
			this.bids[slotName]['amznExpirationDate'] = expirationDate.toString();
			this.targetingKeys.push('amznExpirationDate');
```

src/ad-engine/video/vast-url-builder.ts
```diff
@@ -41,6 +41,8 @@ function getCustomParameters(slot: AdSlot, extraTargeting: Dictionary = {}): str
 		...extraTargeting,
 	};
 
+	eventService.emit(events.INVALIDATE_SLOT_TARGETING, params);
```
fraszczakszymon, line 44: We're losing reference to AdSlot targeting object here. In fact, we'll remove amzn bids from params here, but they will still be available on slot level.

src/ad-bidders/a9/index.ts
```diff
@@ -90,6 +94,18 @@ export class A9Provider extends BidderProvider {
...
 
+	private invalidateSlotTargeting(params: Dictionary) {
+		const expirationDate = Date.parse(params.amznExpirationDate);
+		const currentDate = new Date().getTime();
+
+		if (expirationDate < currentDate) {
+			delete params.amznExpirationDate;
+			delete params.amznbid;
+			delete params.amzniid;
+			delete params.amznp;
```
fraszczakszymon, line 105: What do you think about:
```suggestion
			// adSlot needed
			const slotAlias = this.getSlotAlias(adSlot.getSlotName());

			delete this.bids[slotAlias];
			this.targetingKeys.forEach((key: string) => {
				context.remove(`slots.${adSlot.getSlotName()}.targeting.${key}`);
			});
```

PR #819, opened 2020-06-10
ADEN-10313 SafeFanTakeOverElement boxads.

jacekbj:


src/ad-products/templates/safe-uap/safe-fan-takeover-element.ts
```diff
@@ -37,17 +37,41 @@ export class SafeFanTakeoverElement {
...
+		const divContainer = document.createElement('div');
+		const iframeBuilder = new utils.IframeBuilder();
+		let imageUrl = SafeFanTakeoverElement.config.desktop.images.boxad300x250;
+		let height = '250px';
+		const width = '300px';
+
+		if (this.adSlot.getCreativeSize() === '300x600') {
+			imageUrl = SafeFanTakeoverElement.config.desktop.images.boxad300x600;
+			height = '600px';
+		}
```
Bielik20, line 64: My personal bias. I would close it inside some method `getBoxadProperties`, something like that:

```ts
private loadBoxad(): void {
  ...
  const { height, width, imageUrl } = this.getBoxadProperties();
  ...
}

private getBoxadProperties(): BoxadProperties {
  if (this.adSlot.getCreativeSize() === '300x600') {
    return {
      width: '300px',
      height: '600px',
      imageUrl: SafeFanTakeoverElement.config.desktop.images.boxad300x600,
    };
  }

  return {
    width: '300px',
    height: '250px',
    imageUrl: SafeFanTakeoverElement.config.desktop.images.boxad300x250,
  }
}
```

To make that "decision" moment stand out more. So that it wouldn't be missed. But that is just a suggestion.

PR #838, opened 2020-06-30
ADEN-10317 Implement WatchingThat plugin for JWPlayer

fraszczakszymon:


src/ad-products/video/jwplayer/watching-that/index.ts
```diff
@@ -0,0 +1,47 @@
...
+				window.wtAdTracer({ apiKey }, player.getContainer().id);
+			} catch (e) {
+				utils.warner(logGroup, 'Failed to initialize WatchingThat plugin', e);
+			}
+		});
+	}
+
+	isEnabled(): boolean {
+		return (
+			!!context.get('options.video.watchingThat.enabled') &&
```
dmnsobczak, line 31: I think that `!!` is unnecessary here.

dmnsobczak (approved):

src/ad-products/video/jwplayer/watching-that/index.ts
nandy-andy: Discussed offline :)

PR #843, opened 2020-07-01
ADEN-10386 Fix gdpr undefined error

Bielik20:


platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -1,163 +1,201 @@
...
+			geoRequiresConsent: true,
+			ccpaSignal: false,
+			geoRequiresSignal: true,
+		});
 	}
 
-	async init(): Promise<void> {
-		const initPromise = this.initInstances();
+	private async loadTrackingOptInLibrary(): Promise<void> {
+		const trackingOptInLibraryUrl =
```
dmnsobczak, line 65: In the future we can keep this in the ICBM variable like Prebid library URL, but I'll add this to the list of cool features in TCFv2 ticket I guess.

platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -1,163 +1,201 @@
...
+		} catch (e) {
+			return this.handleLibraryTimeout();
+		}
+	}
 
-		window.ads.consentQueue.flush();
+	private handleDisabledCookies(): void {
+		utils.logger(logGroup, 'Cookies are disabled. Ignoring CMP and USAPI consent checks.');
+		this.dispatchConsents({
+			gdprConsent: true,
```
dmnsobczak, line 57: I'm not sure if we don't need to set `gdprConsent` to `false` if cookies are disabled, but that also mostly apply TCFv2 standard. I'll read some documentation and prepare a patch in the TCFv2 PR if needed.

Bielik20: I would appreciate that. I copied it from how it was before - but better to check.

PR #847, opened 2020-07-07
ADEN-10377 | Add Audigent service

bckatarzyna:


src/ad-services/audigent/index.ts
```diff
@@ -0,0 +1,61 @@
...
+				this.setup();
+				communicationService.dispatch(audigentLoadedEvent({}));
+			});
+			this.isLoaded = true;
+		}
+	}
+
+	configureAudienceTag(): void {
+		// @ts-ignore
+		(function(w,d,t,u){var a=d.createElement(t);a.async=1;a.src=u+'?url='+encodeURIComponent(w.location.href)+'&ref='+encodeURIComponent(d.referrer);var s=d.getElementsByTagName(t)[0];s.parentNode.insertBefore(a,s);})(window,document,'script','https://a.ad.gt/api/v1/u/matches/158');
```
fraszczakszymon, line 41: We have a dedicated module that does exactly the same thing. Use scriptLoader, please

src/ad-services/audigent/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+			window.googletag.cmd = window.googletag.cmd || [];
+			window.googletag.cmd.push(() => {
+				const segments = au_segments.length ? au_segments : 'no_segments';
+
+				googletag.pubads().setTargeting('AU_SEG', segments);
+				context.set('targeting.AU_SEG', segments);
+			});
+		}
+	}
```
fraszczakszymon, line 58: We should not follow these docs 1:1. GptProvider in AdEngine takes targeting directly from context. Set `AU_SEG` value in the context `targeting`, please

.prettierignore
```diff
@@ -5,6 +5,7 @@ node_modules/**/*
 jspm_packages/
 src/ad-products/video/player/porvata/moat/moat-video-tracker-script.ts
 src/ad-products/video/porvata4/plugins/moat/moat-video-tracker-script.ts
+src/ad-services/audigent/index.ts
```
fraszczakszymon, line 8: This line won't be needed if we address the other two comments

src/ad-services/audigent/index.ts
fraszczakszymon: googletag.cmd callback is not needed then

src/ad-services/audigent/index.ts
```diff
@@ -0,0 +1,61 @@
...
+	configureAudienceTag(): void {
+		// @ts-ignore
+		(function(w,d,t,u){var a=d.createElement(t);a.async=1;a.src=u+'?url='+encodeURIComponent(w.location.href)+'&ref='+encodeURIComponent(d.referrer);var s=d.getElementsByTagName(t)[0];s.parentNode.insertBefore(a,s);})(window,document,'script','https://a.ad.gt/api/v1/u/matches/158');
+	}
+
+	setup(): void {
+		// @ts-ignore
+		if (typeof au_seg !== 'undefined') {
+			// @ts-ignore
+			const au_segments = au_seg.segments || [];
```
fraszczakszymon, line 48:
```suggestion
			const auSegments = window['au_seg'].segments || ['no_segments'];
```
and L52 seems not needed then

src/ad-services/audigent/index.ts
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+
+	configureAudienceTag(): void {
+		// @ts-ignore
+		(function(w,d,t,u){var a=d.createElement(t);a.async=1;a.src=u+'?url='+encodeURIComponent(w.location.href)+'&ref='+encodeURIComponent(d.referrer);var s=d.getElementsByTagName(t)[0];s.parentNode.insertBefore(a,s);})(window,document,'script','https://a.ad.gt/api/v1/u/matches/158');
+	}
+
+	setup(): void {
+		// @ts-ignore
```
fraszczakszymon, line 45: I'm not sure but if we define au_seg type in types.d, then `// @ts-ignore` won't be needed

PR #853, opened 2020-07-13
ADEN-10417 Add pagesWithNoAds support to Gamepedia

Bielik20:


platforms/gamepedia/setup/context/base/gamepedia-base-context.setup.ts
```diff
@@ -0,0 +1,20 @@
+import { BaseContextSetup } from '@platforms/shared';
+import { context } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+
+@Injectable()
+export class GamepediaBaseContextSetup extends BaseContextSetup {
```
Loczi94, line 6: Why to implement it in GamepediaBaseContextSetup while it has nothing to do with setting up the context? How about moving this code to GamepediaNoAdsMode class?

Bielik20: No ads mode fires after no ads check. It would have no effect there.
I hope it will be refactored with Startup 2.0

PR #854, opened 2020-07-16
ADEN-10434 UCP new startup POC

Bielik20:


platforms/ucp/ucp-platform.ts
```diff
@@ -0,0 +1,70 @@
...
+			UcpA9ConfigSetup,
+			UcpDynamicSlotsSetup,
+			UcpSlotsStateSetup,
+			CommonBiddersStateSetup,
+			UcpTemplatesSetup,
+			UcpBillTheLizardSetup,
+			LabradorSetup,
+			TrackingSetup,
+			AdEngineRunnerSetup,
+			() => communicationService.dispatch(adEngineConfigured()),
```
fraszczakszymon, line 57: It seems the order has changed. Previously the event was dispatched before wiki context setup, am I right?

Bielik20: I feel like it was after configure: https://github.com/Wikia/ad-engine/blob/f144e3d9bd51c0c722ae55eeeaca182496fc4aa0/platforms/ucp/index.ts#L20-L21

fraszczakszymon: Sorry, I mixed configure with run method. You're right

platforms/ucp/ucp-platform.ts
```diff
@@ -0,0 +1,70 @@
...
+export class UcpPlatform {
+	constructor(private pipeline: ProcessPipeline, private noAdsDetector: NoAdsDetector) {}
+
+	execute(): void {
+		// Config
+		this.pipeline.add(
+			() => context.extend(basicContext),
+			parallel(UcpIocSetup, () => bootstrapAndGetConsent()),
+			UcpWikiContextSetup,
+			() => context.set('state.isMobile', false),
```
Loczi94, line 43: Does it have to be here? 🙈  Maybe we can move it to UcpBaseContextSetup?

PR #856, opened 2020-07-20
ADEN-10443 | IX ID - track ids when all are loaded

bckatarzyna:


src/ad-bidders/prebid/identity-library/index.ts
```diff
@@ -33,25 +33,61 @@ class IdentityLibrary {
...
+			if (
+				identityInfo.hasOwnProperty(rtiPartner) &&
+				identityInfo[rtiPartner].responsePending === true
+			) {
+				responsePending = 1;
+				break;
+			}
+		}
+
+		if (responsePending) {
```
fraszczakszymon, line 57:
```suggestion
		const hasPendingResponses = Object.values(identityInfo).some(response => response.responsePending);
		
		if (hasPendingResponses) {
```

Is it the same?

src/ad-bidders/prebid/identity-library/index.ts
```diff
@@ -33,25 +33,61 @@ class IdentityLibrary {
...
+			window.headertag.subscribeEvent(
+				'rti_partner_request_complete',
+				false,
+				(eventName, eventData) => {
+					communicationService.dispatch(identityLibraryIdsLoadedEvent({}));
+				},
+			);
```
fraszczakszymon, line 64:
```suggestion
			window.headertag.subscribeEvent('rti_partner_request_complete', false, () => communicationService.dispatch(identityLibraryIdsLoadedEvent({})));
```

PR #862, opened 2020-07-27
ADEN-10462 | Call Audigent on platforms

bckatarzyna:


src/ad-services/audigent/index.ts
```diff
@@ -51,3 +50,4 @@ class Audigent {
 }
 
 export const audigent = new Audigent();
+export const audigentLoadedEvent = globalAction('[AdEngine] Audigent loaded', props<{}>());
```
Bielik20, line 53: Are you sure this needs to be a global action I couldn't find usage anywhere?

PR #869, opened 2020-08-04
[ADEN-10467] Create BingeBot bundle MVP

Loczi94:


platforms/bingebot/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Container } from '@wikia/dependency-injection';
+import { BingeBotPlatform } from './bingebot-platform';
+
+window.RLQ = window.RLQ || [];
+window.RLQ.push(async () => {
```
fraszczakszymon, line 5: Most likely it won't exist on BingeBot app, am I right?

PR #872, opened 2020-08-11
ADEN-10449 UCP mobile-wiki bundle

fraszczakszymon:


platforms/ucp-mercury/index.ts
```diff
@@ -0,0 +1,69 @@
...
+		adContext: payload,
+	});
+
+	const container = new Container();
+	container.bind(MEDIA_WIKI_ADS_CONTEXT).value(payload);
+
+	const platform = container.get(UcpMercuryPlatform);
+
+	platform.execute();
+}
```
Bielik20, line 50: Why not make that a step?

platforms/ucp-mercury/styles.scss
```diff
@@ -0,0 +1,95 @@
...
+
+		#bottom_leaderboard {
+			@include bfab-slot();
+		}
+
+		#top_leaderboard,
+		#hivi_leaderboard,
+		#bottom_leaderboard {
+			@include uap-porvata();
+			@include uap-porvata-right-position();
```
Bielik20, line 33: I feel like this should be `uap-porvata-learn-more`.
`uap-porvata-right-position` is for desktop.

platforms/ucp-mercury/ucp-mercury-ioc-setup.ts
```diff
@@ -0,0 +1,36 @@
...
+import * as fallbackInstantConfig from './fallback-config.json';
+
+@Injectable()
+export class UcpMercuryIocSetup implements DiProcess {
+	constructor(private container: Container) {}
+
+	async execute(): Promise<void> {
+		set(window, context.get('services.instantConfig.fallbackConfigKey'), fallbackInstantConfig);
+
+		this.container.bind(InstantConfigService).value(await InstantConfigService.init());
```
Bielik20, line 23: We don't set fallback on window anymore, but in context.
Also let's use `InstantConfigSetup` step instead of that.

platforms/ucp-mercury/ucp-mercury-platform.ts
```diff
@@ -0,0 +1,58 @@
...
+export class UcpMercuryPlatform {
+	static executed = false;
+
+	constructor(private pipeline: ProcessPipeline, private noAdsDetector: NoAdsDetector) {}
+
+	execute(): void {
+		// Config
+		this.pipeline.add(
+			() => context.extend(basicContext),
+			UcpMercuryIocSetup,
```
Bielik20, line 31: Will we be adding TrackingOptIn in here?

platforms/ucp-mercury/ucp-mercury-platform.ts
```diff
@@ -0,0 +1,58 @@
...
+			CommonBiddersStateSetup,
+			LabradorSetup,
+		);
+
+		if (!UcpMercuryPlatform.executed) {
+			this.pipeline.add(TrackingSetup, AdEngineRunnerSetup, () => {
+				communicationService.dispatch(adEngineConfigured());
+				UcpMercuryPlatform.executed = true;
+			});
+		}
```
Bielik20, line 46: How about making this `conditional` step?

src/ad-engine/rxjs/index.ts
```diff
@@ -1,4 +1,3 @@
-export * from './rxjs-operator';
```
Bielik20, line 1: I feel like rxjs operator should be inside `core`. I know you have a problem here with circular dependencies but still it doesn't fit inside `communication`.

Maybe extract models (with rxjs) to separate sub module?

platforms/ucp-mercury/ucp-mercury-platform.ts
fraszczakszymon: TrackingOptIn is loaded from Mobile-Wiki platform for now

platforms/ucp-mercury/index.ts
fraszczakszymon: Good point!

platforms/ucp-mercury/styles.scss
fraszczakszymon: I forgot about styles at all. I don't feel like we need any of them right now. It will be better once we add necessary styles during UAP/outstream development.

platforms/ucp-mercury/setup/hooks/ucp-mercury-after-transition-setup.ts
```diff
@@ -0,0 +1,34 @@
...
+			.pipe(ofType(mercuryAfterTransition), onlyNew())
+			.subscribe(({ payload }: { payload: MediaWikiAdsContext }) => this.afterTransition(payload));
+	}
+
+	private afterTransition(adContext: MediaWikiAdsContext): void {
+		eventService.emit(events.PAGE_RENDER_EVENT, {
+			adContext,
+		});
+
+		const container = new Container();
```
Bielik20, line 27: We shouldn't create new container. Use one already existing.

spec/ad-engine/pipeline/imps/process-pipeline/compound-processes/once-process.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+	});
+
+	it('should work', async () => {
+		pipeline.add(funcProcess, once(executableOnceProcess), () => spy('end'));
+
+		await pipeline.execute();
+
+		assertResults(['func', 'once', 'end']);
+
+		spy = sandbox.spy();
```
Bielik20, line 32: You could just clear history `spy.reset/restory ... something`.

spec/ad-engine/pipeline/imps/process-pipeline/compound-processes/once-process.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+		pipeline = container.get(ProcessPipeline);
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	it('should work', async () => {
+		pipeline.add(funcProcess, once(executableOnceProcess), () => spy('end'));
```
Bielik20, line 26: Would good to add arrow function step, like:
```ts
once(() => executableOnceProcess)),
```

platforms/ucp-mercury/setup/hooks/ucp-mercury-after-transition-setup.ts
```diff
@@ -24,10 +26,9 @@ export class UcpMercuryAfterTransitionSetup implements DiProcess {
 			adContext,
 		});
 
-		const container = new Container();
-		container.bind(MEDIA_WIKI_ADS_CONTEXT).value(adContext);
+		this.container.bind(MEDIA_WIKI_ADS_CONTEXT).value(adContext);
```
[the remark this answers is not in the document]
fraszczakszymon: I have mixed feelings about this. We can have that step, but then we'll need to keep these executions from this file in the same order (sync context setup and platform.execute)

spec/ad-engine/pipeline/imps/process-pipeline/compound-processes/once-process.spec.ts
fraszczakszymon: Why would you change it? It can't be applied in the second test.

PR #874, opened 2020-08-12
[ADEN-10468] BingeBot: sponsored logo ad slot

Loczi94:


platforms/bingebot/ad-context.ts
```diff
@@ -8,7 +8,16 @@ export const basicContext = {
...
+	state: {
+		adStack: [],
+	},
 	targeting: {
 		skin: 'bingebot',
 	},
+	services: {
+		instantConfig: {
+			endpoint: 'https://services.wikia.com/icbm/api/config?app=bingebot',
+			fallbackConfigKey: 'fallbackInstantConfig',
```
Bielik20, line 20: We don't do that anymore. We have fallback inside context.

platforms/bingebot/setup/dynamic-slots/bingebot-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const adSlotInjected = globalAction(
+	'[BingeBot] ad slot injected',
+	props<AdSlotInjectedProps>(),
+);
+
+export const destroyAdSlot = globalAction(
+	'[BingeBot] destroy ad slot',
+	props<AdSlotInjectedProps>(),
+);
```
Bielik20, line 18: Do we need to export that?

platforms/bingebot/setup/dynamic-slots/bingebot-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable()
+export class BingeBotDynamicSlotsSetup implements DynamicSlotsSetup {
+	execute(): void {
+		communicationService.action$.pipe(ofType(adSlotInjected)).subscribe((action) => {
+			this.setAdStack(action.slotId);
+		});
+
+		communicationService.action$.pipe(ofType(destroyAdSlot)).subscribe((action) => {
+			const adSlot = slotService.slots[action.slotId];
```
Bielik20, line 28: What about `slotService.get`?

platforms/bingebot/styles.scss
```diff
@@ -0,0 +1,7 @@
+#sponsored_logo::before {
+	content: 'promoted by';
+	font-size: 8px;
+	font-weight: bold;
+	text-transform: uppercase;
+	margin-bottom: 5px;
+}
```
Bielik20, line 7: How about moving it to AdEngine?

platforms/bingebot/index.ts
```diff
@@ -1,5 +1,6 @@
 import { Container } from '@wikia/dependency-injection';
 import { BingeBotPlatform } from './bingebot-platform';
+import './styles.scss';
```
Bielik20, line 3: ditto

Bielik20 (changes requested):

platforms/bingebot/styles.scss
Loczi94: It is in AdEngine. 🙃  You mean moving it to `src/ad-products/styles/styles.scss`? My point of view is that this style applies to element which exists only in BingeBot app so imo there is no point in moving it from here.

Bielik20 (approved):

PR #884, opened 2020-08-24
ADEN-10540 Move TrackingOptIn version to ICBM variable

dmnsobczak:


platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -65,10 +74,11 @@ class TrackingOptInWrapper {
 	}
 
 	private async loadTrackingOptInLibrary(): Promise<void> {
-		const trackingOptInLibraryUrl =
-			'//static.wikia.nocookie.net/fandom-ae-assets/tracking-opt-in/v4.0.8/tracking-opt-in.min.js';
+		const instantConfig = await InstantConfigService.init();
 
-		await utils.scriptLoader.loadScript(trackingOptInLibraryUrl);
+		await utils.scriptLoader.loadScript(
+			instantConfig.get('icTrackingOptInVersion') || trackingOptInLibraryUrl,
```
nandy-andy, line 80: `icTrackingOptInVersion` won't just contain, the version, will it? It looks like it'll contain an URL to the library, so I vote to rename it to `icTrackingOptInLibraryUrl` :-)

dmnsobczak: Well, I took unification rather than readability, because Prebid's variable is also called `icPrebidVersion` and it also contains full library URL. Is that OK?

dmnsobczak: Discussed on Slack. Name changed.

nandy-andy (approved):

PR #892, opened 2020-08-31
[ADEN-10571] BingeBot: update correlator value before new view rendered

Loczi94:


platforms/bingebot/setup/hooks/bingebot-before-view-rendered-setup.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class BingeBotBeforeViewRenderedSetup implements DiProcess {
+	async execute(): Promise<void> {
+		communicationService.action$
+			.pipe(ofType(destroyAdSlot), onlyNew())
+			.subscribe(() => this.updateCorrelator());
+	}
+
+	private updateCorrelator(): void {
+		window.googletag.pubads().updateCorrelator();
```
fraszczakszymon, line 12: We shouldn't use `window.googletag.pubads` directly. We have a wrapper for it - `gpt-provider`.

fraszczakszymon (changes requested):
It seems to be a workaround - we're subscribing to the `destroy ad slot` message and we're naming it "before view rendered". In my opinion, the `BingeBotBeforeViewRenderedSetup` name does not describe the situation. It's going to be problematic in case we have more slots per view. Also, if it turns out that `view rendered` happens too late (for updating correlator), it probably means that it might be too late for updating `targeting.s2` value too.

We're using useEffect on each view to dispatch `view rendered` message ([example](https://github.com/Wikia/bingebot-front-end/blob/master/app/src/components/Results/ResultsLoading.js#L32)). What do you think about returning a callback which will dispatch `before view change` message and subscribe to it for updating the correlator?

fraszczakszymon (approved):
Approving my own changes 🤓. It's going to wait for someone's else code review.

PR #897, opened 2020-09-17
ADEN-10549 Adding affiliate_slot for IntentX contextual widget

dmnsobczak:


platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
```diff
@@ -17,7 +18,13 @@ export class UcpSlotsStateSetup implements DiProcess {
...
 		slotsContext.setState('hivi_leaderboard', !!context.get('options.hiviLeaderboard'));
 		slotsContext.setState('top_leaderboard', true);
-		slotsContext.setState('top_boxad', true);
+		slotsContext.setState('top_boxad', this.isRightRailApplicable());
+		slotsContext.setState(
+			'affiliate_slot',
+			this.isRightRailApplicable() &&
+				context.get('wiki.opts.enableAffiliateSlot') &&
+				!context.get('custom.hasFeaturedVideo'),
+		);
```
Loczi94, line 27: How about moving all this conditions to a new function e.g. `isAffiliateSlotEnabled`?

Loczi94 (approved):

platforms/ucp/setup/state/slots/ucp-slots-state-setup.ts
dmnsobczak: Good suggestion, I was thinking about it. Moved.

PR #903, opened 2020-09-29
ADEN-10315 Handle video ads in IAS PO

fraszczakszymon:


src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -50,21 +51,26 @@ class IasPublisherOptimization {
...
-			});
+				adSlotId: slotName,
+				size: this.getSlotSize(slot),
+			};
+
+			if (slot.isVideo) {
+				config.type = 'video';
+			}
+
+			iasPETSlots.push(config);
```
nandy-andy, line 73: Do we reset/clean the `iasPETSlots` array on SPA in some way?

nandy-andy: Discussed on slack :)

nandy-andy (approved):

PR #906, opened 2020-10-06
GPUCP-600 | Always prepend the ATF to the top of the siderail.

Alexia:
Right there due to asynchronous loading there is a high chance the `atfmrec` appears below the `middlemrec`.  Instead of trying to guess if it needs to inserted before just always prepend to the top of the node.

https://wikia-inc.atlassian.net/browse/GPUCP-600


Alexia:
Note: I do not have a way to test this and I assume it should work.

nandy-andy (approved):

platforms/ucp-hydra/setup/dynamic-slots/hydra-dynamic-slots-setup.ts
```diff
@@ -63,13 +63,7 @@ export class HydraDynamicSlotsSetup implements DiProcess {
...
 			incontentBoxad.id = 'incontent_boxad_1';
 
-			const gamepediaProBoxWrapper = document.getElementById(`middlemrec_${dbName}_gamepedia`);
-
-			if (gamepediaProBoxWrapper) {
-				siderail.insertBefore(topBoxadWrapper, gamepediaProBoxWrapper);
-			} else {
-				siderail.appendChild(topBoxadWrapper);
-			}
+			siderail.prepend(topBoxadWrapper);
```
nandy-andy, line 66: That should work, thanks! It may get more complicated when we start working on the ad layout unification and the affiliate units will be added but for now, it looks like a solid fix 👍

PR #910, opened 2020-10-13
ADEN-10640 Adding Top1k wikis targeting setting

dmnsobczak:


platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -58,6 +63,10 @@ export class UcpTargetingSetup implements DiProcess {
 			targeting.cid = cid;
 		}
 
+		if (wiki.targeting.wikiIsTop1000) {
+			targeting.top = '1k';
+		}
+
```
Loczi94, line 69: Key val on oasis is named `top_1k` and imo it could be set in the same way:
`top_1k: context.get('wiki.targeting.wikiIsTop1000') ? 1 : 0`

See: https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/ml/configuration.js#L48

Loczi94 (changes requested):

platforms/shared/context/targeting/ucp-targeting.setup.ts
dmnsobczak: You linked Queen of Hearts targeting. I copied this keyval name and value from general, page targeting:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/targeting.js#L232

Loczi94 (approved):

PR #911, opened 2020-10-13
ADEN-10628 Add Permutive to BingeBot

nandy-andy:


platforms/bingebot/modes/bingebot-ads.mode.ts
```diff
@@ -1,10 +1,19 @@
...
+		startAdEngine(inhibitors);
+	}
+
+	private callExternals(): Promise<any>[] {
+		const inhibitors: Promise<any>[] = [];
+
+		permutive.call();
+
+		return inhibitors;
 	}
```
Loczi94, line 18:
```suggestion
		this.callExternals();
		startAdEngine();
	}

	private callExternals(): void {
		permutive.call();
	}
```

There is no need to create `inhibitors` array since we do not push anything to it.

Loczi94 (changes requested):

platforms/shared/setup/base-context.setup.ts
```diff
@@ -117,11 +117,13 @@ export class BaseContextSetup implements DiProcess {
...
 			this.instantConfig.get('icIASPublisherOptimization'),
 		);
+		context.set('services.nielsen.enabled', this.instantConfig.get('icNielsen'));
+
+		const isPageDirectedAtChildren = !!context.get('wiki.targeting.directedAtChildren');
 		context.set(
 			'services.permutive.enabled',
-			this.instantConfig.get('icPermutive') && !context.get('wiki.targeting.directedAtChildren'),
+			this.instantConfig.get('icPermutive') && !isPageDirectedAtChildren,
 		);
```
Loczi94, line 126: We have a method `isEnabled()` in `Permutive` class where we check all conditions. What do you think about including this condition in this method? Also I don't think we need to force boolean since we get `undefined` from context and it's a falsy value in JS. 🤔

nandy-andy: Yep, I think we could move checking `context.get('wiki.targeting.directedAtChildren')` to `isEnabled` in `Permutive` 👌

The issue with `undefined` in this case is that its falsy nature makes Permutive enabled by default if `this.instantConfig.get('icPermutive')` is `true`. That's why I think we need to force the boolean :P

platforms/bingebot/ad-context.ts
```diff
@@ -16,6 +16,11 @@ export const basicContext = {
 	targeting: {
 		skin: 'bingebot',
 	},
+	wiki: {
+		targeting: {
+			directedAtChildren: false,
```
Loczi94, line 21: IMHO we don't need it because `wiki.targeting.directedAdChildren` will have falsy value (`undefined`) anyway and we should be aware that it's set to `false` in ad-contex. It's up to you if you want to leave it as is.

Loczi94 (approved):

platforms/bingebot/ad-context.ts
nandy-andy: 👌  discussed offline

PR #915, opened 2020-10-20
ADEN-10661 De-SPA mercury-ucp bundle for New Mobile-Wiki Experience

dmnsobczak:


platforms/ucp-mercury/index.ts
```diff
@@ -1,19 +1,11 @@
-import { communicationService, ofType } from '@wikia/ad-engine';
 import { Container } from '@wikia/dependency-injection';
-import { take } from 'rxjs/operators';
-import { mercuryInit } from './setup-ucp-mercury';
-import { MEDIA_WIKI_ADS_CONTEXT } from './setup/context/wiki/ucp-mercury-wiki-context.setup';
```
nandy-andy, line 5: My first thought: oh no, ad context is important! Why are we deleting it?! ;)

Second thought: in SPA we needed to wait for a `MercuryApi` response which included the ad context. With new mobile web skin we have the ad context without sending API requests 🥳 Therefore, we don't need this line! :)

nandy-andy (approved):

platforms/ucp-mercury/index.ts
dmnsobczak: Yes, we can get context UCP-style: from `window` global object and that is covered in `WikiContextSetup` base class :)

platforms/ucp-mercury/ucp-mercury-platform.ts
```diff
@@ -27,27 +26,21 @@ import { UcpMercuryPrebidConfigSetup } from './setup/context/prebid/ucp-mercury-
...
-import { UcpMercuryOnTransitionSetup } from './setup/hooks/ucp-mercury-on-transition-setup';
 import { UcpMercuryTemplatesSetup } from './templates/ucp-mercury-templates-setup.service';
 import { UcpMercuryIocSetup } from './ucp-mercury-ioc-setup';
 
 @Injectable()
 export class UcpMercuryPlatform {
 	constructor(private pipeline: ProcessPipeline, private noAdsDetector: NoAdsDetector) {}
 
-	setup(): void {
+	execute(): void {
```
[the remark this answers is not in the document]
dmnsobczak: Just for unification with other platforms: we have only execute there, there is no need to do this in two steps.

Loczi94 (approved):

Loczi94 (approved):

PR #918, opened 2020-10-27
ADEN-10662 Reintroducing all Fandom ad placements

dmnsobczak:


platforms/ucp-mobile/index.ts
```diff
@@ -2,10 +2,30 @@ import { Container } from '@wikia/dependency-injection';
...
+		section.setAttribute('aria-expanded', visible);
+	}
+}
+
 window.RLQ = window.RLQ || [];
 window.RLQ.push(async () => {
 	const container = new Container();
 	const platform = container.get(UcpMobilePlatform);
 
+	uncollapseSections();
```
nandy-andy, line 28: Will it uncollapse sections also when ads will be disabled (showAds=false or externals=0 or logged-in users)?

dmnsobczak: Yes, one exception: AdBlock. But we can move this part of code to MobileSkin code in UCP repository in the future (just like for Mobile Legacy at this moment).

platforms/ucp-mobile/styles.scss
```diff
@@ -1,3 +1,26 @@
...
+
+// #######################################
+
+$closeButtonHeight: 40px;
+
+%high-impact-close-icon {
+	fill: white;
+	height: $closeButtonHeight / 2;
+	width: $closeButtonHeight / 2;
+}
```
Loczi94, line 22: According to Sass docs: https://sass-lang.com/documentation/style-rules/placeholder-selectors placeholder selector is not included in the CSS output. It can be extended but we don't use it in such a way thus I think this parts is needless here. Please, correct me if I'm wrong.

platforms/ucp-mobile/styles.scss
```diff
@@ -11,3 +34,185 @@
...
+			width: auto;
+
+			svg {
+				position: static;
+			}
+		}
+	}
+}
+
+//show the loading spinner on top of ad in interstitial
```
Loczi94, line 181: Will we have interstitial in ucp mobile? If no, please delete this part.

platforms/ucp-mobile/styles.scss
dmnsobczak: You're right, removed :)

platforms/ucp-mobile/styles.scss
dmnsobczak: Sure, removed for now. I will check whether we want to reimplement interstitial or not (hopefully not).

Loczi94 (approved):

PR #920, opened 2020-10-28
ADEN-10679 Fix LiveRamp events

nandy-andy:


src/ad-bidders/prebid/index.ts
```diff
@@ -258,12 +262,16 @@ export class PrebidProvider extends BidderProvider {
...
 
+			utils.logger(logGroup, 'prebid dispatching LiveRamp event');
+
 			liveRamp.dispatchLiveRampPrebidIdsLoadedEvent(userId);
 		}
 	}
 
 	private enableATSAnalytics(): void {
 		if (this.isATSAnalyticsEnabled) {
+			utils.logger(logGroup, 'prebid enabling ATS Analytics');
```
Loczi94, line 273: There is already log from Prebid which says that ATS Analytics is enabled.

src/ad-bidders/prebid/index.ts
```diff
@@ -258,12 +262,16 @@ export class PrebidProvider extends BidderProvider {
 		if (pbjs.getUserIds) {
 			const userId = pbjs.getUserIds()['idl_env'];
 
+			utils.logger(logGroup, 'prebid dispatching LiveRamp event');
```
Loczi94, line 265: This log can be misleading since the event is not always dispatched - see dispatchLiveRampPrebidIdsLoadedEvent() in LiveRamp class.

src/ad-bidders/prebid/live-ramp/live-ramp.ts
```diff
@@ -32,8 +34,16 @@ class LiveRamp {
 	}
 
 	dispatchLiveRampPrebidIdsLoadedEvent(userId): void {
-		if (this.isEnabled()) {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
```
Loczi94, line 38: We print this message in getConfig() and imo once is enough. 😄

Loczi94 (changes requested):

src/ad-bidders/prebid/index.ts
nandy-andy: What do you suggest? Maybe naming could be improved as it's not really dispatching but rather registering? 🤔

src/ad-bidders/prebid/index.ts
nandy-andy: I assume you're referring to ATS module. I wanted to have confirmation from the prebid module as well, so I was able to see in logs how the application behaves in SPA environment.

src/ad-bidders/prebid/live-ramp/live-ramp.ts
nandy-andy: It isn't when you're checking how `Prebid` and `LiveRamp` modules behave in SPA ;)

Loczi94 (approved):

PR #921, opened 2020-10-29
ADEN-10556 | ats.js

Loczi94:


src/ad-bidders/prebid/ats/ats.ts
```diff
@@ -0,0 +1,88 @@
...
+				}
+
+				return Promise.resolve();
+			}
+
+			const performance = window.performance;
+			const loadStart = performance.now();
+
+			// tslint:disable-next-line:tsl-ban-snippets
+			return utils.scriptLoader.loadScript(this.atsScriptSrc).then(() => {
```
nandy-andy, line 35: Don't we want to handle/track errors when loading the script?

Loczi94: We don't do it when loading other scripts (e.g. Audigent, IAS PO), also when discussing tracking we didn't plan to track errors. Imo we can add it in the future if needed. What do you think?

nandy-andy: I think it's useful information we can pass to us and our partners. Let's create a follow-up ticket! 💪

nandy-andy (approved):

PR #928, opened 2020-11-06
ADEN-8879 add adContext and debug properties to window.ads

KonradLinkowski:


src/ad-engine/services/context-service.ts
```diff
@@ -155,3 +155,5 @@ export class Context {
 }
 
 export const context = new Context();
+
+window.ads.adContext = context;
```
dmnsobczak, line 159: I don't know if this is a very elegant place for that. Maybe we can set this variable in the `Context` class constructor, like:
```
Context() {
    window.ads.adContext = this;
}
```
?

dmnsobczak: Also: we don't need to make the whole class Context public with all the set and get methods. We can just return the result of `content.get('')`, a simple object of values.

src/ad-engine/ad-engine.ts
```diff
@@ -31,12 +32,14 @@ export class AdEngine {
...
 	provider: Provider;
 	adStack: OldLazyQueue<AdStackPayload>;
+	cookieStorage = new CookieStorageAdapter();
 
 	constructor(config = null) {
 		context.extend(config);
 
 		window.ads = window.ads || ({} as MediaWikiAds);
 		window.ads.runtime = window.ads.runtime || ({} as Runtime);
+		window.ads.debug = (groups: string) => this.cookieStorage.setItem('adengine_debug', groups);
```
dmnsobczak, line 42: I didn't even know that we were already checking if `adengine_debug` cookie exists :o Nice.

But one suggestion: we can easily enable debug mode: can we also add a method to disable it? It's simpler than looking for and removing cookie manually.

dmnsobczak: https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/utils/logger.ts#L7
This chunk of code happens only at the beginning, during AdEngine start. Is is any way to enable/disable debug mode during AdEngine work without reloading the page? 
If not, maybe it's a good ideal to add automatic page refresh after enabling debug mode?

dmnsobczak: Another bonus: Any message in console like: `AdEngine Debug Mode enabled`?

src/ad-engine/services/context-service.ts
dmnsobczak: Linter is failing:
`TypeError: Cannot set property 'adContext' of undefined`
Maybe this will fix the issue?
```
window.ads = window.ads || {};
window.ads.adContext = window.ads.adContext || context;
```

dmnsobczak (changes requested):
Linter + few suggestions after code review.

src/ad-engine/ad-engine.ts
KonradLinkowski: > But one suggestion: we can easily enable debug mode: can we also add a method to disable it? It's simpler than looking for and removing cookie manually.

I'm pretty sure that using `window.ads.debug('0')` disables debug mode.

> Is is any way to enable/disable debug mode during AdEngine work without reloading the page?

I could make an interval checking if debug mode changed, but it probably isn't a good practice. I believe that there is other way of doing this, but I don't know ad-engine very well right now.

>  Any message in console

Sure

src/ad-engine/services/context-service.ts
KonradLinkowski: > Also: we don't need to make the whole class Context public with all the set and get methods. We can just return the result of content.get(''), a simple object of values.

I thought that the task is about adding access to the whole context object, but it makes sense to expose only the `get` method as you pointed out. Maybe it would be even better if this method would return whole context data as json or js object, but I'm not sure about the usage though.

> Linter is failing

Husky hooks fired when I created the commit, but the linter didn't detect anything then so that's weird.

src/ad-engine/ad-engine.ts
```diff
@@ -31,12 +32,17 @@ export class AdEngine {
...
 	provider: Provider;
 	adStack: OldLazyQueue<AdStackPayload>;
+	cookieStorage = new CookieStorageAdapter();
 
 	constructor(config = null) {
 		context.extend(config);
 
 		window.ads = window.ads || ({} as MediaWikiAds);
 		window.ads.runtime = window.ads.runtime || ({} as Runtime);
+		window.ads.debug = (groups: string | null) =>
```
dmnsobczak, line 42: What will happen if I will run `window.ads.debug()`? I think that this part of code won't run correctly:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/utils/logger.ts#L16
Can we make default value of `groups` parameter as `1`? Then `debug()` will enable debug mode and `debug(null)` will disable it.

src/ad-engine/services/context-service.ts
```diff
@@ -116,6 +116,10 @@ function segment(key: string, newValue?: any, remove = false): any {
 }
 
 export class Context {
+	constructor() {
+		window.ads.adContext = contextObject;
```
dmnsobczak, line 120: Make sure to add to QA notes memo to check whether `adContext` object is always up-to-date, no matter when we'll try to read it.

dmnsobczak (changes requested):
Linter is still failing. Try running `npm run lint` locally before pushing new commits.

Also, what about those two?
> If not, maybe it's a good ideal to add automatic page refresh after enabling debug mode?
> Another bonus: Any message in console like: AdEngine Debug Mode enabled?

src/ad-engine/ad-engine.ts
```diff
@@ -31,12 +32,19 @@ export class AdEngine {
...
 	constructor(config = null) {
 		context.extend(config);
 
 		window.ads = window.ads || ({} as MediaWikiAds);
 		window.ads.runtime = window.ads.runtime || ({} as Runtime);
+		window.ads.debug = (groups: string | null) => {
+			groups !== null
+				? this.cookieStorage.setItem('adengine_debug', groups || '1')
+				: this.cookieStorage.removeItem('adengine_debug');
+			logger('ad-engine', 'AdEngine Debug Mode enabled');
```
dmnsobczak, line 46: `logger('ad-engine', 'AdEngine Debug Mode' + groups !== null ? 'enabled' : 'disabled');`
?

dmnsobczak (approved):

PR #930, opened 2020-11-10
ADEN-10463 Adding missing tracking adengpageprops events

dmnsobczak:


platforms/shared/tracking/tracking.setup.ts
```diff
@@ -171,6 +178,70 @@ export class TrackingSetup {
...
+
+		viewabilityCounter.init();
+	}
+
+	private scrollSpeedTracker(): void {
+		if (!context.get('options.scrollSpeedTracking.enabled')) {
+			return;
+		}
+
+		const scrollTracker = new ScrollTracker([0, 2000, 4000], 'mediawiki');
```
nandy-andy, line 206: For `fandom-mobile` the application area class is still `'mediawiki'`?

nandy-andy (approved):

platforms/shared/tracking/tracking.setup.ts
dmnsobczak: Yep, on Mobile Legacy it was `application-wrapper`. Discussed on Slack.

nandy-andy: Confirmed and discussed offline 👍

PR #934, opened 2020-11-20
ADEN-10686 rotator interval change

KonradLinkowski:


src/ad-products/utils/fmr-rotator.ts
```diff
@@ -43,9 +43,10 @@ export class FmrRotator {
 
 					if (!universalAdPackage.isFanTakeoverLoaded()) {
 						slot.once(AdSlot.SLOT_VIEWED_EVENT, () => {
+							const customDelays = context.get('options.rotatorDelay');
 							setTimeout(() => {
 								this.hideSlot();
-							}, this.refreshInfo.refreshDelay);
+							}, customDelays[slot.lineItemId] || this.refreshInfo.refreshDelay);
```
dmnsobczak, line 49: If somebody disable the icVariable completely then `customDelays` will be undefined, and `undefined[slot.lineItemId]` will result in error. What about:
`const customDelays = context.get('options.rotatorDelay') || {};`
or
`this.instantConfig.get('icRotatorDelay', {})`
?

platforms/shared/setup/base-context.setup.ts
```diff
@@ -70,6 +70,7 @@ export class BaseContextSetup implements DiProcess {
 			this.instantConfig.get('icFeaturedVideoPostroll'),
 		);
 
+		context.set('options.rotatorDelay', this.instantConfig.get('icRotatorDelay'));
```
dmnsobczak, line 73: This is setup file for new MediaWiki+AdEngine integration (aka UCP), but we are still using old AdEngine integration - without platforms in `app` repository. We need second pull request with the same change here:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/setup.js#L121

dmnsobczak (changes requested):
Possible error and second PR for app repository

platforms/shared/setup/base-context.setup.ts
KonradLinkowski: https://github.com/Wikia/app/pull/18246

dmnsobczak (approved):

PR #937, opened 2020-11-24
ADEN-10691 Reimplement FORCE_COLLAPSE scenario in HiVi_Leaderboard logic

dmnsobczak:
The change is caused by the different placement of ad slot pushes. On legacy it looked like that:
https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/src/slots.js#L385

On platforms though we grouped ad slot pushed in one place: PlatformAdsMode. This "feature" was caused by the kind of the gray zone where `force-collapse`, which for this code is something different than just `collapse` was not covered by ad slot pushes. After few tries I decided to copy the old logic.

`HiVi_Leaderboard` is currently enabled globally for Oasis. We can decide to simplify this logic during the old hivi/top-leaderboard cleanup, if we ever want to do that.


KonradLinkowski (approved):

platforms/ucp-hydra/setup/dynamic-slots/hydra-dynamic-slots-setup.ts
```diff
@@ -89,33 +81,12 @@ export class HydraDynamicSlotsSetup implements DiProcess {
 	}
 
 	private configureTopLeaderboard(): void {
-		const hiviLBEnabled = context.get('options.hiviLeaderboard');
```
fraszczakszymon, line 92: 💎

fraszczakszymon (approved):

PR #940, opened 2020-11-30
Turn off ads.adContext in prod mode

KonradLinkowski:


src/ad-engine/services/context-service.ts
```diff
@@ -115,10 +117,12 @@ function segment(key: string, newValue?: any, remove = false): any {
 	return seg[lastKey];
 }
 
+const isDebugMode = Boolean(queryString.get('adengine_debug') || Cookies.get('adengine_debug'));
+
 export class Context {
 	constructor() {
 		window.ads = window.ads || ({} as MediaWikiAds);
-		window.ads.adContext = contextObject;
+		window.ads.adContext = isDebugMode ? contextObject : {};
```
nandy-andy, line 125: Looks like the part that was missing 👍

I'm concerned that it required adding additional dependencies and almost duplicating logic you've introduced in https://github.com/Wikia/ad-engine/pull/928/files#diff-83cc4163d820af6fdec14e8344e30f59a2ce8946fa5953ffcd7ad47b5711a5baR42-R47 

It seems a silly task but once we got into details it's tricky in the details 🤔 I have a couple of different suggestions in order to minimize the code duplication:
1) move the logic from `AdStackPayload` to the context-service,
2) create a new module/reuse the existing one that will have a public method for deciding if it's a debug mode or not,
3) move the decision to reveal the ad context somewhere later than in a constructor and call it once you know (in the `AdStackPayload`) the debug mode is on (but it wouldn't be that clean, would it?) 🤔

src/ad-engine/ad-engine.ts
```diff
@@ -39,12 +39,6 @@ export class AdEngine {
 
 		window.ads = window.ads || ({} as MediaWikiAds);
 		window.ads.runtime = window.ads.runtime || ({} as Runtime);
-		window.ads.debug = (groups: string | null) => {
-			groups !== null
-				? this.cookieStorage.setItem('adengine_debug', groups || '1')
```
fraszczakszymon, line 44: It seems AdEngine.cookieStorage can be removed too, am I wrong?

fraszczakszymon (approved):

fraszczakszymon (approved):

PR #945, opened 2020-12-04
ADEN-10731 Bug fixes for ucp-mobile: tracking, skin name, vm ad unit

fraszczakszymon:


platforms/shared/tracking/data-warehouse.ts
```diff
@@ -113,7 +113,7 @@ export class DataWarehouseTracker {
 			utils.logger(logGroup, msg);
 			throw new Error(msg);
 		}
-		return `${baseUrl}?${queryString.stringify(params)}`;
+		return `${baseUrl}?${queryString.stringify(params, { strict: false })}`;
```
fraszczakszymon, line 116: We don't need "stricter" encoding. Using `encodeURIComponent` is enough in that case.
Read more here: https://github.com/sindresorhus/query-string/blob/master/index.d.ts#L195

dmnsobczak (approved):

PR #946, opened 2020-12-07
ADEN-10647-2 Increase TrackingOptIn timeout delay

dmnsobczak:


nandy-andy (approved):

platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -36,7 +36,7 @@ const setOptInInstances = globalAction(
 );
 
 const trackingOptInLibraryUrl =
-	'//static.wikia.nocookie.net/fandom-ae-assets/tracking-opt-in/v5.0.0/tracking-opt-in.min.js';
+	'//static.wikia.nocookie.net/fandom-ae-assets/tracking-opt-in/v5.2.0/tracking-opt-in.min.js';
```
nandy-andy, line 39: Did we miss one minor version and that's why we're now bumping from `v5.0.0` to `v5.2.0`?

dmnsobczak: It's not required to update it every time: we have an icVariable for that, second line is fallback config. Actually we can even remove it from here.

nandy-andy: Oh, now I see it in https://github.com/Wikia/ad-engine/pull/946/files#diff-a2da621c11d37cddd84428c6c71688b3f58b812c60771444f1d955754f6013e2L86 it's a fallback. All clear, thanks!

PR #951, opened 2020-12-23
ADEN-10737 add realvu support

KonradLinkowski:


platforms/ucp/ad-context.ts
```diff
@@ -121,6 +121,10 @@ export const basicContext = {
 			enabled: false,
 			appId: 'P26086A07-C7FB-4124-A679-8AC404198BA7',
 		},
+		realVu: {
+			id: 'c=E6H4_f=mint',
+			enabled: true,
```
dmnsobczak, line 126: Remember to set this to `false` and use icVariable instead during testing, because this will enable RealVu everywhere.

platforms/shared/setup/base-context.setup.ts
```diff
@@ -113,6 +113,10 @@ export class BaseContextSetup implements DiProcess {
 			this.instantConfig.get('icDurationMediaLibraryUrl'),
 		);
 		context.set('services.distroScale.enabled', this.instantConfig.get('icDistroScale'));
+		context.set(
+			'services.realVu.enabled',
+			this.instantConfig.get('icRealVu') && context.get('wiki.opts.enableRealVu'),
```
nandy-andy, line 118: Do we need to check `'wiki.opts.enableRealVu'` from context? It's inconsistent with the rest of our integrations but maybe we decided at some point we want this feature to be enabled by a community 🤔

src/ad-services/real-vu/index.ts
```diff
@@ -0,0 +1,25 @@
+import { context, utils } from '@ad-engine/core';
+
+const logGroup = 'RealVu';
+
+class RealVu {
+	call(): Promise<void> {
+		const id: string = context.get('services.realVu.id');
+
+		if (!context.get('services.realVu.enabled') || !id) {
+			utils.logger(logGroup, 'disabled');
```
nandy-andy, line 10: I do like the way you used logging. Using our query-string param will inform us about each state of the RealVu loading process 👍

platforms/shared/setup/base-context.setup.ts
nandy-andy: OK, I can see https://github.com/Wikia/unified-platform/pull/4393/files#diff-7d714ea1f52e468d81ef41f5d41b61c71ad9b962b6eeefd55eed0b9da1050969R172 so it means you decided to make it flexible (per platform/geo/community) 👌

nandy-andy (approved):

platforms/shared/setup/base-context.setup.ts
dmnsobczak: `AC: 1. Add the following RealVu tag on the pages of specific communities (communities TBD)`

platforms/ucp/ad-context.ts
```diff
@@ -121,6 +121,10 @@ export const basicContext = {
 			enabled: false,
 			appId: 'P26086A07-C7FB-4124-A679-8AC404198BA7',
 		},
+		realVu: {
```
dmnsobczak, line 124: This config value needs to be copied into `ucp-hydra`, `ucp-minerva` and `ucp-mobile` ad-contexts as well.

dmnsobczak (changes requested):
Add config to 3 remaining platforms

platforms/shared/setup/base-context.setup.ts
nandy-andy: I missed it 🙈 All clear, thanks! :)

nandy-andy (approved):

dmnsobczak (approved):

platforms/shared/setup/base-context.setup.ts
nandy-andy: But then... we don't need `icRealVu`, do we?

dmnsobczak: In case of any issues icVariable is faster than wgVariable. Also I think that in the future we'll drop the wgVariables and RealVu will be managed only using ICBM :)

PR #952, opened 2020-12-29
ADEN-10713 Adding MediaGrid bidder and JWPlayer RDP integration

dmnsobczak:


src/ad-bidders/prebid/index.ts
```diff
@@ -84,8 +84,24 @@ export class PrebidProvider extends BidderProvider {
...
+
+		this.registerBidsRefreshing();
+		this.registerBidsTracking();
+		this.getLiveRampUserIds();
+		this.enableATSAnalytics();
+
+		utils.logger(logGroup, 'prebid created', this.prebidConfig);
+	}
+
+	private configureLiveRamp() {
```
nandy-andy, line 100: @Loczi94 FYI :)

src/ad-bidders/prebid/index.ts
```diff
@@ -84,8 +84,24 @@ export class PrebidProvider extends BidderProvider {
...
 		this.prebidConfig = { ...this.prebidConfig, ...liveRamp.getConfig() };
+	}
 
+	private configureTCF() {
```
nandy-andy, line 104: What do you think of making the `configure*` methods cleaner? Instead of modifying the configuration property, we could pass it a configuration and they'll return a new configuration object without changing the input. This way we'll have a more stable code that isn't prone to side-effects.

dmnsobczak: Great idea, refactored :)

nandy-andy (approved):

PR #953, opened 2021-01-05
ADEN-10760 Fixing UCP-Mobile missing slots on FV pages

dmnsobczak:


platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -49,7 +51,12 @@ export class F2DynamicSlotsSetup implements DiProcess {
 			.subscribe();
 
 		if (!topLeaderboardDefinition) {
-			eventService.once(events.AD_STACK_START, () => btfBlockerService.finishFirstCall());
+			eventService.once(events.AD_STACK_START, () => {
+				btfBlockerService.finishFirstCall();
```
nandy-andy, line 55: Nice catch! Does it mean BTF blocker/UAP was broken also on f2?

nandy-andy (approved):

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
dmnsobczak: Yes, on video pages :<

PR #954, opened 2021-01-08
ADEN-10703 Remove orphaned SessionCookie items

dmnsobczak:


src/ad-engine/services/cookie-storage-adapter.ts
```diff
@@ -6,7 +6,7 @@ interface WikiaCookieAttributes extends Cookies.CookieAttributes {
 }
 
 export class CookieStorageAdapter implements StorageProvider {
-	private readonly cacheMaxAge = 30 * 60 * 1000;
+	private readonly cacheMaxAge = 24 * 60 * 60 * 1000;
```
nandy-andy, line 9: Where does this change come from? What's the reason of changing that?

src/ad-engine/services/session-cookie.ts
```diff
@@ -39,18 +39,33 @@ export class SessionCookie {
 	}
 
 	getItem<T>(key: string): T {
-		return this.storage.getItem<T>(`${this.prefix}_${key}`);
+		return this.storage.getItem<T>(this.getPrefixedKey(key));
 	}
 
 	setItem(key: string, input: {} | string): void {
-		this.storage.setItem(`${this.prefix}_${key}`, input);
+		this.removeOrphanedItems(key);
```
nandy-andy, line 46: Can we execute the new method somewhere else? What about the constructor? This change breaks the single responsibility principle. Most people will expect the `setItem()` method to just set the item not remove some items and set the item.

src/ad-engine/services/session-cookie.ts
```diff
@@ -39,18 +39,33 @@ export class SessionCookie {
...
 	clear(): void {
 		this.storage.clear();
 	}
+
+	private getPrefixedKey(key: string, prefix: string = this.prefix): string {
+		return `${prefix}_${key}`;
+	}
+
+	private removeOrphanedItems(key: string): void {
+		Object.keys(this.storage.getItem(''))
```
nandy-andy, line 63: Do I read it correctly passing an empty string to `getItem()` works like not existing `getItems()`? ;-)

dmnsobczak: Yes, it will run `Cookies.get(false-value)` and it will return all cookies available. I didn't want to add `getItems()` method to all storage providers because it would used only here so I used this simple "hack" or `js-cookie` library feature.

src/ad-engine/services/session-cookie.ts
dmnsobczak: I can't run it in the constructor, because I need `key` parameter. I can create a constant of possible keys in this class and iterate through it, but this also doesn't seem right. From my perspective and due to the fact that this is specialized `SessionCookie` class it is single responsibility to remove previous cookie while saving a new one - we want to keep only one, newest record. I could rename the method, but it needs to be in compliant with class it's implementing.
Another idea is to make `removeOrphanedItems` public and run next to the `setItem` call, but that will require other developers to remember to run it both and to know why they need to do that. I prefer to pass this duty to the class itself.
What you think? We can discuss it on Slack.

src/ad-engine/services/cookie-storage-adapter.ts
dmnsobczak: I'm almost sure that in previous, legacy version of `SessionCookie` feature the cookie was rewritten every pageview to reset the cookie TTL. Now it's not and it's weird for me, because if user will use the page longer than 30 minutes the cookie will disappear and Basset will assign the user to the A/B group again. We could rewrite the cookies again, but now I think it's not efficient. I decided to extend the TTL for the cookies to one day - after this time for the majority of traffic the session id should change, unless we don't remove the cookies manually before that.

dmnsobczak: Discussed on grooming - reverted.

fraszczakszymon (approved):

PR #963, opened 2021-02-05
ADEN-10814 Fixing bids refreshing + cleanup

dmnsobczak:


platforms/ucp-mobile/ad-context.ts
```diff
@@ -7,45 +7,18 @@ export const basicContext = {
...
-				top_boxad: {
-					sizes: [
-						[300, 250],
-						[300, 600],
-					],
-				},
-				featured: {
-					type: 'video',
-				},
+			bidsRefreshing: {
```
nandy-andy, line 14: Don't we want to keep here `slots` with the configuration for `top_leaderboard` and `bottom_leaderboard`?

dmnsobczak: These configs are incorrect, they are copied from desktop. The `slots` key is immediately overwritten by A9 UCP-Mobile config:
https://github.com/Wikia/ad-engine/blob/f11e82f90b5dc08ebba6a2dd30bf5bf619fc3a92/platforms/ucp-mobile/setup/context/a9/ucp-mobile-a9-config.setup.ts

nandy-andy (approved):

PR #967, opened 2021-02-08
ADEN-10808 remove LiveRamp integration

Tomasz-at-Fandom:


platforms/shared/tracking/tracking.setup.ts
```diff
@@ -248,12 +246,6 @@ export class TrackingSetup {
 		});
 	}
 
-	private liveRampTracker(): void {
-		communicationService.action$.pipe(ofType(liveRampPrebidIdsLoadedEvent)).subscribe((props) => {
-			this.pageTracker.trackProp('live_ramp_prebid_ids', props.userId);
-		});
-	}
-
 	private atsTracker(): void {
```
Loczi94, line 249: This is also LiveRamp-connected code thus it should be deleted.

Loczi94 (changes requested):
There is still ATS-related code in ad-engine. Also, I think we can update prebid.js here: https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/pbjs-factory.ts#L5 :)

platforms/shared/tracking/tracking.setup.ts
Tomasz-at-Fandom: Thanks! Done.

src/ad-engine/services/pbjs-factory.ts
```diff
@@ -2,7 +2,7 @@ import { logger, scriptLoader } from '../utils';
 import { context } from './context-service';
 
 const prebidLibraryUrl =
-	'//static.wikia.nocookie.net/fandom-ae-assets/prebid.js/v3.27.1/20201020.min.js';
+	'//static.wikia.nocookie.net/fandom-ae-assets/prebid.js/v4.17.0/20210217.min.js';
```
dmnsobczak, line 5: This line is a third fallback, we don't need to worry about that and actually we can clean this up, but thanks for remembering about that :)

src/ad-bidders/index.ts
```diff
@@ -155,4 +155,3 @@ export const bidders = new Bidders();
 export * from './tracking';
 export * from './wrappers';
 export * from './prebid/ats';
```
dmnsobczak, line 157: I think we can remove that too.

dmnsobczak (changes requested):
I think we can remove `/src/ad-bidders/prebid/ats` module too.

src/ad-bidders/index.ts
Tomasz-at-Fandom: As per our Zoom discussion, since this point to more ATS Analytics code, I will talk to Paulina to make sure we are not removing too much. To be continued...

Tomasz-at-Fandom: The Ats class was removed

src/ad-engine/services/pbjs-factory.ts
Tomasz-at-Fandom: Done

src/ad-engine/services/pbjs-factory.ts
```diff
@@ -15,7 +13,7 @@ class PbjsFactory {
 		if (!this.instancePromise) {
 			const libraryUrl = context.get('bidders.prebid.libraryUrl');
 
-			scriptLoader.loadScript(libraryUrl || prebidLibraryUrl, 'text/javascript', true, 'first');
```
Loczi94, line 18: I think we have Prebid.js url in code to prevent situation when Instant Config service does not respond. 🤔  But still, we have fallback-config.json where we store Prebid.js url. What do you think @dmnsobczak?

dmnsobczak: 1. Response from ICBM
2. fallback-config
3. This line
It can happen only for new platform which don't have fallback-config saved yet or variable was not enabled for this service in the past 24 hours. It's quite impossible to achieve and sounds like a small impact. I'm fine with removing it :D

Loczi94: Okay, thanks @dmnsobczak! I'm ok with it, too. It's one place less where we need to update the url.

Loczi94 (approved):

dmnsobczak (approved):

src/ad-engine/services/pbjs-factory.ts
Tomasz-at-Fandom: Ideally, from developers perspective, all configuration variables with fallback values should be managed / loaded in one place. If we have something like that then let's stick to it, to avoid duplication and for nice encapsulation of concerns :)

PR #969, opened 2021-02-09
ADEN-10831 Fixing one-pixel impact image visible issue

dmnsobczak:


platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -69,6 +69,7 @@ export class UapDomManager {
 
 	setResolvedImage(): void {
 		if (this.params.image2 && this.params.image2.background) {
+			this.manipulator.element(this.params.image1.element).addClass('hidden-state');
```
nandy-andy, line 72: interesting :) so, the thing showing up when the resolved state image (image2) height isn't round is the default state image? that should help 👍

nandy-andy (approved):

platforms/shared/templates/helpers/uap-dom-manager.ts
dmnsobczak: Yep, that's exactly the issue. We can round here, round there and check different edge cases or just hide it and done, since we don't use image transition anymore in Templates 2.0 :)

PR #971, opened 2021-02-10
ADEN-10812 Adding New Interstitial implementation

dmnsobczak:


platforms/shared/templates/handlers/close-to-redirect-button-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+				if (location.hash !== '#interstitial') {
+					transition('hidden');
+				}
+			},
+			false,
+		);
+	}
+
+	async onLeave(): Promise<void> {
+		this.button.remove();
```
Tomasz-at-Fandom, line 37: Mamy pewność że button nie będzie zdefiniowany?

Loczi94 (approved):

platforms/shared/templates/handlers/interstitial/interstitial-click-listener-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+		document.addEventListener('click', callback, false);
+	}
+
+	async onLeave(): Promise<void> {
+		this.adSlot.show();
+
+		window.location.hash = 'interstitial';
+
+		const cookieAdapter = new CookieStorageAdapter();
+		cookieAdapter.setItem('interstitial-impression', '1');
```
Tomasz-at-Fandom, line 39: Co oznacza ta jedyneczka? 
Czy za tym się kryje jakaś reguła określająca zachowywanie się reklamy? Że np ma się wyświetlić tylko raz czy coś takiego?
Może można by to ubrać w jakiegoś consta którego nazwa by coś więcej wyjaśniła?

platforms/shared/templates/handlers/close-to-redirect-button-handler.ts
Loczi94: `onEnter` is fired first and it defines `this.button` so it should be. 🤔

Tomasz-at-Fandom: I dug around a bit and I see that this is only used in the TemplateState class, cool.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -217,6 +218,36 @@ export class UcpMobileSlotsDefinitionRepository {
 		return this.instantConfig.get('icFloorAdhesion') && !context.get('custom.hasFeaturedVideo');
 	}
 
+	getInterstitialConfig(): SlotSetupDefinition {
+		if (!this.isInterstitialApplicable()) {
```
Tomasz-at-Fandom, line 222: Is this widely accepted in this project, that a method which is being called decides on whether or not it should have been called in the first place?

Tomasz-at-Fandom (approved):
Don't let my minor questions and suggestions stop you.

platforms/shared/templates/handlers/interstitial/interstitial-click-listener-handler.ts
dmnsobczak: `1` means anything. I just wanted to create a non-empty cookie. This will block `interstitial` from calling GAM for 30 minutes, which can be translated to one session. The real capping will be configured in GAM. If ad should get an impression once a day, the next GAM call will return `collapse` and cookie will be saved again without interstitial impression.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
dmnsobczak: Yes, exactly. New slot configuration mechanism needs some refactor, we will probably modernize it in the future, great you noticed that :)

PR #977, opened 2021-02-23
ADEN-10829 Adding IncontentPlayer placeholder

dmnsobczak:


platforms/ucp-mobile/styles.scss
```diff
@@ -105,6 +105,21 @@
 	}
 }
 
+#incontent_player {
+	display: block !important;
```
Loczi94, line 109: Is it really necessary? 🙈

platforms/ucp-mobile/styles.scss
```diff
@@ -105,6 +105,21 @@
...
+
+	&.ic-ad-slot-placeholder {
+		@include ic-ad-slot-placeholder;
+
+		min-height: calc((100vw - 2.25rem) / 1.3333);
```
Loczi94, line 114: How did you figure out that this is the exact min-height? Do wee need to use two different units here? And why `rem`?

platforms/ucp-mobile/styles.scss
dmnsobczak: Yes, because collapsed slot gets `hide` class and I need to overwrite that. But I put this in the wrong place :D Fixed!

platforms/ucp-mobile/styles.scss
dmnsobczak: 100% viewport width - 2.25rem (1.125 rem is a content padding) divided by 1.333 to get proper size for 4:3 aspect ratio :)

Loczi94: Thanks for clarification!

dmnsobczak: Oh, you're right. Changed :)

Loczi94 (approved):

PR #979, opened 2021-03-04
ADEN-10861 New interstitial skin v2.1

dmnsobczak:


platforms/shared/templates/handlers/interstitial/interstitial-bootstrap-handler.ts
```diff
@@ -20,6 +20,12 @@ export class InterstitialBootstrapHandler implements TemplateStateHandler {
 			await slotTweaker.adjustIframeByContentSize(this.adSlot);
 		}
 
+		if (window.location.hash === '#interstitial') {
+			window.location.hash = '';
+		}
+
+		window.ads.runtime.interstitial = window.ads.runtime.interstitial || {};
```
nandy-andy, line 27: What do you think of creating a new class which handles changes to this `window.ads.runtime.interstitial` state? I'm worried when I see a direct modification of global scope.

emadry (approved):
Went through CR already. Probably reviewer forgot to add checkmark

PR #981, opened 2021-03-05
ADEN-10799 Start using ad-tag-manager tags from inlined data instead of additional call to the service

nandy-andy:
More context in https://github.com/Wikia/unified-platform/pull/4860


platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -38,18 +38,26 @@ export class UcpTargetingSetup implements DiProcess {
...
+			pub: wiki.targeting?.adTagManagerTags?.pub || [],
 			s0: wiki.targeting.mappedVerticalName,
 			s0v: wiki.targeting.wikiVertical,
 			s0c: wiki.targeting.newWikiCategories,
 			s1: this.getRawDbName(wiki),
 			s2: this.getAdLayout(wiki.targeting),
+			sex: wiki.targeting?.adTagManagerTags?.sex || [],
 			skin: this.skin,
+			theme: wiki.targeting?.adTagManagerTags?.theme || [],
+			tv: wiki.targeting?.adTagManagerTags?.tv || [],
```
Loczi94, line 60: Correct me if I am wrong but I don't think we need `?` for all ad-tags key-vals since in case for example `adTagManagerTags` in `wiki.targeting?.adTagManagerTags?.age` is undefined then `[]` is assigned to `age`.

src/ad-services/permutive/index.ts
```diff
@@ -62,6 +62,8 @@ class Permutive {
 			const permutiveGptTargeting = window.googletag.pubads().getTargeting('permutive');
 
 			if (permutiveGptTargeting.length) {
+				utils.logger(logGroup, ['gets targeting from GPT', permutiveGptTargeting]);
```
Loczi94, line 65: Why `gets targeting from GPT` is in array? 🤔

nandy-andy: No specific reason. It's an old way (like AdEngine v2.0 old) of old AdEngers to push debugging info - I can change it ;)

emadry (approved):

PR #982, opened 2021-03-12
ADEN-10892 New interstitial skin v2.2

dmnsobczak:


platforms/shared/templates/handlers/interstitial/interstitial-bootstrap-handler.ts
```diff
@@ -20,14 +22,16 @@ export class InterstitialBootstrapHandler implements TemplateStateHandler {
...
-		transition('listen');
+		transition('display');
 	}
 
-	async onLeave(): Promise<void> {}
+	async onLeave(): Promise<void> {
+		eventService.emit(events.INTERSTITIAL_DISPLAYED);
+		this.adSlot.show();
+
+		window.ads.runtime.interstitial.visible = true;
```
Loczi94, line 35: Either I've missed something out or it is not needed.

platforms/shared/templates/handlers/interstitial/interstitial-bootstrap-handler.ts
```diff
@@ -20,14 +22,16 @@ export class InterstitialBootstrapHandler implements TemplateStateHandler {
 			await slotTweaker.adjustIframeByContentSize(this.adSlot);
 		}
 
-		if (window.location.hash === '#interstitial') {
-			window.location.hash = '';
-		}
-
 		window.ads.runtime.interstitial = window.ads.runtime.interstitial || {};
+		window.ads.runtime.interstitial.available = true;
```
Loczi94, line 26: Either I've missed something out or it is not needed.

src/ad-engine/services/slot-service.ts
```diff
@@ -93,6 +93,20 @@ class SlotService {
...
+			const found = id.match(matcher);
+
+			if (found['groups'] && found['groups'].slotName && this.slots[found['groups'].slotName]) {
+				const slotName = found['groups'].slotName;
+
+				this.slots[slotName].setConfigProperty('insertId', id);
+
+				return this.slots[slotName];
+			}
+		}
```
Loczi94, line 108: Imo at some time in the future one won't know that this parts was added because of interstitial so it would be useful to add a comment or make this code more self-explanatory.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -135,17 +134,40 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
+			'#dismiss-button > div',
+		);
+
+		header.style.backgroundColor = '#002a32';
+		button.style.border = '2px solid #00d6d6';
+		button.style.borderRadius = '24px';
+		button.style.width = '17px';
+		button.style.textAlign = 'center';
+		text.innerText = 'Advertisement';
+	}
```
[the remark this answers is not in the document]
Loczi94: Ok, I tried to refactor it and style using classes and now I understand why you did it this way.

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -183,6 +183,14 @@ export class GptProvider implements Provider {
 	/** @private */
 	createGptSlot(adSlot: AdSlot, sizeMap: GptSizeMap) {
 		if (adSlot.isOutOfPage()) {
+			if (adSlot.getConfigProperty('outOfPageFormat')) {
+				return window.googletag.defineOutOfPageSlot(
+					adSlot.getAdUnit(),
+					// @ts-ignore
```
Loczi94, line 189: I tried to update `@types/doubleclick-gpt` so we do not need to use `@ts-ignore` but it looks like even in the newest version property `enums` does not exist on type GoogleTag.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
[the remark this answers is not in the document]
dmnsobczak: That's right: we can't apply parent window CSS styles to iframes :< We could copy `<style>` snippet into an iframe, but it's the same hack :<

src/ad-engine/providers/gpt-provider.ts
dmnsobczak: Yeah, this library is not updated since 2019 :<

Tomasz-at-Fandom (approved):

PR #984, opened 2021-03-23
ADEN-10873 Create UCP-Desktop bundle based on UCP bundle

dmnsobczak:


platforms/ucp-desktop/ad-context.ts
```diff
@@ -0,0 +1,194 @@
+import fallbackInstantConfig from './fallback-config.json';
+
+export const basicContext = {
+	adUnitId:
```
nandy-andy, line 4: Shall we include here changes introduced in https://github.com/Wikia/ad-engine/pull/985 already or in a follow-up ticket?

Loczi94 (approved):

Mariusz8Walczak (approved):

PR #985, opened 2021-03-24
ADEN-10889 | Add original host name to AdUnitId 5th part on ucp-mobile

Tomasz-at-Fandom:


platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -47,6 +47,7 @@ export class UcpTargetingSetup implements DiProcess {
 			kid_wiki: wiki.targeting.directedAtChildren ? '1' : '0',
 			lang: wiki.targeting.wikiLanguage || 'unknown',
 			media: wiki.targeting?.adTagManagerTags?.media || [],
+			original_host: wiki.opts?.isGamepedia ? 'gamepedia' : 'fandom',
```
nandy-andy, line 50: I can't think of a better name 🤷‍♂️

nandy-andy (approved):

platforms/shared/context/targeting/ucp-targeting.setup.ts
nandy-andy: OK, let's leave it as is since it's unified with the name in GA: https://github.com/Wikia/unified-platform/pull/5052/files#diff-352293e35029c5ff7ee44c3210e1b50affaa6689de0cb0f074ddb8d14eb39666R339

PR #988, opened 2021-04-02
ADEN-10866 | ucp-mobile: Lazy requesting for ICBs

Loczi94:


platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -99,7 +100,23 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
 	}
 
 	private configureICBPlaceholderHandler(): void {
-		if (context.get('wiki.opts.enableICBPlaceholder')) {
+		if (context.get('wiki.opts.enableICLazyRequesting')) {
+			eventService.on(AdSlot.SLOT_RENDERED_EVENT, (adSlot) => {
+				const slotName = adSlot.getSlotName();
+				if (slotName.includes('incontent_boxad')) {
+					const slotElement = document.querySelector(`#${slotName}`);
+					slotElement.parentElement.classList.remove('loading');
```
nandy-andy, line 108: This will be needed to be changed to `slotElement.parentElement.classList.remove('is-loading');` after the `top_boxad` changes with cleanup kicks in. Let's keep it in mind and coordinate accordingly depending on which changes get merged first 👌

Loczi94: Ok, good point!

src/ad-engine/services/slot-placeholder-injector.ts
```diff
@@ -0,0 +1,82 @@
+import { getTopOffset, isInTheSameViewport, logger } from '../utils';
+
+const logGroup = 'slot-placeholder-injector';
+
+export interface SlotPlaceholderConfig {
+	classList: string[];
+	anchorSelector: string;
+	insertMethod: 'append' | 'prepend' | 'after' | 'before';
+	avoidConflictWith?: string[];
+	repeat?: number;
```
nandy-andy, line 10: This may be confusing that in the `SlotCreatorConfig` is [an object](https://github.com/Wikia/ad-engine/pull/988/files#diff-e1e9d500924692b52cba6ee6b57650e6b0798027970960c7ee2bca5e2e93e1ebR152-R162) and in `SlotPlaceholderConfig` it's a number. I don't know what to do with it. Just sharing my thoughts ;)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -130,6 +179,29 @@ export class UcpMobileSlotsDefinitionRepository {
 		return context.get('wiki.opts.pageType') !== 'search';
 	}
 
+	private injectIncontentAdsPlaceholders(): void {
+		const slotName = 'incontent ad';
```
nandy-andy, line 183: The value assigned to `slotName` looks odd.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -99,7 +100,23 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
+			});
+
+			eventService.on(AdSlot.HIDDEN_EVENT, (adSlot) => {
+				const slotName = adSlot.getSlotName();
+				if (slotName.includes('incontent_boxad')) {
+					const slotElement = document.querySelector(`#${slotName}`);
+					slotElement.parentElement.classList.remove('loading');
+				}
+			});
+		} else if (context.get('wiki.opts.enableICBPlaceholder')) {
```
nandy-andy, line 119: Don't we need here one more `else if` for `'wiki.opts.enableICPPlaceholder'` case?

src/ad-engine/services/slot-placeholder-injector.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+class SlotPlaceholderInjector {
+	inject(placeholderConfig: SlotPlaceholderConfig, adSlotName: string): void {
+		let repeat = 2;
```
nandy-andy, line 15: Where does the magical `2` come from?

src/ad-engine/services/slot-placeholder-injector.ts
Loczi94: I agree. I renamed it to `repeatLimit`.

src/ad-engine/services/slot-placeholder-injector.ts
Loczi94: If we repeat a placeholder it means that it is a second placeholder in the row. Do you have any ideas how to refactor this part?

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
Loczi94: We have this part in separate method - `configureICPPlaceholderHandler`.

src/ad-engine/services/slot-placeholder-injector.ts
```diff
@@ -0,0 +1,82 @@
...
+
+		conflictingElementsSelectors.forEach((selector) => {
+			const selected: HTMLElement[] = Array.from(document.querySelectorAll(selector));
+
+			elements.push(...selected);
+		});
+
+		return elements;
+	}
+}
```
nandy-andy, line 80: These methods are similar but much simpler than https://github.com/Wikia/ad-engine/pull/988/files#diff-23599229e919b398dd805ce5482ab0504f06c3e54ea1eb97d2bb00928e552839L60-L93 I wonder if we can make a reusable abstraction for it in the future. Nothing to address now, just to keep in mind when thinking about potential future dev improvements :)

src/ad-engine/services/slot-placeholder-injector.ts
nandy-andy: Not right now. My brain does not work anymore ;) It starts thinking about getting the start number from somewhere else where it's already there (for example the index from the slot name `incontent_boxad_2`) but it's overcomplicating, isn't it? Right now just a comment why it's 2 will do, I guess 🤷‍♂️

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -130,6 +179,29 @@ export class UcpMobileSlotsDefinitionRepository {
...
+			insertMethod: 'before',
+			avoidConflictWith: ['.ad-slot', '.ic-ad-slot-placeholder', '.ad-slot-wrapper'],
+			repeat: 19,
+		};
+
+		communicationService.action$.pipe(ofType(uapLoadStatus), take(1)).subscribe((action) => {
+			if (!action.isLoaded) {
+				slotPlaceholderInjector.inject(icbPlaceholderConfig, slotName);
+
+				context.set('slots.incontent_boxad_1.insertBeforeSelector', '');
```
nandy-andy, line 196: What do you think of moving this logic to the `SlotPlaceholderInjector`? I assume it'll make `injectIncontentAdsPlaceholders()` more readable.

Loczi94: This part is strictly connected with ICBs logic. I'd rather keep these two things separate:

- SlotPlaceholderInjector class with code which injects placeholders (in the future it can be used for injecting not only incontent ads placeholders)
- UcpMobileSlotsDefinitionRepository class with all 'configuration' code like setting context.

Both - injecting placeholders and injecting ICB1 wait for the same UAP action thus moving this part to SlotPlaceholderInjector won't change an order in which the code is run.

Tomasz-at-Fandom:
Was there any change done to the content of all the modified .json files or were they just reformatted?

Loczi94:
@Tomasz-at-Fandom there wasn't any change. Please, just ignore them.

Tomasz-at-Fandom (changes requested):

Tomasz-at-Fandom (changes requested):

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -99,27 +103,70 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 			]);
-
-			eventService.on(AdSlot.SLOT_RENDERED_EVENT, (adSlot) => {
-				adSlot.removeClass('loading');
-			});
+			communicationService.action$
+				.pipe(
+					ofType(adSlotEvent),
+					filter((action) => shouldRemoveICBLoader(action)),
+				)
```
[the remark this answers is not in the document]
Loczi94: I can see your point but my point was not to listen to and pipe actions if it is not necessary. 🤔  That is why I placed this code within 'if' blocks.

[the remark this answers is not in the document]
Tomasz-at-Fandom: A yes I see, well we could just make another `if` block checking if the subscription function was defined and if so then make the subscription :)

Tomasz-at-Fandom (changes requested):

Tomasz-at-Fandom (approved):

PR #989, opened 2021-04-08
ADEN-10868 removed one style rule

Mariusz8Walczak:


dmnsobczak (approved):
Removal looks fine and I don't really remember why we added this rule. Let's emphasize in QA Notes to test it against regression with Amazon A9 ads.

Loczi94 (approved):

PR #1005, opened 2021-04-23
ADEN-10866 | Refactor SlotPlaceholderInjector + unit tests

Loczi94:


src/ad-engine/services/slot-placeholder-injector.ts
```diff
@@ -90,6 +109,10 @@ class SlotPlaceholderInjector {
 
 		return elements;
 	}
+
+	private getLastPlaceholderNumber(repeat: number): number {
```
nandy-andy, line 113: Do we really need this method? It drove my attention because the name suggested it was getting the placeholder number from somewhere. And it's not getting anything from anywhere ;) After taking a closer look I'd just subtracting `1` from `repeat` in lines 34 and 42 :D Am I missing something? Maybe it tries to describe what the subtraction means 🤔

Loczi94: My goal was to create a method with descriptive name so I don't have to add comments in code. 🙈

dmnsobczak (approved):

PR #1007, opened 2021-04-29
ADEN-10948 Adding feature to disable floor_adhesion pushOnScroll

dmnsobczak:


platforms/shared/setup/base-context.setup.ts
```diff
@@ -74,6 +74,10 @@ export class BaseContextSetup implements DiProcess {
 			this.instantConfig.get('icFeaturedVideoPostroll'),
 		);
 
+		context.set(
+			'options.floorAdhesionNumberOfViewportsFromTopToPush',
+			this.instantConfig.get('icFloorAdhesionViewportsToStart'),
+		);
```
Loczi94, line 80: `options.floorAdhesionNumberOfViewportsFromTopToPush` is being read only in ucp platform. In ucp-mobile we read it directly from ICBM in `ucp-mobile-slots-definition-repository.ts`. Imo we should either read `options.floorAdhesionNumberOfViewportsFromTopToPush` or get `icFloorAdhesionViewportsToStart` directly form ICB in both platforms.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -306,25 +306,25 @@ export class UcpMobileSlotsDefinitionRepository {
...
 			activator: () => {
-				const numberOfViewportsFromTopToPush: number =
-					this.instantConfig.get('icFloorAdhesionViewportsToStart') || 0;
-
-				context.set('slots.floor_adhesion.disabled', !this.instantConfig.get('icFloorAdhesion'));
-				context.set(
-					'slots.floor_adhesion.numberOfViewportsFromTopToPush',
-					this.instantConfig.get('icFloorAdhesionViewportsToStart'),
-				);
+				context.set(`slots.${slotName}.disabled`, !this.instantConfig.get('icFloorAdhesion'));
```
Loczi94, line 309: Imo it'd be better to leave `floor_adhesion` instead of using template string. This way, when we look at this line of code, we know (right away) what value is set in context.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -306,25 +306,25 @@ export class UcpMobileSlotsDefinitionRepository {
...
 				context.set(
-					'slots.floor_adhesion.forceSafeFrame',
+					`slots.${slotName}.forceSafeFrame`,
```
Loczi94, line 311: ditto

platforms/shared/setup/base-context.setup.ts
dmnsobczak: I'm aware of that but I did it on purpose. We have a ticket to replace old slot injecting mechanism to SlotCreator:
https://wikia-inc.atlassian.net/browse/ADEN-10381
SlotCreator has ICBM on a list of dependencies. We will unify that while removing old logic. 
Second thing is that FloorAdhesion is currently disabled on Desktop, so this is a low priority change/unification request.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
dmnsobczak: You right, reverted.

Loczi94 (approved):

PR #1008, opened 2021-04-29
ADEN-10952 Fixing JWPlayer types

dmnsobczak:


src/ad-products/video/jwplayer/jwplayer-actions.ts
```diff
@@ -3,13 +3,15 @@ import { props } from 'ts-action';
 
 export interface VideoTargeting {
 	plist?: string;
-	videoTags?: string | string[]; // not sure about `string`
-	v1?: string; // playlist item id
+	vtags?: string;
```
nandy-andy, line 6: If anyone wonders why, here it is: https://github.com/Wikia/unified-platform/pull/5701/files#diff-a79c014f148717d562146a8d5953e7cba8e26a9bffe4adf4cda6ae3348b7aedaR46-R47 :)

src/ad-products/video/jwplayer/jwplayer-actions.ts
```diff
@@ -3,13 +3,15 @@ import { props } from 'ts-action';
...
 }
 
 export interface JwPlayerAdsFactoryOptions {
-	slotName: string;
+	audio: boolean;
+	ctp: boolean;
 	featured: boolean;
```
nandy-andy, line 12: Where do we set `featured`? It's missing here: https://github.com/Wikia/unified-platform/pull/5701/files#diff-a79c014f148717d562146a8d5953e7cba8e26a9bffe4adf4cda6ae3348b7aedaR49-R54

dmnsobczak: I thought that it can be used on other platform, JWPlayer related code gives a TypeScript error after removing it. I did not checked and I didn't wanted to modify that code, because that's expanding the scope of the ticket :(

nandy-andy (approved):

src/ad-products/video/jwplayer/jwplayer-actions.ts
nandy-andy: Yeah, it's a P2, a regression. Let's fix it and follow-up on what uses this flag somewhere else 👍

PR #1018, opened 2021-05-14
ADEN-10692 add isMobile targeting parameter

jakubbanas:


platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -62,6 +62,7 @@ export class UcpTargetingSetup implements DiProcess {
 			uap: 'none',
 			uap_c: 'none',
 			wpage: wiki.targeting.pageName && wiki.targeting.pageName.toLowerCase(),
+			is_mobile: utils.client.isMobileSkin(this.skin) ? '1' : '0',
```
nandy-andy, line 65: Is there a particular reason for returning strings here and in f2 whereas for sports we're good with numbers (integers)?

nandy-andy: I think we can make it consistent and return ints everywhere. However, at the end GAM will probably change it to strings, anyway 🤷‍♂️ Yet, we expect to end up with ints ;) https://github.com/Wikia/dfp-creative-templates/pull/129/files#diff-fe8ea114341e4fa86fa2aefba8c026ddb7c5b0677bffccedbf0116f1af3c4c64R184

jakubbanas: Good point, I changed it to int everywhere 👍

nandy-andy (approved):

platforms/shared/context/targeting/ucp-targeting.setup.ts
jakubbanas: It turned out that it has to be string, so I made it string everywhere.
When there was int, the template was rendered with an empty string.

Tomasz-at-Fandom (approved):

PR #1019, opened 2021-05-18
ADEN-10943 - integrate ad-engine with SilverSurfer

preludek:


dmnsobczak:
Missing label. Is `Minor` ok?

src/ad-services/silver-surfer/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { AdTags } from '../taxonomy/taxonomy-service.loader';
+import { silverSurferServiceLoader } from './silver-surfer.loader';
+
+const logGroup = 'silver-surfer';
+
+class SilverSurferService {
+	async configureUserTargeting(): Promise<AdTags> {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+			return {};
```
dmnsobczak, line 11: Inhibitors has a type of `Promise<any>[]` and they are used to call `Promise.all(this.inhibitors)`. It's not an error to return non-Promise value, because it will be skipped by `Promise.all`, but consider rewriting it to return `Promise.resolve()` once disabled.

src/ad-services/silver-surfer/index.ts
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		const targetingConfig = context.get('services.silverSurfer');
+		const userProfile = await silverSurferServiceLoader.getUserProfile();
```
dmnsobczak, line 15: I think it's worth try-catching that part as it's an external library.

src/ad-services/taxonomy/taxonomy-service.ts
```diff
@@ -11,7 +11,7 @@ export class TaxonomyService {
 
 		context.set('targeting.txn', '-1');
 
-		const adTags: AdTags = await taxonomyServiceLoader.getAdTags();
+		const adTags = await taxonomyServiceLoader.getAdTags();
```
dmnsobczak, line 14: Why?

dmnsobczak (approved):

src/ad-services/taxonomy/taxonomy-service.ts
preludek: I was looking at taxonomy integration while doing this task and tried to be a good scout 🙈  this type is infered by TS, we don't need that.

Loczi94: Yep, we don't need that but on the other hand, when we look at thus code it is clear right away what type we expect from taxonomyServiceLoader.getAdTags().

dmnsobczak (approved):

PR #1024, opened 2021-05-26
ADEN-11011 - call for silver-surfer service for lazy loaded slots

preludek:


src/ad-engine/models/ad-slot.ts
```diff
@@ -82,6 +82,7 @@ export const adSlotEvent = action(
 export class AdSlot extends EventEmitter {
 	static CUSTOM_EVENT = 'customEvent';
 	static PROPERTY_CHANGED_EVENT = 'propertyChanged';
+	static SLOT_ADDED = 'slotAdded';
```
dmnsobczak, line 85: Rename to `SLOT_ADDED_EVENT` to meet convention?

src/ad-engine/services/slot-service.ts
```diff
@@ -70,6 +70,10 @@ class SlotService {
 
 		slotTweaker.addDefaultClasses(adSlot);
 		eventService.emit(events.AD_SLOT_CREATED, adSlot);
+		utils.communicator('Ad Slot added', {
+			name: slotName,
+			state: AdSlot.SLOT_ADDED,
```
dmnsobczak, line 75: Actually why don't you use existing `AD_SLOT_CREATED` event?

src/ad-services/silver-surfer/index.ts
```diff
@@ -1,8 +1,14 @@
+import { communicationService, globalAction, ofType } from '@ad-engine/communication';
 import { context, utils } from '@ad-engine/core';
+import { props } from 'ts-action';
 import { AdTags } from '../taxonomy/taxonomy-service.loader';
 import { silverSurferServiceLoader } from './silver-surfer.loader';
 
 const logGroup = 'silver-surfer';
+const adSlotLoadedEvent = globalAction(
```
dmnsobczak, line 8: Can we reuse this chunk and create `getCommunicatorAction()` method? Or maybe it's an overkill?
https://github.com/Wikia/ad-engine/blob/b531e87ef923a0dd19c0afacf544ffff7c2cba60/src/ad-engine/utils/logger.ts#L41

src/ad-engine/services/slot-service.ts
preludek: Was thinking about it, but It is from a deprecated class and it's also a `symbol` type, not a `string` so I figured I'll do it this way.

src/ad-engine/models/ad-slot.ts
preludek: Fixed

PR #1025, opened 2021-05-26
ADEN-10794 Adding Stroer integration

Mariusz8Walczak:


src/ad-services/stroer/index.ts
```diff
@@ -0,0 +1,23 @@
...
+			utils.logger(logGroup, 'disabled');
+
+			return Promise.resolve();
+		}
+
+		const libraryUrl = `//js.adscale.de/map.js`;
+
+		utils.logger(logGroup, 'loading', libraryUrl);
+
+		return utils.scriptLoader.loadScript(libraryUrl, 'text/javascript', true).then(() => {
```
preludek, line 17: You don't need to pass `'text/javascript'` and `true` here. Those are default values in scriptLoader.loadScript method.

src/ad-services/stroer/index.ts
```diff
@@ -0,0 +1,23 @@
...
+			utils.logger(logGroup, 'ready');
+		});
+	}
+}
+
+export const stroer = new Stroer();
```
preludek, line 23: This class is not used anywhere. You need to call it in on platforms in callExternals, fe:
https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts#L69

preludek (changes requested):

Loczi94 (dismissed):
Label major/minor/patch is missing.

What is more, unit tests are missing. We have unit tests for all services in `ad-engine/spec/ad-services.`

src/ad-services/stroer/index.ts
Mariusz8Walczak: thanks

src/ad-services/stroer/index.ts
Mariusz8Walczak: thanks, done

Mariusz8Walczak:
There are no test, becouse when ic flag is on, script will be loaded, now is false, I think makeing test that are depended on something external for very basic code, that was used many times is mistake.

Loczi94:
In tests we don't read icVar from Instant Config service. We set it to a specific value in ad-engine context. I know it is a dummy test but better have one dummy test than none.

platforms/shared/setup/base-context.setup.ts
```diff
@@ -133,6 +133,7 @@ export class BaseContextSetup implements DiProcess {
 			'services.realVu.enabled',
 			this.instantConfig.get('icRealVu') && context.get('wiki.opts.enableRealVu'),
 		);
+		context.set('services.stroer.enabled', this.instantConfig.get('icStroer'));
```
dmnsobczak, line 136: Can we move it to L144 to keep alphabetical order?

platforms/ucp-mobile/modes/ucp-mobile-ads.mode.ts
```diff
@@ -51,6 +52,7 @@ export class UcpMobileAdsMode implements DiProcess {
 		iasPublisherOptimization.call();
 		confiant.call();
 		realVu.call();
+		stroer.call();
```
dmnsobczak, line 55: Missing this set for `platforms/ucp/modes/ucp-ads-mode.ts`

src/ad-services/stroer/index.ts
```diff
@@ -0,0 +1,23 @@
+import { context, utils } from '@ad-engine/core';
+
+const logGroup = 'Stroer';
```
dmnsobczak, line 3: Let's keep (sadly, not always followed :<) conversion to use lowercase + dash for log groups :)

dmnsobczak (changes requested):
Missing UCP (Oasis) implementation.

src/ad-services/stroer/index.ts
Mariusz8Walczak: ok

platforms/ucp-mobile/modes/ucp-mobile-ads.mode.ts
Mariusz8Walczak: Ok, I will add

src/ad-services/stroer/index.ts
Mariusz8Walczak: added lowercase

platforms/shared/setup/base-context.setup.ts
Mariusz8Walczak: L143 is done

dmnsobczak (approved):

preludek (approved):

src/ad-services/index.ts
```diff
@@ -10,5 +10,6 @@ export * from './instant-config';
 export * from './nielsen';
 export * from './permutive';
 export * from './real-vu';
+export * from './stroer';
```
dmnsobczak, line 13: `ST` should be after `SI` :D

PR #1006, opened 2021-04-27
ADEN-10890 Send RealVu detection result with bidders calls

dmnsobczak:

PR #1026, opened 2021-05-27
ADEN-11007 | feat: move TOC under BFAA on start and complete state

Tomasz-at-Fandom:


src/ad-products/video/porvata4/porvata-player.ts
```diff
@@ -77,6 +77,22 @@ export class PorvataPlayer {
...
+			this.moveTocUnderBfaa(Math.round(this.settings.getHeight())),
+		);
+	}
+
+	moveTocUnderBfaa(height: number): void {
+		const bfaa = document.querySelector<HTMLElement>('.bfaa-sticky-big');
+		const toc = document.getElementById('toc');
+		if (bfaa !== null && toc !== null && height !== undefined) {
+			toc.style.setProperty('--sticky-height', `${height.toString()}px`);
+		}
```
Loczi94, line 95: What about creating a separate handler and adding it to stickyBig and stickySmall states here: https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp-mobile/templates/bfaa-new-template.ts?

Loczi94 (changes requested):

src/ad-products/video/porvata4/porvata-player.ts
```diff
@@ -77,6 +77,20 @@ export class PorvataPlayer {
 		this.addEventListener('wikiaAdCompleted', () =>
 			this.dom.getInterfaceContainer().classList.add('hide'),
 		);
+
+		this.addEventListener(google.ima.AdEvent.Type.STARTED, (_) =>
+			utils.communicator('Porvata Started', {
+				name: 'Porvata',
```
dmnsobczak, line 83: In RealVu pull request I added a feature to skip second argument:
https://github.com/Wikia/ad-engine/pull/1006/files#diff-50e4c1249aa3095f147eea2fd8ee626809020070724d90d931b08ce57724467cR39
You can copy those changes and remove object passed as second parameter since you're ignoring `props` on UCP side anyway 🤔

dmnsobczak (approved):

PR #1033, opened 2021-06-09
ADEN-11044 Move UAP on FandomDesktop to the right of global nav

dmnsobczak:


platforms/ucp-desktop/templates/configs/register-ucp-desktop-uap-dom-elements.ts
```diff
@@ -3,7 +3,7 @@ import { TemplateDependency } from '@wikia/ad-engine';
 
 export function registerUcpDesktopUapDomElements(): TemplateDependency[] {
 	return [
-		{ bind: NAVBAR, value: document.querySelector('.global-navigation') },
+		{ bind: NAVBAR, value: document.querySelector('.fandom-sticky-header') },
```
preludek, line 6: Did you check how if there are changes in BFAB? It seems this function is also used in `registerBfabTemplate`.

platforms/shared/styles/uap/_bfaa-slot.scss
```diff
@@ -1,12 +1,12 @@
 @import 'src/ad-products/styles/mixin';
 
-@mixin bfaa-slot($uap-z-index: 202) {
+@mixin bfaa-slot($uap-z-index: 202, $uap-margin: 0px) {
```
preludek, line 3: I'm thinking `uap-margin` => `navbar-size`?  it's pretty much navbar-size right?

preludek (approved):

platforms/shared/styles/uap/_bfaa-slot.scss
preludek: or `global-navigation-width` :) We would know what is happening here just looking at it.

dmnsobczak: I wanted to make it more generic. It's basically how much less wide than 100% UAP is, it might be because of navbar or any other element on any platform.

platforms/ucp-desktop/templates/configs/register-ucp-desktop-uap-dom-elements.ts
dmnsobczak: Yes, it works fine. NAVBAR is ignored for BFAB and at the same time BFAB is inside `.main-container` element which is already smaller and BFAB is not overlapping navbar.

preludek: Sweet 💎

platforms/ucp-desktop/styles.scss
```diff
@@ -16,10 +16,18 @@
...
+			margin-top: -46px;
+		}
+
 		#top_leaderboard,
 		#hivi_leaderboard {
-			@include bfaa-slot(1000);
-			@include sticky-tlb-slot();
+			$global-navigation-width: 66px;
+
+			@include bfaa-slot(600, $global-navigation-width);
```
latata, line 27: changing z-index to 401-499 should resolve most of the issues

latata (commented):
You would also need to override z-index of .fandom-sticky-header to 401-499 to prevent covering the notifications box for logged in users

PR #1042, opened 2021-06-14
ADEN-10944 Fixing navbar on News&Stories when menu is open

dmnsobczak:


platforms/f2/styles.scss
```diff
@@ -71,6 +71,10 @@
 			@include uap-navbar();
 		}
 
+		&.is-menu-active .global-navigation-mobile-wrapper {
+			top: 0 !important;
```
nandy-andy, line 75: 🙈

dmnsobczak: `top: xxx px` is added inline by AdEngine to this element, so only `!important` is stronger than that :<

nandy-andy: That's sad. On the other hand, it's only when UAP is on the page. Yet, I'm still afraid it'll bite us in the future. Maybe with the new UAP 3.0 we'll be able to do it less hacky 🤞

preludek (approved):

PR #1044, opened 2021-06-16
ADEN-11073 change UAP z-index to fix navigation viewability

preludek:


platforms/shared/styles/uap/_bfab-slot.scss
```diff
@@ -1,7 +1,7 @@
 @import 'src/ad-products/styles/mixin';
 
 @mixin bfab-slot() {
-	z-index: 201;
```
Loczi94, line 4: I would not change that here since these styles apply to UAP BFAB on all platforms. What about setting z-index for BFAB slot here: https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp-mobile/styles.scss#L61? We can even create a new mixin dedicated for UAP 3.0 BFAB.

Loczi94 (changes requested):

platforms/shared/styles/uap/_bfab-slot.scss
preludek: New mixin is a great idea given the potential regression! 👍

PR #1046, opened 2021-06-18
ADEN-11079 | ucp-mobile: UAP 3.0 fix bug on infobox pages + fix transitions between embedded and sticky states

Loczi94:


platforms/shared/styles/uap/_bfaa-embedded-small.scss
```diff
@@ -1,5 +1,7 @@
 @mixin bfaa-embedded-small() {
 	&.bfaa-embedded-small {
+		position: relative;
+
 		#top_leaderboard {
 			.video-overlay {
 				position: absolute;
```
Mariusz8Walczak, line 7: why we don't use common style rule, for embedded-big and small? they have very much in common

platforms/shared/styles/uap/_bfaa-sticky-small.scss
```diff
@@ -1,5 +1,8 @@
 @mixin bfaa-sticky-small() {
 	&.bfaa-sticky-small {
+		position: fixed;
```
Mariusz8Walczak, line 3: why we don't use common style rule, for embedded-big and small? they have very much in common

Mariusz8Walczak (commented):
I found that sticky small and sticky big are almoust the same, can we just make some common rules in css for both of them and TS classes. It will bite us, when something will change. We need only one place to make changes, not few ones.

platforms/shared/styles/index.scss
```diff
@@ -7,6 +7,7 @@
 @import 'uap/bfaa-slot';
 @import 'uap/bfaa-sticky-big';
 @import 'uap/bfaa-sticky-small';
+@import 'uap/_bfaa-with-infobox';
```
preludek, line 10: underscore 😈

platforms/shared/styles/uap/_bfaa-with-infobox.scss
```diff
@@ -0,0 +1,14 @@
...
+	.portable-infobox-wrapper ~ .top-leaderboard.ad-slot-placeholder {
+		left: -12px;
+
+		#top_leaderboard {
+			margin: 0;
+		}
+
+		&.bfaa-sticky-big,
+		&.bfaa-sticky-small {
+			left: 0px;
```
preludek, line 11: left: 0

platforms/shared/styles/uap/_bfaa-sticky-big.scss
```diff
@@ -1,5 +1,8 @@
 @mixin bfaa-sticky-big() {
 	&.bfaa-sticky-big {
+		position: fixed;
+		top: 55px;
```
preludek, line 4: Since this is coming back, maybe we should move it to some sort of a variable so we know that this is navbar height?

platforms/shared/templates/handlers/slot/slot-decision-embedded-small-to-sticky-small-handler.ts
```diff
@@ -3,22 +3,27 @@ import { Injectable } from '@wikia/dependency-injection';
...
 				startWith({}),
-				filter(
-					() =>
-						window.scrollY >= this.manager.getAdSlotTopOffset() + this.manager.getNavbarHeight(),
-				),
-				tap(() => transition('stickySmall')),
+				filter(() => this.reader.getAdSlotTopOffset() <= this.manager.getNavbarHeight()),
+				tap(() => {
+					this.manager.setAdSlotTopOffset(window.scrollY);
+					transition('stickySmall');
```
preludek, line 25: I'm still not sure either this should be returned or not (or does it even matter 🙈 ).

platforms/shared/styles/uap/_bfaa-embedded-small.scss
preludek: Also wondering if this can be somehow reused 🤔

preludek (changes requested):

platforms/shared/templates/handlers/slot/slot-decision-embedded-small-to-sticky-small-handler.ts
Loczi94: Wait, what should be returned (or not) and why? 😄

platforms/shared/styles/uap/_bfaa-sticky-small.scss
Loczi94: I replaced mixins `bfaa-embedded-big`, `bfaa-embedded-small` and `bfaa-embedded-resolved` by one mixin - `bfaa-embedded`.

Loczi94:
@Mariusz8Walczak @preludek I replaced mixins `bfaa-embedded-big`, `bfaa-embedded-small` and `bfaa-embedded-resolved` by one mixin - `bfaa-embedded` because all three above were the same. However, I'd rather leave separate styles for `bfaa-sticky-big` and `bfaa-sticky-small`. I know, they have common parts but imo it's more clear if we have separate mixins and separate CSS classes for these two states instead of having some common styles in one mixin and then other styles for example in separate classes. What do you think?

preludek (approved):

Tomasz-at-Fandom (approved):

PR #1050, opened 2021-06-23
ADEN-11083 Fixing BLB position

dmnsobczak:


Tomasz-at-Fandom (approved):

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -43,22 +43,17 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
 	private injectSlots(): void {
 		const slots: Dictionary<SlotConfig> = context.get('slots');
 		Object.keys(slots).forEach((slotName) => {
-			if (slots[slotName].insertBeforeSelector) {
+			if (
+				slotName !== 'incontent_boxad_1' &&
+				(slots[slotName].insertBeforeSelector || slots[slotName].parentContainerSelector)
```
Tomasz-at-Fandom, line 48: This is getting hard to read. 

When the slot contains the `parentContainerSelector` property, shouldn't we also check if the slot definition contains the `insertIntoParentContainerMethod` property?

I'm not completely sure what is the purpose of this conditional statement. I understand the first part that the slot cannot be `incontent_boxad_1`. But the second condition is a bit confusing. The slot must contain either a selector to insert the slot before it (on the same level of nesting) or a selector for an element inside which the slot will be nested? 

So we are basically validating the slot definition to check if it contains information on how to insert the slot. So for the validation to be complete we should also check if the `insertIntoParentContainerMethod` exists in the case of a slot that is to be nested inside `parentContainerSelector`.

If I'm correct, can we extract this second condition into a private method, like `insertionMethodValidation()`?

Tomasz-at-Fandom (changes requested):

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
dmnsobczak: > I understand the first part that the slot cannot be incontent_boxad_1. But the second condition is a bit confusing. The slot must contain either a selector to insert the slot before it (on the same level of nesting) or a selector for an element inside which the slot will be nested?
That's right. 

> we should also check if the insertIntoParentContainerMethod exists in the case of a slot that is to be nested inside parentContainerSelector
No, this option have a default values defined in `slotInjector`.

> can we extract this second condition into a private method
Ideally would be to extract this to `SlotInjector` class, but since we still have old style slot injecting mechanism used on legacy platforms it's better not to add anything there before we do some cleanup first.

Tomasz-at-Fandom: OK, thanks for explaining.

Tomasz-at-Fandom (approved):

PR #1054, opened 2021-06-24
[ADEN-11045] ucp-mobile: UAP 3.0 change placeholder size in embeddedR…

Loczi94:
…esolved state


platforms/shared/templates/helpers/uap-dom-reader.ts
```diff
@@ -5,12 +5,22 @@ import { NAVBAR } from '../configs/uap-dom-elements';
 
 @Injectable({ autobind: false })
 export class UapDomReader {
+	private adSlotHeight = 0;
```
preludek, line 8: Is there a reason why we won't move the whole logic of returning 0 or this.adSlot.element.offsetHeight in getAdSlotHeight():?

Loczi94: Good point. 👍

preludek (approved):

PR #1057, opened 2021-06-29
ADEN-11064 Adding Instant Search Ads implementation

dmnsobczak:


src/ad-services/ad-marketplace/index.ts
```diff
@@ -0,0 +1,171 @@
...
+		url: string,
+		image: string,
+		impression: string,
+	): string {
+		return (
+			`<a href="${url}">` +
+			`<img class="logo" src="${image}" alt="" />` +
+			`<span class="title">${title}</span><span class="label">Sponsored</span>` +
+			`<img class="pixel" src="${impression}" width="1" height="1" />` +
+			'</a>'
```
jakubbanas, line 166: IMO, the multiline string would be more readable here.
```
return(
`<a href="${url}">
    <img class="logo" src="${image}" alt="" />
    <span class="title">${title}</span><span class="label">Sponsored</span>
    <img class="pixel" src="${impression}" width="1" height="1" />
  </a>`
);
```

jakubbanas (approved):

src/ad-services/ad-marketplace/index.ts
dmnsobczak: You right! Changed.

PR #1058, opened 2021-06-30
ADEN-11088 fixed viewability for UAP

preludek:


preludek:
Changes here also affect the old UAP (transition state). This was working correctly imo only because old UAP is always in the viewport at first (either top of page or immediately get's sticked).

Also I removed `shouldStickUntillSlotViewed` since in design it should always stay sticky until viewability is counted.

platforms/shared/templates/handlers/slot/slot-decision-embedded-big-to-embedded-resolved-handler.ts
```diff
@@ -39,13 +42,17 @@ export class SlotDecisionEmbeddedBigToEmbeddedResolvedHandler implements Templat
 	private async staticTransitionHandler(
 		transition: TemplateTransition<'embeddedResolved'>,
 	): Promise<void> {
-		this.timeout
-			.isViewedAndDelayed()
+		merge(this.getViewabilityStream())
```
Loczi94, line 45: I don't think we need `merge` here.

platforms/shared/templates/handlers/slot/slot-decision-embedded-big-to-embedded-resolved-handler.ts
```diff
@@ -39,13 +42,17 @@ export class SlotDecisionEmbeddedBigToEmbeddedResolvedHandler implements Templat
...
 			.pipe(
-				filter((viewedAndDelayed) => !this.isVideo && viewedAndDelayed),
+				filter(() => !this.isVideo),
```
Loczi94, line 47: I don't like the idea of using private variable to switch between these two streams but I couldn't figure out how to implement it using only RxJs operators.

preludek: Yeah, me neither 🙈  couldn't come up with anything better 🤷

platforms/shared/templates/handlers/slot/slot-decision-embedded-big-to-embedded-resolved-handler.ts
preludek: Yup, fixed!

Loczi94 (approved):

PR #1060, opened 2021-07-02
[ADEN-11092] ucp-mobile UAP 3.0 - remove transition state

Loczi94:


preludek:
slot-transition-new-template-handler.ts is not used anywhere now, let's 🔪 it

preludek (changes requested):
slot-transition-new-template-handler.ts is not used anywhere now, let's 🔪 it

preludek (approved):

PR #1061, opened 2021-07-05
ADEN-11084 make adSlotEvent a global action

preludek:


src/ad-engine/models/ad-slot.ts
```diff
@@ -69,8 +69,7 @@ export interface WinningBidderDetails {
 	price: number | string;
 }
 
-// TODO: This should be split into separate action for each event
```
preludek, line 72: This is Bielik approved.

dmnsobczak (approved):

src/ad-engine/models/ad-slot.ts
Bielik20: -_-

PR #1064, opened 2021-07-07
ADEN-11117 Changing video ad units for ucp_mobile platform

dmnsobczak:


platforms/ucp-mobile/ad-context.ts
```diff
@@ -29,11 +29,11 @@ export const basicContext = {
 		},
 	},
 	custom: {
+		dbNameForAdUnit: '_not_a_top1k_wiki',
```
nandy-andy, line 32: It has the same value as `custom.wikiIdentifier`. Why do we need both?

dmnsobczak: It is being overwritten depending if wiki is top1k or not:
https://github.com/Wikia/ad-engine/blob/71fee8c43ad618b356da72564f246ea0fc8c1e24/platforms/shared/context/targeting/ucp-targeting.setup.ts#L27

nandy-andy (approved):

PR #1065, opened 2021-07-07
ADEN-11066 | ucp-desktop: fix injecting incontent boxad ads

Loczi94:


platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -43,10 +43,7 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
 	private injectSlots(): void {
 		const slots: Dictionary<SlotConfig> = context.get('slots');
 		Object.keys(slots).forEach((slotName) => {
-			if (
-				slotName !== 'incontent_boxad_1' &&
-				(slots[slotName].insertBeforeSelector || slots[slotName].parentContainerSelector)
-			) {
+			if (slots[slotName].insertBeforeSelector || slots[slotName].parentContainerSelector) {
```
nandy-andy, line 46: Was it the root cause? The first element in `slots` was `incontent_boxad_1` and it didn't start the rest of the logic because of that? 🤔 The rest of the changes seems rather cosmetic. But then... do we need the lines below: https://github.com/Wikia/ad-engine/pull/1065/files#diff-559d1d8926e7b8a77ce95977e3e1928acfbcd8c66ff3b54e983778c812407e7eR51 and the whole logic its connected with? 🤔

Loczi94: The root cause was that `incontent_boxad_1` was not pushed to `events.pushOnScroll.ids` in context. Let mi double check if this part with `appendIncontentBoxad` is needed.

nandy-andy: Right, and the line within this if is doing that - pushing the `incontent_boxad_1` to the `events.pushOnScroll.ids` 👌 https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/slot-injector.ts#L69 Thanks for helping me understand this 🙇

Loczi94: So yes - this part is needed since this is the actual part of code which injects ICB1. ICB1 is not injected by `inject` method in `SlotInjector` class since it uses `insertBeforeSelector` from config which for `ICB1 is `#incontent_boxad_1`. That's why we can see `There is not enough space for incontent_boxad_1` in logs on Oasis and FandomDesktop. I don't like how it works because it's confusing and imo it should be refactored to either inject ICB1 by `SlotInjector.inject()` or to have separate logic for ICBs and exclude injecting ICB1 from `SlotInjector`.

nandy-andy (approved):
We both agreed the way it's coded isn't perfect but we want the functionality to work properly on the growing traffic of FandomDesktop. We'll discuss possibilities of refactoring it at our internal technical refinements.

PR #1072, opened 2021-07-20
ADEN-11133 Adding FandomMobile ads labels

dmnsobczak:
Ticket: https://wikia-inc.atlassian.net/browse/ADEN-11133


platforms/ucp-mobile/styles.scss
```diff
@@ -254,6 +254,33 @@ not(.ic-ad-slot-placeholder) {
...
+	position: relative;
+	padding: 18px 0 24px 0;
+
+	&::after {
+		position: absolute;
+		left: 0;
+		bottom: 0;
+		width: 100%;
+		color: #5f7a7b;
+		content: 'Advertisement';
```
nandy-andy, line 270: What about translations? 🙈

nandy-andy: We used to had it translated in the legacy app: https://github.com/Wikia/app/blob/dev/extensions/wikia/AdEngine3/AdEngine3.i18n.php Let's discuss it today at daily.

dmnsobczak: We do not translate it on Fandom.com 🤔 Let's discuss!

nandy-andy (approved):

PR #1075, opened 2021-07-22
ADEN-11045 | feat: UAP embedded placeholder size always big

Tomasz-at-Fandom:


platforms/ucp-mobile/styles.scss
```diff
@@ -41,21 +41,20 @@
 			position: sticky;
 			width: 100vw;
 			z-index: $new-uap-bfaa-z-index;
+			min-height: 211px;
```
dmnsobczak, line 44: What is this value? UAP can have different ratios and different sizes. It probably will work for most of use cases, but it's not ideal and can cause layout shift for some campaigns. Maybe let's introduce some additional padding and position UAP in the center?

platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -15,8 +15,8 @@ export class UapDomManager {
 		private reader: UapDomReader,
 	) {}
 
-	addClassToAdSlotPlaceholder(className: string): void {
-		this.manipulator.element(this.adSlot.element.parentElement).addClass(className);
+	addClassToAdSlot(className: string): void {
```
dmnsobczak, line 18: This is a breaking change and it conflict with the fix I prepared here:
https://github.com/Wikia/ad-engine/pull/1071/files#diff-e78632dbc387b1dfd9e869ee83eedecb275dd1dbde54932d78d8582c3910de0eR4
Let's talk and figure out merge conflicts :)

dmnsobczak (changes requested):
Changes will conflict with changes/fixes in different ticket. This one will probably go second so let's discuss how to solve them.

platforms/shared/templates/helpers/uap-dom-manager.ts
Tomasz-at-Fandom: We can have both of these functions if they are needed, but we should review their uses together.

platforms/ucp-mobile/styles.scss
dmnsobczak: @sseyda What's the biggest (most square) UAP aspect ratio we're using?

dmnsobczak (approved):

platforms/ucp-mobile/styles.scss
sseyda: @dmnsobczak as per slack, UAPs are using following sizes: 1600x400 for Desktop, 640x360 for mobile for impacts and 1600x160 and 640x213 for resolved state.

dmnsobczak: Thank you! @Tomasz-at-Fandom consider using `min-height: 213px;` here then to remove that micro content shift :)

Tomasz-at-Fandom: Thanks guys! I bumped it up to 213px 👍

PR #1077, opened 2021-07-23
ADEN-11139 Pass "uap" keyval to Featured Video calls

dmnsobczak:


jakubbanas (approved):

platforms/ucp-mobile/setup/context/slots/ucp-mobile-slots-context.setup.ts
```diff
@@ -272,7 +271,6 @@ export class UcpMobileSlotsContextSetup implements DiProcess {
 			video: {
```
nandy-andy, line 271: What's the difference between `featured` and `video`? Why for FandomDesktop we don't have `video` defined? 🤔

dmnsobczak: `video` AFAIR is used here: https://www.fandom.com/video
This is a garbage, probably came here during one of many migrations. I removed it from UCP-Mobile.

nandy-andy (approved):

preludek (approved):

PR #1080, opened 2021-07-28
ADEN-11149 Adjust F2 bundle for new page design

dmnsobczak:


platforms/f2/setup/context/slots/f2-slots-context.setup.ts
```diff
@@ -160,53 +138,28 @@ export class F2SlotsContextSetup implements DiProcess {
...
-					[300, 600],
-				],
+				defaultSizes: [],
 				targeting: {
 					loc: 'footer',
 					uap: 'none',
 					rv: 1,
 				},
 			},
-			featured: {
```
nandy-andy, line 190: Don't we need to keep it, so the video ads on pages with JWPlayer will be displayed?

platforms/f2/setup/dynamic-slots/f2-slots-definition-repository.ts
```diff
@@ -53,99 +45,26 @@ export class F2SlotsDefinitionRepository {
...
 
-		if (!this.isArticle && this.f2Env.hasRightRail) {
+		if (!this.isArticle) {
 			return {
 				activator,
 				slotCreatorWrapperConfig,
 				slotCreatorConfig: {
 					slotName,
-					anchorSelector: '.feed-layout__right-rail',
+					anchorSelector: '.feed-section__ad',
```
nandy-andy, line 61: That's a bit confusing. I wonder if we can just change the element's class in `f2` to `feed-section__top_boxad` 🤔 Then we'll do the change in this line and in the f2 repo: https://github.com/Wikia/f2/pull/2434/files#diff-78543651ce4a822efd853e9d235dfe0432b3aafba514c860a5d03a44a9762d15R14 It will be less confusing from the AdEng point of view.

platforms/f2/setup/context/slots/f2-slots-context.setup.ts
dmnsobczak: WOW, I removed too much :D Reverted 🙇

platforms/f2/templates/configs/register-f2-uap-dom-elements.ts
```diff
@@ -14,7 +14,7 @@ export function registerF2UapDomElements(): TemplateDependency[] {
 
 				return f2Env.isPageMobile
 					? document.querySelector('.global-navigation-mobile-wrapper')
-					: document.querySelector('.wds-global-navigation-wrapper');
+					: document.querySelector('.nav-bg-hack');
```
nandy-andy, line 17: @mklucsarits let us know please if you decide to change the class name :)

dmnsobczak: I needed the first class on the page with zero height (or some height if this class will be sticky) :)

nandy-andy (approved):

platforms/f2/setup/dynamic-slots/f2-slots-definition-repository.ts
mklucsarits: Yeah please change that class to whatever you like. I introduced the `feed-section__ad` class but it wasn't intended to be permanent.

platforms/f2/templates/configs/register-f2-uap-dom-elements.ts
mklucsarits: That should be fine, but also if you need to add or change elements or rename/add CSS classes just for ad placement, please feel free to do so.

mklucsarits:
What's the status of this PR? Does it need to be tested?

nandy-andy:
> What's the status of this PR? Does it need to be tested?

We've just finished testing it and merged today.

PR #1083, opened 2021-08-02
ADEN-11174 Fix ads placeholders + Blockthrough

dmnsobczak:
Ticket: https://wikia-inc.atlassian.net/browse/ADEN-11174

### DIFF
--- a/src/ad-engine/services/bab-detection.ts
+++ b/src/ad-engine/services/bab-detection.ts
@@ -30,6 +30,7 @@ class BabDetection {
 
 		utils.logger(logGroup, 'BAB detection, AB detected:', isBabDetected);
 
+		this.setBodyClass(isBabDetected);
 		this.setRuntimeParams(isBabDetected);
 		this.updateSrcParameter(isBabDetected);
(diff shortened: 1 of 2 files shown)


nandy-andy (approved):

src/ad-engine/services/bab-detection.ts
```diff
@@ -100,6 +101,12 @@ class BabDetection {
 		// Post-QueCast
 		communicationService.dispatch(babDetectedEvent({ detected: isBabDetected }));
 	}
+
+	private setBodyClass(isBabDetected: boolean): void {
+		if (isBabDetected) {
```
jakubbanas, line 106: IMO the `isBabDetected` name is misleading because in fact the AdBlocker is detected, not BlockAdBlock.
Maybe `isAdblockDetected` would be more descriptive.

dmnsobczak: When we were writing it we wanted to avoid using in code phrases like `adblock`, `blockadblock` etc.

jakubbanas (approved):

PR #1087, opened 2021-08-06
ADEN-11141 Implementing UAP 4.0

dmnsobczak:


src/ad-engine/utils/try-property.ts
```diff
@@ -1,6 +1,10 @@
 import { Dictionary } from '../models';
 
 export function whichProperty(obj: Dictionary = {}, properties: string[] = []): string | undefined {
+	if (!obj) {
```
Tomasz-at-Fandom, line 4: This smells hacky :) How is this related to any other changes in this PR?

dmnsobczak: Discussed on Slack, removing for now :)

Tomasz-at-Fandom (approved):

PR #1094, opened 2021-08-24
ADEN-11170 Implementing UAP:JWPlayer feature

dmnsobczak:
Ticket: https://wikia-inc.atlassian.net/browse/ADEN-11170


src/ad-products/video/jwplayer/helpers/jwplayer-inhibitor.ts
```diff
@@ -0,0 +1,40 @@
+import { context, utils } from '@ad-engine/core';
+import { universalAdPackage } from '../../../templates/uap/universal-ad-package';
```
preludek, line 2: I think this import can be shortened to `import { universalAdPackage } from '../../../templates';`

src/ad-products/video/jwplayer/helpers/jwplayer-inhibitor.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	get(): Promise<void> {
+		if (!this.isEnabled()) {
+			return Promise.resolve();
+		}
+
+		return this.getExtendedPromise();
+	}
+
+	resolve(lineItemId: string = null, creativeId: string = null): void {
```
preludek, line 19: Is TypeScript ok with assigning null to string? :D

preludek (approved):

src/ad-products/video/jwplayer/helpers/jwplayer-inhibitor.ts
dmnsobczak: Actually `undefined` is also false, so I removed default values :D

src/ad-products/video/jwplayer/helpers/jwplayer-inhibitor.ts
dmnsobczak: Done!

PR #1102, opened 2021-09-10
ADEN-11216 'x' button on mobile adhesion should always be on top

Tomasz-at-Fandom:


jakubbanas (approved):

platforms/shared/index.ts
```diff
@@ -33,7 +33,7 @@ export * from './templates/interstitial-template';
 export * from './templates/configs/uap-dom-elements';
 export * from './templates/handlers/advertisement-label-handler';
 export * from './templates/handlers/close-to-embedded-resolved-button-handler';
-export * from './templates/handlers/close-to-hidden-ihi-button-handler';
+export * from './templates/handlers/interstitial/interstitial-close-to-hidden-button-handler';
```
dmnsobczak, line 36: `Legacy Interstitial` is a dead template and it will/should be removed in the future. There's no need to create a directory for it's handlers :D

platforms/shared/templates/handlers/floor-adhesion/floor-adhesion-anchor-hidden-handler.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Injectable } from '@wikia/dependency-injection';
+import { DomManipulator } from '../../helpers/manipulators/dom-manipulator';
+
+@Injectable({ autobind: false })
+export class FloorAdhesionAnchorHiddenHandler implements TemplateStateHandler {
+	constructor(private manipulator: DomManipulator) {}
+
+	async onEnter(): Promise<void> {
+		this.manipulator
+			.element(document.getElementById('floor_adhesion_anchor'))
```
dmnsobczak, line 11: Why you added FloorAdhesion anchor anyway? It's making things a bit inconsistent, because there are more out-of-page slots and only floor is anchored.

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
```diff
@@ -10,15 +10,12 @@ import {
...
+export class InterstitialCloseToHiddenButtonHandler implements TemplateStateHandler {
 	private button: HTMLButtonElement;
 	private readonly showCloseButtonAfter?: number;
 
 	constructor(@Inject(TEMPLATE.SLOT) private adSlot: AdSlot, instantConfig: InstantConfigService) {
-		this.showCloseButtonAfter =
-			this.adSlot.getSlotName() === 'floor_adhesion'
-				? instantConfig.get('icFloorAdhesionTimeToCloseButton', 0)
-				: instantConfig.get('icInvisibleHighImpact2TimeToCloseButton', 0);
+		this.showCloseButtonAfter = instantConfig.get('icInvisibleHighImpact2TimeToCloseButton', 2);
```
dmnsobczak, line 18: Why default value is set to `2`? It's undefined in ICBM now.

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
```diff
@@ -10,15 +10,12 @@ import {
...
 
 @Injectable({ autobind: false })
-export class CloseToHiddenIhiButtonHandler implements TemplateStateHandler {
+export class InterstitialCloseToHiddenButtonHandler implements TemplateStateHandler {
 	private button: HTMLButtonElement;
 	private readonly showCloseButtonAfter?: number;
 
 	constructor(@Inject(TEMPLATE.SLOT) private adSlot: AdSlot, instantConfig: InstantConfigService) {
-		this.showCloseButtonAfter =
-			this.adSlot.getSlotName() === 'floor_adhesion'
```
dmnsobczak, line 19: I think this is not right. FloorAdhesion in this context is a template, not a slot and this if is checking which slot is loading this template. Currently it can be:
* FloorAdhesion template loaded by floor_adhesion slot (with default FloorAdhesion template setting)
* FloorAdhesion template loaded by invisible_high_impact_2 (loaded by `top.loadCustomAd`)
After this Option 2 will use `icFloorAdhesionTimeToCloseButton` for IHI loaded Floors instead of `icInvisibleHighImpact2TimeToCloseButton`.

dmnsobczak: My suggestion: if we want to refactor it, let's ask Seba and then:
* remove Legacy Interstitial template
* decide to not load FloorAdhesion through IHI2
* Enable FloorAdhesion globally (or communicate that it needs to be enabled if Floor should be available)
* remove legacy ICBM variables

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
Tomasz-at-Fandom: Ups, good catch, I was just playing around with that and forgot to reset to 0.

platforms/shared/index.ts
Tomasz-at-Fandom: Just tidying up, the folder was there already, I just moved this handler there because nothing else is using it.
Unfortunately there is nothing in the code to inform us that this is legacy or deprecated :/

platforms/shared/templates/handlers/floor-adhesion/floor-adhesion-anchor-hidden-handler.ts
Tomasz-at-Fandom: Because there is a need for the slot itself to be empty, but I wanted the X button and the slot to be somehow encapsulated together and not wander around the top level of the body like a couple of strangers ;) Other slots should also get their anchors like we agreed with SiteX.

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
Tomasz-at-Fandom: > I think this is not right. FloorAdhesion in this context is a template, not a slot and this if is checking which slot is loading this template. Currently it can be:
> 
> * FloorAdhesion template loaded by floor_adhesion slot (with default FloorAdhesion template setting)
> * FloorAdhesion template loaded by invisible_high_impact_2 (loaded by `top.loadCustomAd`)
> 
> After this Option 2 will use `icFloorAdhesionTimeToCloseButton` for IHI loaded Floors instead of `icInvisibleHighImpact2TimeToCloseButton`.

Why do we confuse ourselves like that? Why not just load template A with slot A? Is there a reason for matching different templates with different slots?



> My suggestion: if we want to refactor it, let's ask Seba and then:
> 
> * remove Legacy Interstitial template
> * decide to not load FloorAdhesion through IHI2
> * enable FloorAdhesion globally (or communicate that it needs to be enabled if Floor should be available)
> * remove legacy ICBM variables

- Yes
- Yes
- Yes
- Yes

dmnsobczak: > Why do we confuse ourselves like that? Why not just load template A with slot A? Is there a reason for matching different templates with different slots?

In the past it was only IHI2 and it was loading Interstitials, Floor Adhesions, Popups, custom scripts and many other stuff. Floor Adhesion slot was a first step to tidy this up (aka removing IHI2). There was also an advantage of moving Floor Adhesion to the named slot: we could wrap AdX and bidders with a Floor Adhesion template.

> 4x Yes
💪

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
```diff
@@ -10,15 +10,12 @@ import {
...
-				? instantConfig.get('icFloorAdhesionTimeToCloseButton', 0)
-				: instantConfig.get('icInvisibleHighImpact2TimeToCloseButton', 0);
+		this.showCloseButtonAfter = instantConfig.get('icInvisibleHighImpact2TimeToCloseButton', 0);
```
nandy-andy, line 18: Shouldn't we use here `icFloorAdhesionTimeToCloseButton` isntead of `icInvisibleHighImpact2TimeToCloseButton`?

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
Tomasz-at-Fandom: That was before Sebastian said that "indirects" might be using it for something if I understood him right. So we should make sure it's safe, perhaps it would be better to do it in a separate ticket?

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
Tomasz-at-Fandom: That would couple both solutions to the same variable. My understanding of this code was that we want to be able to set different time delay for the floor adhesion and for the interstitial.

platforms/shared/templates/handlers/interstitial/interstitial-close-to-hidden-button-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+		this.button = new CloseButton({
+			onClick: () => {
+				this.adSlot.emitEvent(SlotTweaker.SLOT_CLOSE_IMMEDIATELY);
+				transition('hidden');
+			},
+		}).render();
+
+		setTimeout(() => {
+			this.adSlot.getElement().appendChild(this.button);
+		}, 0);
```
nandy-andy, line 27: NOOOOOOO! ;)

Tomasz-at-Fandom: Yea I know, but that's how it's been done in the original class. I don't want to refactor too much at once 🤷‍♂️

preludek (approved):

platforms/ucp-mobile/ucp-mobile-platform.ts
```diff
@@ -26,7 +26,7 @@ import { UcpMobileBaseContextSetup } from './setup/context/base/ucp-mobile-base-
 import { UcpMobilePrebidConfigSetup } from './setup/context/prebid/ucp-mobile-prebid-config.setup';
 import { UcpMobileSlotsContextSetup } from './setup/context/slots/ucp-mobile-slots-context.setup';
 import { UcpMobileDynamicSlotsSetup } from './setup/dynamic-slots/ucp-mobile-dynamic-slots.setup';
-import { UcpMobileTemplatesSetup } from './templates/ucp-mobile-templates-setup.service';
+import { UcpMobileTemplatesSetup } from './templates/ucp-mobile-templates-setup';
```
dmnsobczak, line 29: Why not `./templates/ucp-mobile-templates.setup`? :( Quick fix? :D

dmnsobczak (approved):

platforms/ucp-mobile/ucp-mobile-platform.ts
Tomasz-at-Fandom: OK but this Pull request was already closed so here is a new one
https://github.com/Wikia/ad-engine/pull/1116/files

PR #1108, opened 2021-09-20
ADEN-11234 Force TLB to be sticky

szymon-nawrocki:
I've added support for displaying top_leaderboard as sticky to maximize viewability. 
For more information see ticket ADEN-11234.

When configuration in icbm variable 'icForceStickyTlb' is met, top_leaderboard gets defaultTemplate='stickyTlb' and appears as sticky


platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -33,6 +33,10 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
 	private isLineAndGeo(): boolean {
 		const lines: string[] = context.get('templates.stickyTlb.lineItemIds') || [];
 
+		if (context.get('templates.stickyTlb.forced')) {
```
preludek, line 36: I would move that to a separate function f.e. isStickyForced or something like that. It should not be a part of isLineAndGeo imo.

preludek (changes requested):
There are changes in package-lock.json that shouldn't be here.

platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
szymon-nawrocki: @preludek thanks for your input

### package-lock.json
Changes were caused by installing node dependencies with use of 'npm install' command instead of 'npm ci'. 
I installed dependencies once again with 'npm ci' command, but 'git status' did not show any changes, so i manually deleted the line of code which was added before.

### sticky-tlb-blocking-handler.ts 
I moved the functionality to separate function stickyTlbForced() as you suggested.

platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -33,6 +33,10 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
 	private isLineAndGeo(): boolean {
 		const lines: string[] = context.get('templates.stickyTlb.lineItemIds') || [];
 
+		if (this.stickyTlbForced()) {
```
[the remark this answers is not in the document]
szymon-nawrocki: Okay, I've done the changes

preludek (approved):

preludek (approved):

PR #1111, opened 2021-09-21
ADEN-11248 Removing old ICBM variables

dmnsobczak:


platforms/muthead/ad-context.ts
```diff
@@ -94,6 +94,7 @@ export const basicContext = {
 	services: {
 		durationMedia: {
 			enabled: false,
+			libraryUrl: '//tag.durationmedia.net/sites/10655/dm.js',
```
nandy-andy, line 97: I wonder if we have automated tests checking the numbers for different platforms. It's a silly thing to make a mistake with: 10654, 10655 and probably couple more for FandomDesktop/FandomMobile etc. (//cc @soniackozubek not urgent 😊 )

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -215,13 +211,6 @@ export class GptProvider implements Provider {
 		});
 	}
 
-	forceSafeFrame(gptSlot: googletag.Slot) {
```
nandy-andy, line 218: Goodbye, good friend 😭 I hope we'll meet again one day...

nandy-andy (approved):

PR #1117, opened 2021-10-04
ADEN-11185 Replace placeholder animation

szymon-nawrocki:
## Links
Ticket: https://fandom.atlassian.net/browse/ADEN-11185
Changes in UCP: https://github.com/Wikia/unified-platform/pull/7485

## Description
Basically removing: _is-loading_ animation on ads that are being loaded (to not distract the user and draw more attention to the area than is prefer).


platforms/ucp-mobile/styles.scss
```diff
@@ -120,12 +120,13 @@
 	align-items: center;
 	display: flex;
 	justify-content: center;
+	margin: 2px auto;
 	// 250px (slot height) + 18px (top padding) + 24px (bottom padding = label height)
 	min-height: 292px;
```
nandy-andy, line 125: Is the 2px margin necessary? Because I think you'll need to update min-height in the line #125 if you add margin or verify if there is no 4px shift after the ad is loaded (and placeholder disappears).

platforms/ucp-mobile/styles.scss
```diff
@@ -120,12 +120,13 @@
 	align-items: center;
 	display: flex;
 	justify-content: center;
+	margin: 2px auto;
```
nandy-andy, line 123: Is the 2px margin necessary? Because I think you'll need to update min-height in the line 125 if you add margin or verify if there is no 4px shift after the ad is loaded (and placeholder disappears).

szymon-nawrocki: Tbh I took the inspiration for `margin: 2px auto;` from UCP FandomMobile styles:
https://github.com/Wikia/unified-platform/blob/e09d505bdd4a55df4d90af3d0f4d69d13df12997/skins/FandomMobile/styles/components/ads.scss#L13

And I've added this 2px to make the site consistent, so without bigger or lower gaps between ad and text between the ad slots, and those 2px are margin for "ic-ad-slot-placeholder" NOT for ".is-loading" class so I think it shouldn't shift the content when placeholder dissapears.

But please correct me if I am wrong.

platforms/ucp-mobile/styles.scss
```diff
@@ -120,12 +120,13 @@
...
 	// 250px (slot height) + 18px (top padding) + 24px (bottom padding = label height)
 	min-height: 292px;
 
 	&.is-loading {
-		animation: is-loading 2s infinite;
-		border-radius: 20px;
+		background-color: #d8d8d8;
```
jakubbanas, line 128: I believe you could reuse a variable declared in UCP here to avoid repetition.

But I'm not sure if it's a good idea to use it cross-repo. Just a thought to consider.

jakubbanas (approved):

PR #1118, opened 2021-10-05
ADEN-11258 add clilck position tracking

jakubbanas:


src/ad-tracking/click-tracker.ts
```diff
@@ -55,19 +56,32 @@ class AdClickTracker {
...
+		slot: AdSlot,
+		event?: MouseEvent,
+	): void {
+		const data = {
+			ad_status: AdSlot.STATUS_CLICKED,
+		};
+		if (event) {
+			const clickData = {
+				click: { x: event.x, y: event.y },
+				// @ts-ignore
```
jakubbanas, line 76: I put `ts-ignore` here because it complained that `target` doesn't have `offsetWidth` property, but it does 🤷‍♂️

src/ad-tracking/click-tracker.ts
```diff
@@ -55,19 +56,32 @@ class AdClickTracker {
...
+				size: { x: event.target.offsetWidth, y: event.target.offsetHeight },
+			};
+			data['unused_6'] = JSON.stringify(clickData);
```
jakubbanas, line 79: I'm not sure if I shouldn't have named the column properly instead of `unused_6`, but it's just temporary data we are about to drop after the UAP4 experiment.
What do you think?

src/ad-tracking/click-tracker.ts
```diff
@@ -55,19 +56,32 @@ class AdClickTracker {
...
+		event?: MouseEvent,
+	): void {
+		const data = {
+			ad_status: AdSlot.STATUS_CLICKED,
+		};
+		if (event) {
+			const clickData = {
+				click: { x: event.x, y: event.y },
+				// @ts-ignore
+				size: { x: event.target.offsetWidth, y: event.target.offsetHeight },
```
preludek, line 77: We have `slot_size` sent already so I think this is unnecessary.

src/ad-tracking/click-tracker.ts
preludek: I'm not sure as well 🙈  but I feel that this data may be useful along the line. Maybe we should ask DE to expand the table? I'll ping other's for opinion on that :)

preludek (changes requested):

src/ad-tracking/click-tracker.ts
jakubbanas: Right, I'll remove it. Thanks!

src/ad-tracking/click-tracker.ts
Tomasz-at-Fandom: Don't overthink it, just give it a name that means something :)

src/ad-tracking/click-tracker.ts
Tomasz-at-Fandom: Not according to the docs
https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.eventtarget.html

You are declaring event to be MouseEvent and it contains `target` param that is implementing the EventTarget interface and that interface does not require `offsetWidth` as far as I can see. Are we appending this somewhere?

preludek (approved):

PR #1121, opened 2021-10-08
ADEN-11074 Adding targeting bundles functionality

dmnsobczak:


spec/ad-engine/utils/targeting.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { expect } from 'chai';
+import { createSandbox } from 'sinon';
+import { context } from '../../../src/ad-engine';
```
preludek, line 3: This import can be shortened to `'@wikia/ad-engine';`, below one can be shortened as well.

spec/ad-engine/utils/targeting.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	it('getTargetingBundles returns empty array for wrong entry data', () => {
+		expect(
+			targeting.getTargetingBundles({
+				bundle1: {
+					// @ts-ignore
+					s1: '_project43',
+					// @ts-ignore
+					esrb: true,
+				},
```
preludek, line 84: instead of ts ignores you can go with `as any` with this object and TS should be happy :P

PR #1163, opened 2021-11-02
ADEN-11320 Nativo ad in Fan Feed on FandomMobile

nandy-andy:


platforms/ucp-mobile/styles.scss
```diff
@@ -199,6 +199,10 @@ not(.ic-ad-slot-placeholder) {
 	@include nativo;
 }
 
+#ntv_feed_ad {
```
preludek, line 202: We should style everything by classes and not id's. Having both used here and there is making maintenance and potential changes way harder.

src/ad-engine/services/slot-creator.ts
```diff
@@ -58,6 +63,8 @@ export class SlotCreator {
 			this.throwNoPlaceToInsertError(slotConfig.slotName);
 		}
 
+		logger(groupName, 'SlotCreator::getAnchorElement() called', slotConfig, result);
```
preludek, line 66: There is group name used so `getAnchorElement() called` seems sufficient.

preludek (changes requested):

platforms/ucp-mobile/styles.scss
nandy-andy: I agree. I'd like to address it in the follow-up work as I'm thinking of removing `.gpt-ad` class from Nativo slot that are adding the style I'm removing with this CSS rule. Once `.gpt-ad` is not added to Nativo slots we can remove lines 202-204 from this file 👌

src/ad-services/nativo/nativo.service.ts
```diff
@@ -0,0 +1,97 @@
...
+				utils.logger(logGroup, 'ready');
+				this.sendEvent();
+			});
+	}
+
+	start(): void {}
+
+	requestAd(): void {
+		if (utils.queryString.get('native_ads_test') === '1') {
+			utils.logger(logGroup, 'Displaying an dummy test ads');
```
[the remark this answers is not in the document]
nandy-andy: Trolling is a art 😈

src/ad-services/nativo/index.ts
```diff
@@ -25,25 +34,68 @@ class Nativo {
...
+	private nativeFanFeedAdTestDummy(): void {
+		const nativeAdFeedPlaceholder = document.getElementById(Nativo.FEED_AD_SLOT_NAME);
+
+		if (nativeAdFeedPlaceholder) {
+			nativeAdFeedPlaceholder.innerHTML = `<a href="https://bit.ly/NewFanLabRecirc" class="recirculation-prefooter__item-anchor recirculation-prefooter__sponsored mcf-card mcf-card-article ntv-ad-anchor"  style="background-image: linear-gradient(to bottom, rgba(0, 0, 0, 0), #000000), url(https://placekitten.com/386/259);">
+				<p class="recirculation-prefooter__personalization wds-font-size-s wds-font-weight-medium wds-leading-tight mcf-card-article__subtitle ntv-ad-label">AD · best buy</p>
+				<p class="recirculation-prefooter__title wds-font-size-base wds-font-weight-bold wds-leading-tight mcf-card-article__title ntv-ad-title">Buy Opla: Very good Astra 1.0 TDI</p>
+				<p class="recirculation-prefooter__subtitle wds-font-size-s wds-font-weight-medium wds-leading-tight mcf-card-article__subtitle ntv-ad-offer">Available from komis for $213.7</p>
+			</a>`;
+		} else {
```
Tomasz-at-Fandom, line 89: If we would reverse the statement, move logger before the HTML block and make an early return then the else would be obsolete  and the HTML block would have one less indentation :)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -76,7 +74,7 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
 	}
 
 	private injectNativeAdsPlaceholder(): void {
-		if (!isNativeAdApplicable()) {
+		if (!nativo.isEnabled()) {
 			return;
 		}
```
Tomasz-at-Fandom, line 80: Could we refactor contents of the following callback? 
There are two`if` statements there that cause deep indentations. We could reverse them and make early returns instead of nesting.

src/ad-services/nativo/index.ts
```diff
@@ -56,6 +108,9 @@ class Nativo {
 						<button class="ntv-ad-button">buy now</button>
 					</div>
 				</div>`;
+		} else {
```
Tomasz-at-Fandom, line 111: If we would reverse the statement, move logger before the HTML block and make an early return then the else would be obsolete and the HTML block would have one less indentation :)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -87,12 +85,12 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
...
+					container.setAttribute('id', Nativo.INCONTENT_AD_SLOT_NAME);
+					container.classList.add(...Nativo.SLOT_CLASS_LIST);
 
 					anchor.before(container);
 
-					nativo.start();
+					nativo.requestAd();
 				}
 			}
 		});
```
Tomasz-at-Fandom, line 96: Could we refactor contents of this callback? 
There are two`if` statements here that cause deep indentations. We could reverse them and make early returns instead of nesting.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -104,31 +102,48 @@ export class UcpMobileSlotsDefinitionRepository {
...
+						nativo.requestAd();
 					}
 				});
 			},
 		};
 	}
 
+	getNativoFeedAdConfig(): SlotSetupDefinition {
+		if (!nativo.isEnabled()) {
+			return;
```
Tomasz-at-Fandom, line 129: It's strange that this is getting compiled, we should add `void` as a possible return value.

nandy-andy: Yep, it's weird but it doesn't break the code as we filter out from the array incorrect elements: https://github.com/Wikia/ad-engine/pull/1163/files#diff-0a44597c77526a4a5c919898980fff7ba9ae62c3fd10c4e11b658d9372ce9bfbL79

Tomasz-at-Fandom (approved):

dmnsobczak (approved):

kwfandom (approved):

preludek (approved):

PR #1164, opened 2021-11-02
ADEN-11322 add bottom label for sticky bfaa

jakubbanas:


platforms/shared/templates/handlers/bfaa/bfaa-new-big-advertisement-label-handler.service.ts
```diff
@@ -0,0 +1,29 @@
+import { DomListener, TemplateStateHandler } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { Subject } from 'rxjs';
+import { startWith, takeUntil, tap } from 'rxjs/operators';
+import { UapDomManager } from '../../helpers/uap-dom-manager';
+
+@Injectable({ autobind: false })
+export class BfaaBigAdvertisementLabelHandler implements TemplateStateHandler {
```
preludek, line 8: I'd probably go with something like `BfaaAdvertismentSectionLabelHandler` or something like that, but since it's just for test purposes let's leave it as it is.

platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -83,6 +83,10 @@ export class UapDomManager {
 		this.setPlaceholderHeight(`${this.reader.getSlotHeightImpact()}px`);
 	}
 
+	setBigAdvertisementLabel(): void {
```
preludek, line 86: I feel like this should be a private method in BfaaBigAdvertisementLabelHandler.

preludek (approved):

PR #1171, opened 2021-11-09
ADEN-11342 - Remove slot: invisible_skin

szymon-nawrocki:
## Links:
https://fandom.atlassian.net/browse/ADEN-11342

## Description
Slot _invisible_skin_ is no longer used, so it should be removed from AdEngine.

### DIFF
--- a/spec/ad-engine/models/ad-slot.spec.ts
+++ b/spec/ad-engine/models/ad-slot.spec.ts
@@ -57,13 +57,6 @@ describe('ad-slot', () => {
 		expect(adSlot.getAdUnit()).to.equal('/5441/something/_article/top_boxad');
 	});
 
-	it('with other ad unit', () => {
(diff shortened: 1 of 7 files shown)


spec/ad-engine/models/ad-slot.spec.ts
```diff
@@ -57,13 +57,6 @@ describe('ad-slot', () => {
 		expect(adSlot.getAdUnit()).to.equal('/5441/something/_article/top_boxad');
 	});
 
-	it('with other ad unit', () => {
```
nandy-andy, line 60: Instead of removing the unit test we could replace the `INVISIBLE_SKIN` with some other slot here and keep the check of ad unit string for other page type.

nandy-andy (changes requested):

PR #1153, opened 2021-10-20
ADEN-11290 Collapse ad placeholders on pages with UAP

szymon-nawrocki:
## Ticket: 
https://fandom.atlassian.net/browse/ADEN-11290

## Concurrent Pull Request in UCP
https://github.com/Wikia/unified-platform/pull/7774

## Description
~~1. Create PlaceholderService class for all functionalities connected with placeholders.~~
Refactor placeholder logic so it's more generic.

PR #1180, opened 2021-11-12
ADEN 10981 - Experiment: add collapse button to not filled slot

szymon-nawrocki:
## Links:
https://fandom.atlassian.net/browse/ADEN-10981

## Description
This experiment is meant to implement a Call To Action message for collapsed slot once nothing was served from GAM.


platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -234,28 +235,61 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
+						action['payload'][1] === statusToUndoCollapse
+					) {
+						placeholder?.classList.remove('hide');
+						return;
+					}
 
-				if (statusesToStopLoadingSlot.includes(action['event'])) {
-					placeholder?.classList.remove('is-loading');
-				}
+					if (
```
szymon-nawrocki, line 247: I've added these checks, because in almost every case slot sends:
* 2x events from  statusesToStopLoading (example: "success" & "rendered")
* 2x events from statusesToCollapse (example: "slotHidden" & "slotBlocked")

So the actions:  hide placeholder, hide ad label, stop loading slot, are almost always being called twice.

I was thinking about better and more elegant way to do it (without so much `ifs`) but so far couldn't find any.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -234,28 +235,61 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
-					if (this.isUapLoaded) {
-						placeholder?.classList.add('hide');
-					} else {
-						adSlot.getAdLabel(adLabelParent)?.classList.add('hide');
+					if (statusesToCollapse.includes(action['event'])) {
+						if (this.isUapLoaded && !placeholder.classList.contains('hide')) {
+							placeholder.classList.add('hide');
+						} else {
+							const adLabel = adSlot.getAdLabel(adLabelParent);
+							if (adLabel && !adLabel.classList.contains('hide')) {
```
nandy-andy, line 259: It's hard to read - too much indentations 🙈 Consider refactoring it in a way the ifs are replaced with functions or maybe adding early returns somewhere 🤔

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -234,28 +235,61 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
+								adLabel.classList.add('hide');
+								this.addMessageBoxToCollapsedElement(placeholder, adSlot);
+							}
+						}
 					}
 				}
 			});
 	}
 
+	private addMessageBoxToCollapsedElement(placeholder: HTMLElement, adslot: AdSlot): void {
```
nandy-andy, line 269: You can move this function to a separate module and even split it into smaller ones. This way the `ucp-mobile-dynamic-slots.setup.ts` won't get that big and we'll have one module to remove once we decide the experiment is finished and the code is necessary. It may also help with introducing instant-service variable that switches on/off the functionality.

src/ad-tracking/click-tracker.ts
```diff
@@ -12,6 +12,11 @@ export const videoLearnMoreDisplayedEvent = globalAction(
 	props<{ adSlotName: string; learnMoreLink: HTMLElement }>(),
 );
 
+export const hideCollapsedAdEvent = globalAction(
```
nandy-andy, line 15: I'm thinking about the naming here. Can't we go with just `hiddenAdEvent` or `collapsedAdEvent`? 🤔

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
nandy-andy: Is focusing on `success`, `slotHidden` and `slotBlocked` makes anything easier? I wanted to suggest just focusing on `success` and `slotHidden` but I imagine there is some weird difference between `slotHidden` and `slotBlocked` (maybe there isn't and then we can simplify it!) 🙈 However, I think `success` and `rendered` can be the same here if we first handle blocked and hidden 🤔 Just food for thoughts ;)

src/ad-tracking/click-tracker.ts
```diff
@@ -43,6 +55,14 @@ class AdClickTracker {
 			.subscribe(async ({ adSlotName, learnMoreLink }) => {
 				this.addClickVideoLearnMoreTrackingListeners(callback, adSlotName, learnMoreLink);
 			});
+
+		communicationService.action$
+			.pipe(ofType(hideCollapsedAdEvent))
+			.subscribe(async ({ adSlotName, collapseButton, ad_status }) => {
+				collapseButton.addEventListener('click', () => {
+					this.handleClickEvent({ ad_status, callback, slot: slotService.get(adSlotName) });
```
nandy-andy, line 63: It's not Python, let's follow the camelCase naming convention for variables here ;) `r/ad_status/adStatus`

src/ad-tracking/click-tracker.ts
```diff
@@ -20,6 +25,13 @@ interface AdClickContext {
 	};
 }
 
+interface HandleClickParams {
+	callback: FuncPipelineStep<AdClickContext>;
+	slot: AdSlot;
+	event?: MouseEvent;
+	ad_status?: string;
```
nandy-andy, line 32: Do we need the `ad_status` here? Can't we get it from the `AdSlot` instance?

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
szymon-nawrocki: Great question, which made me thinking about refactoring _registerAdPlaceholderHandler()_

- Status `rendered` was used, to be able to undo the collapse of TLB on UAP, and it is connected with this PR:
https://github.com/Wikia/ad-engine/pull/1153/commits/25991a6a40a1056100f705aa37b749c33fb1d6c2
But for _stopLoading_ itself, listening to status `success` should be enough.

- If it comes to statusesToCollapse, I observed that: when slots gets one of these statuses: `blocked`, `collapsed`, `force_collapsed`, it always also gets status `slotHidden`.
So maybe we could listen only to `slotHidden` status, which will take care of every case and simplify the whole logic.

I will try to test this solution.

szymon-nawrocki: Okay, I included these thoughts into code, and indeed it simplified the code.
Thanks for the great tip.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -234,23 +250,30 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 				const adLabelParent = adSlot.getConfigProperty('placeholder')?.adLabelParent;
 
-				if (
-					action['event'] === AdSlot.SLOT_RENDERED_EVENT &&
-					action['payload'][1] === statusToUndoCollapse
-				) {
-					placeholder?.classList.remove('hide');
-					return;
-				}
+				if (placeholder) {
```
nandy-andy, line 253: This condition adds one more indentation. More indentation === more unreadable code. If you think we need it here, consider using an early return like we do in line #248 in order to safe indentation 🤔

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -234,23 +250,30 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
+					if (shouldUndoCollapse(action['event'], action['payload'][1])) {
```
nandy-andy, line 254: Judging by what's happening within this condition I wouldn't name the function `shouldUndoCollapse()`. Once it's a match we unhide the placeholder, don't we? If my thinking is correct I suggest naming the condition function `shouldDisplayPlaceholder()` 🤔

platforms/shared/utils/message-box.ts
```diff
@@ -0,0 +1,25 @@
+import { AdSlot, collapsedAdEvent, communicationService } from '@wikia/ad-engine';
+
+export const addMessageBoxToCollapsedElement = (placeholder: HTMLElement, adSlot: AdSlot): void => {
+	const messageBox = document.createElement('div');
+	messageBox.className = 'message-box';
+
+	const button = document.createElement('button');
+	button.className = 'collapse-btn';
```
nandy-andy, line 8: We could "configure" the button in a separate function and in this function just assign it to the `button` and append to the `messageBox` (as we already do in the line 23). This way the `addMessageBoxToCollapsedElement()` shorter and more readable.

platforms/shared/utils/message-box.ts
```diff
@@ -0,0 +1,25 @@
...
+	button.innerHTML = 'hide';
+	button.onclick = () => placeholder.classList.add('hide');
+	adSlot.setStatus(AdSlot.STATUS_CLICKED_COLLAPSE);
```
nandy-andy, line 11: This line and the following one don't seem to me there are in the right place. I'm reading it in the way that we're changing slot status to "clicked collapsed" and dispatch the event when we add the message box to the collapsed element. I believe we should dispatch the event when user clicks the hide button. Unfortunately, it means we need to rethink the implementation of click tracking and get rid off `AdSlot.STATUS_CLICKED_COLLAPSE` 🙈

src/ad-tracking/click-tracker.ts
```diff
@@ -12,6 +12,11 @@ export const videoLearnMoreDisplayedEvent = globalAction(
 	props<{ adSlotName: string; learnMoreLink: HTMLElement }>(),
 );
 
+export const collapsedAdEvent = globalAction(
+	'[AdEngine] Collapsed ad hidden',
```
nandy-andy, line 16: What do you think of a simple message: `[AdEngine] Collapsed ad`? I also wonder if the action messages are in the present or past tense 🤔 Small details, nit-picking ;)

szymon-nawrocki: Great idea. Action messages are in the past tense.

Actually `collapsedAdEvent` doesn't fit that good for a click event, which hides `messageBox` so I decided to change it all together to:
`export const hideMessageBoxEvent = globalAction(
	'[AdEngine] MessageBox hidden',
	props<{ adSlotName: string }>(),
);`

platforms/shared/utils/placeholder-service.ts
```diff
@@ -0,0 +1,62 @@
...
+	communicationService,
+	ofType,
+	slotService,
+	uapLoadStatus,
+} from '@wikia/ad-engine';
+import { filter, take } from 'rxjs/operators';
+import { messageBox } from './message-box';
+import { placeholderHelper } from './placeholder-service-helper';
+
+class PlaceholderService {
```
szymon-nawrocki, line 12: @dmnsobczak @Tomasz-at-Fandom 
Could you tell me where it is the best place, to put:
- this service
- placeholder-service-helper.ts
- message-box.ts

I though that the best place will be: _src->ad-engine->services_ 
But due to eslint error: "blacklisted-paths", I couldn't import and use another services in those files.

I'll appreciate your feedback

dmnsobczak: You can't import other services in ad-engine core. You can only import ad-engine core in other services.

dmnsobczak: What about `src/ad-products` maybe? 🤔

Tomasz-at-Fandom: Weren't we supposed to move away from putting things in `/src` and move over to platforms?
We should probably talk about writing down a plan/definition for the "big picture" of ad-engine architecture.

Btw it doesn't look like these files are using anything that is not already in ad-engine/core or is not already imported in to it, so maybe it's just a problem with how you import it. Have you tried using relative imports instead of that budler `@wikia` path or whatever it's called?

Tomasz-at-Fandom: Oh and if you're wondering what's ad-engine/core, you can check out `src/tsconfig.json` and look for the `paths` key AFAIK there are the definitions of those @wikia paths.

Tomasz-at-Fandom: btw how about making some unit tests for these classes? :D

szymon-nawrocki: So if we shouldn't put more things int `/src` I think I'll stick to `/platforms` directory.

This service is currently being used ONLY by UcpMobile, but I put it into `/platforms/shared/utils` even that nothing is "shared" between the platforms.

I put it there, because I couldn't find a better place in `/platforms/ucp-mobile` directory. Maybe we should create one?

szymon-nawrocki: @Tomasz-at-Fandom great idea about making unit tests for these classes. I'm down for it!

Tomasz-at-Fandom: I'm not sure about that `src` thing, it's just a foggy recollection of something someone said, might be incorrect.

However `/platforms/shared/utils` sounds ok for now as long as it doesn't do anything mobile specific :)

jakubbanas (approved):

nandy-andy (approved):

PR #1181, opened 2021-11-15
ADEN-11358 remove Real Vu integration

preludek:


platforms/shared/setup/base-context.setup.ts
```diff
@@ -137,7 +137,6 @@ export class BaseContextSetup implements DiProcess {
 		context.set('services.silverSurfer', this.instantConfig.get('icSilverSurfer'));
 		context.set('services.durationMedia.enabled', this.instantConfig.get('icDurationMedia'));
 		context.set('services.distroScale.enabled', this.instantConfig.get('icDistroScale'));
-		context.set('services.realVu.enabled', this.instantConfig.get('icRealVu'));
```
szwajkowski, line 140: should we create ticket to remove it from ICBM or can it be already removed ?

dmnsobczak: If it's disabled: we can remove it right away.

dmnsobczak (approved):

PR #1182, opened 2021-11-17
ADEN-11321 - native ads code level prio

preludek:


platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -21,6 +21,7 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
 			this.logger('Disabling incontent_player and affiliate_slot');
 			slotService.disable('incontent_player', 'hivi-collapse');
 			slotService.disable('affiliate_slot', 'hivi-collapse');
+			context.set('services.nativo.enabled', false);
```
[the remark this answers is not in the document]
preludek: That's a good idea 👍  done

jakubbanas (approved):

platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -21,6 +21,7 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
 			this.logger('Disabling incontent_player and affiliate_slot');
 			slotService.disable('incontent_player', 'hivi-collapse');
 			slotService.disable('affiliate_slot', 'hivi-collapse');
+			this.adSlot.emitEvent('Stick TLB');
```
nandy-andy, line 24: Let's stick to the convention and use lower-cases with dash in the event name. Additionally, in order to have it all in once place we could out the new event in https://github.com/Wikia/ad-engine/blob/d3dcb14ee763ade57c15a0389c5917e1698116d8/src/ad-engine/services/events.ts#L6-L30

src/ad-services/nativo/index.ts
```diff
@@ -32,11 +37,31 @@ export class Nativo {
...
+
+	checkCodePriority(): boolean {
+		return !(context.get('custom.hasFeaturedVideo') || context.get('templates.stickyTlb.forced'));
+	}
+
+	setCodePriorityListener(): void {
+		communicationService.action$
+			.pipe(
+				ofType(adSlotEvent),
+				filter((action) => action.payload.some((el) => el.status === 'Stick TLB')),
```
nandy-andy, line 56: Instead of hardcoded string you could reuse a defined event from https://github.com/Wikia/ad-engine/blob/d3dcb14ee763ade57c15a0389c5917e1698116d8/src/ad-engine/services/events.ts#L6-L30

platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
preludek: done 👍

src/ad-services/nativo/index.ts
preludek: done 👍

preludek: done

PR #1183, opened 2021-11-17
ADEN-11324 | [Prebidium] fix: hide Popular Pages on FRM load when

Tomasz-at-Fandom:


src/ad-products/utils/fmr-rotator.ts
```diff
@@ -36,8 +36,15 @@ export class FmrRotator {
...
+				if (context.get('state.provider') === 'prebidium') {
+					slot.once(AdSlot.STATUS_SUCCESS, () => {
+						this.swapRecirculation(false);
+					});
+					return;
+				}
+
 				if (universalAdPackage.isFanTakeoverLoaded()) {
 					slot.once(AdSlot.STATUS_SUCCESS, () => {
 						this.swapRecirculation(false);
```
dmnsobczak, line 50: If we're starting to use `returns` here then you can add `return` in L51 and move `else` content one level higher to remove one nest and make code easier to read. Just a suggestion :D

src/ad-engine/providers/prebidium-provider.ts
```diff
@@ -32,6 +32,7 @@ export class PrebidiumProvider implements Provider {
 				if (doc && adId) {
 					pbjs.renderAd(doc, adId);
 					adSlot.getElement()?.classList?.remove('hide');
+					adSlot.success();
```
dmnsobczak, line 35: Calling success here is such a excellent idea 💎

dmnsobczak (approved):

PR #1184, opened 2021-11-18
ADEN-11327 revert liveramp

jakubbanas:


src/ad-bidders/prebid/index.ts
```diff
@@ -47,15 +49,20 @@ export class PrebidProvider extends BidderProvider {
...
 
 		this.isLazyLoadingEnabled = this.bidderConfig.lazyLoadingEnabled;
+		this.adUnits = setupAdUnits(this.isLazyLoadingEnabled ? 'pre' : 'off');
+		this.bidsRefreshing = context.get('bidders.prebid.bidsRefreshing') || {};
+		this.isATSAnalyticsEnabled = context.get('bidders.liveRampATSAnalytics.enabled');
+
 		this.prebidConfig = {
 			debug: ['1', 'true'].includes(utils.queryString.get('pbjs_debug')),
-			enableSendAllBids: true,
+			enableSendAllBids: !!context.get('bidders.prebid.sendAllBids'),
```
Tomasz-at-Fandom, line 65: Do we want this back? I don't see this being set anywhere.

Tomasz-at-Fandom: @dmnsobczak removed this only recently as part of ADEN-11248, are we certain it is related?

Tomasz-at-Fandom: This is also being called on line 236.

jakubbanas: Good point, let me check that.

Tomasz-at-Fandom (approved):

PR #1186, opened 2021-11-22
ADEN-11352 Fixing labels translation system

dmnsobczak:


src/ad-engine/utils/i18n.ts
```diff
@@ -0,0 +1,25 @@
+import { context } from '../services/context-service';
+import { TRANSLATIONS } from './translations';
+
+const defaultLanguage = 'en';
+
+export function getTranslation(category: string, key: string): string {
```
nandy-andy, line 6: Do we need `category` if there is [only one](https://github.com/Wikia/ad-engine/blob/dev/src/ad-products/common/translations.ts)?

platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
```diff
@@ -48,7 +48,8 @@ export class UcpDesktopAdsMode implements DiProcess {
 		this.setAdStack();
 		this.trackAdEngineStatus();
 		this.trackTabId();
-		this.translateLabels();
+
+		utils.translateLabels();
```
nandy-andy, line 52: Why do we still need to call it? We translate the labels in all connected places like: [slotCreator](https://github.com/Wikia/ad-engine/pull/1186/files#diff-23599229e919b398dd805ce5482ab0504f06c3e54ea1eb97d2bb00928e552839R152), [slotPlaceholderInjector](https://github.com/Wikia/ad-engine/pull/1186/files#diff-59abc8877958d56e9895f2f494298fda339c4fe03986a946c0a818991da54fd7R79) and [ad products](https://github.com/Wikia/ad-engine/pull/1186/files#diff-22828f538a026ac4d69901c900fbabb7cd6204c32c50d62e750124648972abfcR8)? Not calling it will stop us from traversing the DOM unnecessarily.

dmnsobczak: This translates labels for `.top-ads-container` and `.bottom-ads-container`, which are served with HTML. We want to translate them too as fast as possible :)

src/ad-engine/utils/i18n.ts
dmnsobczak: I think we had two categories back in the past, but we removed one of them and we never added any new translations. I would leave it like it is: we now know how to use Crowdin and we have plenty of texts in AdEngine (like disclaimers or buttons), which we can translate in the future. But if you want, I can remove this extra feature in a separate, easy to revert commit :D

dmnsobczak: Discussed on Slack: removing 👍

platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
nandy-andy: Oh, but then we have translation in MediaWiki and we can translate it there instead of traversing DOM 🤔 Damn, but that would be cached on Varnishes, so for anon users we'd need to translate it on the front-end anyway. Damn!

nandy-andy (approved):

PR #1187, opened 2021-11-23
ADEN-11373 Fix a forced_collapse behaviour

nandy-andy:
A quick-fix for P2/P3 reported in [ADEN-11373](https://fandom.atlassian.net/browse/ADEN-11373). This code changes the way AdEngine reacts to "forced_collapse" signal from GAM, so it really collapses the slot causing a content shift.


platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -156,6 +153,9 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
 		} else if (statusesToHideLabel.includes(slotStatus)) {
 			placeholder?.classList.remove('is-loading');
 			adSlot.getAdLabel(adLabelParent)?.classList.add('hide');
+		} else if (statusesToCollapse.includes(slotStatus)) {
+			placeholder.style.minHeight = '0';
+			adSlot.getAdLabel(adLabelParent)?.classList.add('hide');
```
nandy-andy, line 158: Additionally, to make it cleaner we could:
1. move this line to the ad slot under `hideAdLabel()` method,
2. move lines 152 and 154 to the `AdSlot` as well and make it a method `stopLoadingPlaceholder()`,
3. move line 157 to the `AdSlot` under `hidePlaceholder()` method.

I didn't do it now because changes in the `AdSlot` affect FandomMobile and FandomDesktop and I've got an impression we're supposed to rush with fixing the FandomDesktop part now 😊

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -156,6 +153,9 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
 		} else if (statusesToHideLabel.includes(slotStatus)) {
 			placeholder?.classList.remove('is-loading');
 			adSlot.getAdLabel(adLabelParent)?.classList.add('hide');
+		} else if (statusesToCollapse.includes(slotStatus)) {
+			placeholder.style.minHeight = '0';
```
dmnsobczak, line 157: It works fine I guess, but can't we just hide parent placeholder element? There is almost 50px of padding there which is still visible and still ads some unexpected blank space 🤔 
I did it 5 majors ago, but it was removed while working on ad placeholders:
https://github.com/Wikia/ad-engine/blob/v93.0.0/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts#L37
```} else if (withHide === 'parent') { placeholder?.parentElement?.setAttribute('style', 'display: none');```
And then we were hiding both HiVi-Leaderboard force-collapse and Top_leaderboard standard collapse:
https://github.com/Wikia/ad-engine/blob/3cb99627bc7753c953270f93d65ff4c693829b9d/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts#L149

dmnsobczak (approved):
Code works fine, just questioning if this is what we want to achieve. Please accept or reject the questions and then move it to QA or re-CR :)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
nandy-andy: Sure, let me try hide it in a more elegant way by adding a class `hide` 🤞

dmnsobczak (approved):

PR #1192, opened 2021-12-01
ADEN-11397 featured-video events fix

preludek:


dmnsobczak (approved):

src/ad-services/permutive/video-tracker.ts
```diff
@@ -30,7 +30,7 @@ interface VideoEvent {
 		adPlayId: string;
 		ima: {
 			ad: {
-				g: EventAdInfo;
+				h: EventAdInfo;
```
mfarul, line 33: I guess next will be 'i' :). This was API change for all GEO? They will not revert to it?

PR #1197, opened 2021-12-07
ADEN-11424 Fix regression for Permutive

nandy-andy:
This is a quick-fix for regression we introduced while working on src=uap. More info in the ticket: [ADEN-11424](https://fandom.atlassian.net/browse/ADEN-11424)

### DIFF
--- a/spec/ad-services/permutive/permutive.spec.ts
+++ b/spec/ad-services/permutive/permutive.spec.ts
@@ -91,18 +91,15 @@ describe('Permutive integration', () => {
 
 		expect(permutiveKeys).to.eql([]);
 	});
-});
 
-describe('Permutive - make sure src is a string', () => {
-	it('Permutive gets string if src is a string', () => {
+	it('Permutive pass src in correct format', () => {
 		context.set('src', 'gpt');
-
 		expect(permutive.getSrcFromContext()).to.equal('gpt');
-	});
 
-	it('Permutive gets string if src is an array', () => {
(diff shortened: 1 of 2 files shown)


src/ad-services/permutive/index.ts
```diff
@@ -22,6 +22,12 @@ class Permutive {
 		this.setAddon();
 	}
 
+	getSrcFromContext() {
+		const src = context.get('src');
+
+		return typeof(src) === 'object' ? src[0] : src;
```
preludek, line 28: Is there a possibility src to be an empty array, object or null? We will get undefined or an error in that case.

spec/ad-services/permutive/permutive.spec.ts
```diff
@@ -92,3 +92,17 @@ describe('Permutive', () => {
 		expect(permutiveKeys).to.eql([]);
 	});
 });
+
+describe('Permutive - make sure src is a string', () => {
```
preludek, line 96: I'm thinking about not making it another `describe` but moving those 2 specs in `Permutive integration` and renaming them. Maybe `Permutive pass src in correct format when src is type string` and the other similarly?

src/ad-services/permutive/index.ts
nandy-andy: IMHO there is none possibility like that because AFAIK we set `src` on every platform even Bingebot. But it does not hurt to add additional check 🤔

preludek (approved):

PR #1198, opened 2021-12-07
ADEN-10763 Adding 4x4 and 5x5 companion sizes

dmnsobczak:


platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
```diff
@@ -25,10 +26,15 @@ export class BfaaF2ConfigHandler implements TemplateStateHandler {
...
+
+		if (this.f2Env.siteType === 'app') {
+			slotsContext.addSlotSize('top_boxad', [4, 4]);
+			slotsContext.addSlotSize('incontent_boxad', [4, 4]);
+			slotsContext.setSlotSize('bottom_leaderboard', [2, 2]);
+		} else {
+			slotsContext.addSlotSize('top_boxad', [5, 5]);
+			slotsContext.addSlotSize('incontent_boxad', [5, 5]);
+			slotsContext.setSlotSize('bottom_leaderboard', [3, 3]);
+		}
```
Tomasz-at-Fandom, line 38: Can we make this code a bit more self explanatory?
E.g. maybe let's put this these arrays with numbers into variables with meaningful names.
Maybe we could even put them in a separate file dedicated to such "special sizes" so that we could import them across the project, because this is happening elsewhere.
If the above doesn't make sense then let's at least leave a comment explaining why this.f2Env.siteType makes us change the magical sizes that we set on these slots and what do they mean.

platforms/shared/slots/slots-context.ts
```diff
@@ -9,12 +9,20 @@ import {
...
-			sizeMap.sizes.push(size);
-		});
+		if (definedViewportSizes) {
+			definedViewportSizes.forEach((sizeMap) => {
+				sizeMap.sizes.push(size);
+			});
+		}
+	}
+
+	setSlotSize(slotName: string, size: [number, number]): void {
```
Tomasz-at-Fandom, line 23: `slotName` could be an enum or at least we could define a list of acceptable strings, to be more precise and avoid spelling errors and thing like that.

dmnsobczak: The problem is that this file in is the `shared` directory. The good thing to keep is to use slot names only inside the platform directory, not in the shared, cores or templates. If they are required there, the we should pass them as a parameter. To avoid that, but to still keep some level of data validation I added additional if clause to check whether the slot is defined in the platform context.

platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
dmnsobczak: I thought about it and I decided that it might be overcomplicating the code, but since there are two of us... :D Sure, I putted the sizes in the `UAP constants` file and I'm importing them from there. They have readable names so it should be easier to understand the code now :)

platforms/shared/slots/slots-context.ts
```diff
@@ -9,12 +9,28 @@ import {
 
 class SlotsContext {
 	addSlotSize(slotName: string, size: [number, number]): void {
-		const definedViewportSizes = context.get(`slots.${slotName}.sizes`);
+		if (!context.get(`slots.${slotName}`)) {
+			return;
```
Tomasz-at-Fandom, line 13: Is there a scenario where calling this functions with a unset slot name is acceptable/necessary?
If not then how about throwing an error here, so that we don't do it by accident :)

dmnsobczak: Good idea, changed :D

Tomasz-at-Fandom (approved):

PR #1199, opened 2021-12-07
ADEN-11411 Add impressions tracking to the collapse message

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11411

### Description
Because experiment with adding CTA message in places of collapsed ads was successful we want to add more messages, and we need additional tracking of the impressions of such messages


platforms/shared/tracking/tracking.setup.ts
```diff
@@ -125,6 +127,20 @@ export class TrackingSetup {
 			});
 	}
 
+	private ctaTracker(): void {
+		if (this.slotTrackingMiddlewares.length === 0) {
```
nandy-andy, line 131: Why can the `slotTrackingMiddlewares` be empty? I can see we do it in the `adClickTracker()` and similar thing in `bidderTracker()` but it seems like a workaround, a red flag. No need to address now but maybe we need a follow-up to understand why the condition is needed/what causes `slotTrackingMiddlewares` to be empty.

src/ad-tracking/cta-tracker.ts
```diff
@@ -0,0 +1,60 @@
+import { communicationService, globalAction, ofType } from '@ad-engine/communication';
+import { AdSlot, FuncPipeline, FuncPipelineStep, slotService } from '@ad-engine/core';
+import { props } from 'ts-action';
+
+type AdStatus =
```
nandy-andy, line 5: It looks clean TBH :) Also, it follows our good practice of [type over enum](https://fandom.atlassian.net/wiki/spaces/ADEN/pages/506364058/TypeScript+Style+Guide#TypeScriptStyleGuide-TypeoverEnum) 👍

The only part I'm not sure of... is the first pipe necessary, doesn't it cause any side effect? I'd without the confusing pipe at the beginning ;)
```suggestion
type AdStatus = 'cm_register_impression'
	| 'cm_register_clicked'
	| 'cm_newsletter_impression'
	| 'cm_newsletter_clicked'
	| 'cm_fanlab_impression'
	| 'cm_fanlab_clicked';
```

nandy-andy (approved):

platforms/shared/tracking/tracking.setup.ts
szymon-nawrocki: I admit that I took this part from `adClickTracker()` and I am not really sure why it can be empty. And now when you pointed this out it indeed looks like a workaround which could use further investigation.

PR #1200, opened 2021-12-08
ADEN-10381 Using SlotsCreator on all platforms

dmnsobczak:


platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -28,7 +24,7 @@ export class F2DynamicSlotsSetup implements DiProcess {
 	private injectSlots(): void {
 		const topLeaderboardDefinition = this.slotsDefinitionRepository.getTopLeaderboardConfig();
 
-		this.insertSlots([
+		insertSlots([
```
dmnsobczak, line 27: 4 copies replaced with one module

platforms/f2/setup/dynamic-slots/f2-slots-definition-repository.ts
```diff
@@ -1,15 +1,10 @@
-import { context, SlotCreatorConfig, SlotCreatorWrapperConfig } from '@wikia/ad-engine';
+import { SlotSetupDefinition } from '@platforms/shared';
+import { context, SlotCreatorWrapperConfig } from '@wikia/ad-engine';
 import { Inject, Injectable } from '@wikia/dependency-injection';
 import { F2_ENV, F2Environment } from '../../setup-f2';
 import { F2State } from '../../utils/f2-state';
 import { F2_STATE } from '../../utils/f2-state-binder';
 
-export interface SlotSetupDefinition {
```
dmnsobczak, line 7: Duplicates removed

platforms/futhead/futhead-platform.ts
```diff
@@ -49,8 +48,7 @@ export class FutheadPlatform {
 			FutheadTargetingSetup,
 			FutheadPrebidConfigSetup,
 			SportsA9ConfigSetup,
-			FutheadDynamicSlotsSetup,
-			CurseSlotsStateSetup,
+			CurseDynamicSlotsSetup,
```
dmnsobczak, line 51: Unification

platforms/shared/dynamic-slots/curse-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,29 @@
...
+@Injectable()
+export class CurseDynamicSlotsSetup implements DiProcess {
+	private uapFirstCallSlotName = 'cdm-zone-01';
+
+	constructor(private slotsDefinitionRepository: CurseSlotsDefinitionRepository) {}
+
+	execute(): void {
+		this.configureUap();
+		insertSlots([
+			this.slotsDefinitionRepository.getCdmZoneConfig(1),
```
dmnsobczak, line 16: There is no need to create 5 almost the same methods

platforms/shared/index.ts
```diff
@@ -25,9 +24,11 @@ export * from './services/no-ads-detector';
 export * from './utils/get-domain';
 export * from './utils/get-sports-page-type';
 export * from './utils/get-media-wiki-variable';
+export * from './utils/insert-slots';
 export * from './utils/instant-config-setup';
 export * from './utils/message-box';
 export * from './utils/placeholder-service';
+export * from './utils/placeholder-service-helper';
```
dmnsobczak, line 31: Fixed direct import

platforms/shared/setup/base-context.setup.ts
```diff
@@ -71,7 +71,6 @@ export class BaseContextSetup implements DiProcess {
 		);
 		context.set('options.tracking.slot.bidder', this.instantConfig.get('icBidsTracking'));
 		context.set('options.tracking.postmessage', this.instantConfig.get('icPostmessageTracking'));
-		context.set('options.hiviLeaderboard', this.instantConfig.get('icHiViLeaderboardSlot'));
```
dmnsobczak, line 74: HiViLB removed step 2: https://fandom.atlassian.net/browse/ADEN-11390

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -75,8 +49,6 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 				aboveTheFold: true,
 				slotNameSuffix: '',
 				group: 'MR',
-				parentContainerSelector: '.main-page-tag-rcs, #rail-boxad-wrapper',
```
dmnsobczak, line 78: Selectors moved to SlotsCreator config

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -106,18 +77,6 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 					[300, 250],
 					[300, 600],
 				],
-				repeat: {
```
dmnsobczak, line 109: Repeat config will be reapplied here by SlotCreator

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -164,8 +122,6 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 				audio: false,
 				isVideo: true,
 				insertBeforeSelector: '#mw-content-text > div > h2',
-				insertBelowFirstViewport: true,
-				disabled: true,
```
dmnsobczak, line 168: We don't need to predisable slots anymore

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -206,7 +159,6 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 				adProduct: 'featured',
 				slotNameSuffix: '',
 				group: 'VIDEO',
-				lowerSlotName: 'featured',
```
dmnsobczak, line 209: Old and forgotten

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
...
-	scrollListener,
-	SlotConfig,
-	slotInjector,
 	slotService,
-	uapLoadStatus,
-	utils,
 } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
-import { take } from 'rxjs/operators';
-import { desktopFanFeedNativeAdListener } from './desktop-fan-feed-native-ad-listener';
```
dmnsobczak, line 28: File contents moved

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
...
-		this.injectIncontentPlayer();
-		this.injectFloorAdhesion();
-		this.injectBottomLeaderboard();
-		this.injectNativeAdsPlaceholder();
-		this.injectNativeFanFeed();
 		this.configureTopLeaderboard();
+		this.configureIncontentBoxad();
 		this.configureIncontentPlayerFiller();
+		this.configureFloorAdhesionCodePriority();
+		this.registerAdPlaceholderService();
```
dmnsobczak, line 32: New placeholders added

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
...
 			}
-		});
-	}
-
-	private injectNativeAdsPlaceholder(): void {
-		if (!nativo.isEnabled()) {
-			return;
-		}
-
-		const pageHeaders = document.querySelectorAll('.mw-headline');
```
dmnsobczak, line 77: Now handled by SlotCreator

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -0,0 +1,243 @@
...
+				classList: ['hide', 'ad-slot'],
+			},
+			activator: () => {
+				context.push('state.adStack', { id: slotName });
+			},
+		};
+	}
+
+	getIncontentBoxadConfig(): SlotSetupDefinition {
+		if (!this.isRightRailApplicable()) {
```
dmnsobczak, line 65: New condition for ICB

platforms/ucp-mobile/ad-context.ts
```diff
@@ -39,7 +39,7 @@ export const basicContext = {
 			ids: [],
 		},
 		pushAfterRendered: {
-			top_boxad: ['incontent_boxad_1', 'incontent_player'],
+			top_boxad: ['incontent_player'],
```
dmnsobczak, line 42: Unificated with redundant context change

platforms/ucp-mobile/setup/context/base/ucp-mobile-base-context.setup.ts
```diff
@@ -13,17 +13,10 @@ export class UcpMobileBaseContextSetup extends BaseContextSetup {
 		);
 
 		context.set('options.tracking.tabId', this.instantConfig.get('icTabIdTracking'));
-		// FIXME: sourced from front/scripts/shared/tracking/Tracker.js getUserIdForInternalTracking()
```
dmnsobczak, line 16: Old and forgotten

platforms/ucp-mobile/setup/context/slots/ucp-mobile-slots-context.setup.ts
```diff
@@ -165,8 +151,6 @@ export class UcpMobileSlotsContextSetup implements DiProcess {
 			},
 			interstitial: {
 				adProduct: 'interstitial',
-				disabled: true,
-				disableManualInsert: true,
```
dmnsobczak, line 169: Old value

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -9,18 +15,13 @@ import {
...
 	fillerService,
 	PorvataFiller,
-	SlotCreator,
 	slotService,
 	uapLoadStatus,
 	universalAdPackage,
 	utils,
 } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
-import { PlaceholderServiceHelper } from '../../../shared/utils/placeholder-service-helper';
```
dmnsobczak, line 19: Fixed import

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -139,13 +122,15 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
 				if (this.CODE_PRIORITY.floor_adhesion.active) {
 					this.CODE_PRIORITY.floor_adhesion.active = false;
 					slotService.disable('floor_adhesion', AdSlot.STATUS_CLOSED_BY_PORVATA);
+					document.getElementById('floor_adhesion_anchor').classList.add('hide');
```
dmnsobczak, line 125: Porvata should close Floor Adhesion: fix

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -184,6 +178,7 @@ export class UcpMobileSlotsDefinitionRepository {
 						adProduct: '{slotConfig.slotName}',
 						'targeting.rv': '{slotConfig.repeat.index}',
 						'targeting.pos': ['incontent_boxad'],
+						'placeholder.createLabel': false,
```
dmnsobczak, line 181: It was inconsistent

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -198,9 +193,6 @@ export class UcpMobileSlotsDefinitionRepository {
 						this.injectIncontentAdsPlaceholders();
 					}
 				});
-				// We need to reset it here, because otherwise ucp-targeting-setup throws an error in line:
-				// https://github.com/Wikia/ad-engine/blob/dev/platforms/shared/context/targeting/ucp-targeting.setup.ts#L101
-				context.set('slots.incontent_player.insertBeforeSelector', '');
```
dmnsobczak, line 203: Fixed, can be removed

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -324,12 +316,6 @@ export class UcpMobileSlotsDefinitionRepository {
 				classList: ['hide', 'ad-slot'],
 			},
 			activator: () => {
-				context.set('slots.floor_adhesion.disabled', !this.instantConfig.get('icFloorAdhesion'));
-				context.set(
-					'templates.floorAdhesion.showCloseButtonAfter',
-					this.instantConfig.get('icFloorAdhesionTimeToCloseButton', 0),
```
dmnsobczak, line 330: Variable not used here

src/ad-engine/models/ad-slot.ts
```diff
@@ -26,7 +26,7 @@ export interface Targeting {
 	[key: string]: googletag.NamedSize | number;
 }
 
-interface RepeatConfig {
+export interface RepeatConfig {
```
dmnsobczak, line 29: Exported to replace object

src/ad-engine/services/slot-creator.ts
```diff
@@ -41,6 +43,10 @@ export class SlotCreator {
 		slotLooseConfig: SlotCreatorConfig,
 		wrapperLooseConfig?: SlotCreatorWrapperConfig,
 	): HTMLElement {
+		if (!slotLooseConfig) {
```
dmnsobczak, line 46: Allow to push slots without config, but with activator

src/ad-engine/services/slot-creator.ts
```diff
@@ -54,6 +60,10 @@ export class SlotCreator {
 			this.addAdLabel(slot.parentElement, slotConfig.slotName);
 		}
 
+		if (slotConfig.repeat) {
```
dmnsobczak, line 63: One place to hold repeat config

src/ad-engine/services/slot-injector.ts
```diff
@@ -39,10 +39,10 @@ class SlotInjector {
 		const config: SlotConfig = context.get(`slots.${slotName}`);
 		const slotConfig: SlotCreatorConfig = {
 			slotName,
-			anchorSelector: config.insertBeforeSelector || config.parentContainerSelector,
-			insertMethod: config.insertBeforeSelector
-				? 'before'
-				: config.insertIntoParentContainerMethod || 'append',
+			anchorSelector: config.parentContainerSelector || config.insertBeforeSelector,
```
dmnsobczak, line 42: Switched to simplify some logic, will be cleaned up here: https://fandom.atlassian.net/browse/ADEN-11453

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
preludek: What is the purpose of this condition?

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
preludek: 💎

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: If page is not wider than 1024px then right rail is hidden. In that case we don't want to load right rail ads, because their viewability will be 0% ;_;

dmnsobczak: In previous code this condition was not included and it should be for all right rail slots.

preludek: Oh yeah, I mistook width for height and couldn't figure it out 🙈 

Great idea 💎

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -247,29 +102,11 @@ export class UcpDesktopDynamicSlotsSetup implements DiProcess {
...
-			this.handleAdPlaceholders(slotName, AdSlot.STATUS_FORCED_COLLAPSE);
-		});
-	}
+	private registerAdPlaceholderService(): void {
+		const placeholderHelper = new PlaceholderServiceHelper();
+		const messageBox = new MessageBox();
 
-	private injectNativeFanFeed(): void {
-		desktopFanFeedNativeAdListener();
+		const placeholderService = new PlaceholderService(placeholderHelper, messageBox);
```
szymon-nawrocki, line 109: I see that you are adding PlaceholderService to FandomDesktop.

I think that is a good idea, because it can be very useful for removing "is-loading" animation from loaded slots.

But it's worth mentioning that that PlaceholderService also adds Message Box in place of collapsed slot.
So far Message Boxes are meant only for FandomMobile platform, so this `if` should not be entered on Desktop platform:
https://github.com/Wikia/ad-engine/blob/afa7af030657865157e495cc77435d3cd85ffc34/platforms/shared/utils/placeholder-service.ts#L49

Because on FandomDesktop we don't need to:
- hide ad label
- hide placeholder
- add message box (yet)

jakubbanas (approved):

platforms/shared/dynamic-slots/curse-slots-definition-repository.ts
```diff
@@ -0,0 +1,32 @@
+import { context } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { SlotSetupDefinition } from '../utils/insert-slots';
+
+@Injectable()
+export class CurseSlotsDefinitionRepository {
```
nandy-andy, line 6: What is Curse? Curse does not exist anymore and the great folks are part of Fandom family :) How about renaming it to `SportsSlotsDefinitionRepository` or `EsportsSlotsDefinitionRepository`?

platforms/shared/utils/insert-slots.ts
```diff
@@ -0,0 +1,25 @@
+import { SlotCreator, SlotCreatorConfig, SlotCreatorWrapperConfig } from '@wikia/ad-engine';
+import { slotsContext } from '../slots/slots-context';
+
+export interface SlotSetupDefinition {
```
nandy-andy, line 4: What do you think of keeping definitions in its own files and having just one class defined per file?

platforms/shared/utils/insert-slots.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const slotCreator = new SlotCreator();
+
+export function insertSlots(slotsToInsert: SlotSetupDefinition[]): void {
+	slotsToInsert
+		.filter((config) => !!config)
+		.forEach(({ slotCreatorConfig, slotCreatorWrapperConfig, activator }) => {
+			try {
+				slotCreator.createSlot(slotCreatorConfig, slotCreatorWrapperConfig);
+				if (activator) {
```
nandy-andy, line 18: Personally, I'd be happy if we'd add some additional logging here in line #14, in the `else` of this condition and in the `catch`, so we'd be able to easily debug what's wrong if a slot we expect to appear does not.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
...
-		const icbSlotName = 'incontent_boxad_1';
-
-		if (context.get('custom.hasFeaturedVideo')) {
-			context.set(`slots.${icbSlotName}.defaultSizes`, [300, 250]);
-		}
-
-		communicationService.action$.pipe(ofType(railReady), take(1)).subscribe(() => {
-			const parent = document.querySelector<HTMLDivElement>(slotConfig.parentContainerSelector);
+	private configureTopLeaderboard(): void {
+		const slotName = 'top_leaderboard';
```
nandy-andy, line 50: I imagine that's quite popular string. How about creating a global const for it and reusing it? 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
-import { slotsContext } from '@platforms/shared';
+import {
+	insertSlots,
+	MessageBox,
```
nandy-andy, line 3: We need to confirm within the team if MessageBoxes are OK to be released on FandomDesktop. It wasn't in the internship plan but maybe it's OK 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
-import { slotsContext } from '@platforms/shared';
+import {
```
nandy-andy, line 1: ❤️ I love how much smaller the file became.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -0,0 +1,243 @@
...
+			return;
+		}
+
+		const slotName = 'incontent_boxad_1';
+
+		return {
+			slotCreatorConfig: {
+				slotName,
+				anchorSelector: '#WikiaAdInContentPlaceHolder',
```
nandy-andy, line 74: `r/PlaceHolder/Placeholder`

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -0,0 +1,243 @@
...
+						'targeting.rv': '{slotConfig.repeat.index}',
+					},
+					insertBelowScrollPosition: false,
+					disablePushOnScroll: true,
+				},
+			},
+			activator: () => {
+				communicationService.action$.pipe(ofType(railReady), take(1)).subscribe(() => {
+					const slotNamePattern = context.get(`slots.${slotName}.repeat.slotNamePattern`);
+					const prefix = slotNamePattern.replace(slotNamePattern.match(/({.*})/g)[0], '');
```
nandy-andy, line 93: How about extracting it to a separate method and unit-test it? :)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -0,0 +1,243 @@
...
+					slotNamePattern: 'incontent_boxad_{slotConfig.repeat.index}',
+					updateProperties: {
+						adProduct: '{slotConfig.slotName}',
+						'targeting.rv': '{slotConfig.repeat.index}',
+					},
+					insertBelowScrollPosition: false,
+					disablePushOnScroll: true,
+				},
+			},
+			activator: () => {
```
nandy-andy, line 90: Lots of indentation here. How about defining the activator function earlier and here just assigning it to the property of the returned literal object?

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -0,0 +1,243 @@
...
+					utils.listener(events.AD_STACK_START, () => {
+						rotator.rotateSlot();
+					});
+				});
+			},
+		};
+	}
+
+	private isRightRailApplicable(): boolean {
+		return utils.getViewportWidth() >= 1024;
```
nandy-andy, line 105: Shall we defined a const for lovely `1024`?

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
nandy-andy: Also, nitpicking but the ICB number is more like a sufix than a prefix 😊

platforms/ucp-mobile/ad-context.ts
nandy-andy: Do I get it correctly? This line means "push request for incontent_player after top_boxad is rendered", correct? And we don't need here `incontent_boxad_1` because we've moved the push in some other place, right? It's still not being requested before top_boxad.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -139,13 +122,15 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 				}
 			});
 
 			eventService.on(events.INTERSTITIAL_DISPLAYED, () => {
 				if (this.CODE_PRIORITY.floor_adhesion.active) {
 					this.CODE_PRIORITY.floor_adhesion.active = false;
 					slotService.disable('floor_adhesion', AdSlot.STATUS_CLOSED_BY_INTERSTITIAL);
+					document.getElementById('floor_adhesion_anchor').classList.add('hide');
```
nandy-andy, line 133: Duplicated code. Please wrap it in a function or something  and call it here and in lien #125

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
nandy-andy: But... we want the labels for incontent_boxads 🤔

platforms/shared/dynamic-slots/curse-slots-definition-repository.ts
```diff
@@ -0,0 +1,32 @@
...
+	constructor() {}
+
+	getCdmZoneConfig(counter: number, lazyLoaded: boolean = false): SlotSetupDefinition {
+		if (counter === 4 && context.get('state.isMobile')) {
```
nandy-andy, line 10: Can you think of some other place we can make the decision and just pass it to the `getCdmZoneConfig()` method as a parameter? This way we won't need dependency of context in such a small module and the method will always return an instance of `SlotSetupDefinition`.

I think the decision if it's a mobile device and therefore we don't want to inject `cdm-zone-04` slot could be made in the dynamic-slots setup and then `getCdmZoneConfig()` will just create the proper `slotName` and return the config. We could also unit test the method. What do you think?

platforms/shared/utils/insert-slots.ts
dmnsobczak: Agree for classes, but `interface` is not a class. It is used only to describe TS types and it is not implemented. For those we are pasting them in the file where there are used on in global type files. This is a change in the unwritten AdEngine coding style, so we can discuss it on the next grooming if you want :D

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
dmnsobczak: <3

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
dmnsobczak: We once decided with Bielik not to use global variables. Also this change would require changes in all platforms using TLB as a first call slot. I would like not add additional "regressive changes" to those, at least in this PR. 
But I think better idea is to start creating some kind of Platform Global Configuration in which `top_leaderboard` will be marked as a "special" slot. We can discuss it with the team.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,73 @@
...
 
-	private createNativeAdPlaceholder(id: string, classList): HTMLElement {
-		const container = document.createElement('div');
-		container.setAttribute('id', id);
-		container.classList.add(...classList);
-
-		return container;
-	}
+	private configureIncontentBoxad(): void {
+		const icbSlotName = 'incontent_boxad_1';
```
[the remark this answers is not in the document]
dmnsobczak: Well, agree, but I tried to create as much patterns as possible, because in the future they are easier to refactor in the bigger batches and personally it's easier for me to read it 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: This requires a change in UCP, because that's where this selector is defined by SiteX. And due to the cache this needs to be done in the 3-steps release cycle (or 2 if we we'll put 2 selectors there for a while). I would like to keep the scope to the AdEngine only. We can do this in the separate ticket: named or pirated (I'll try to remember). Ok?

platforms/ucp-mobile/ad-context.ts
dmnsobczak: Correct. `incontent_boxad_1` is injected manually and `incontent_player` is injected after rendering `top_boxad` and it's creation cycle is handled by SlotInjector used within SlotRepeater :)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
nandy-andy: Sure, good point for a technical refinement 👌 However, this variable in the `configureTopLeaderboard()` is not needed as there is only one place where we need `'top_leaderboard'` and since we don't want to go with any kind of a global variables we can just remove this line and use `top_leaderboard` directly in the no. 61: `context.push('slots.top_leaderboard.defaultTemplates', 'stickyTlb');`

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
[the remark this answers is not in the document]
nandy-andy: Sorry, it's a code-smell - we don't need a variable, so let's not use it. I think you'll get used to it and reading solid strings like: `slots.incontent_boxad_1.defaultSizes` will become as easy as reading strings concatenated with a variable 🤞 A bonus point for using the whole string is a better searching result when instead of a line 67 you'll get the actual line when `incontent_boxad_1` is used.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
dmnsobczak: I think that PlaceholderRepeater is pre-creating all the required labels and we don't want to duplicate them, that's why there is this option for ICB2+. I just copied it and tested, but I remember that @szymon-nawrocki introduced it 🤔

platforms/shared/dynamic-slots/curse-slots-definition-repository.ts
dmnsobczak: Good idea since we removed Gamepedia there's no need to keep `Curse` as `Sports + Gamepedia`. Renamed 👍

platforms/shared/dynamic-slots/curse-slots-definition-repository.ts
dmnsobczak: `context` is still required in this file, because activators are pushing to the ads queue. But I refactored it a bit: I created a list of ids in the dynamic slots file and I'm mapping instead of duplicating lines :)

platforms/shared/utils/insert-slots.ts
dmnsobczak: Agree, I caught myself on that once. Added!

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
dmnsobczak: Thanks for the tip! I disabled MessageBoxes on UCP-Desktop.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: I would rather stick to the code style than breaking it and defining a variable just for one use case. Indentation reaches `6` just for one line and it won't exceed the maximum chars in line limit, so I think it's ok 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: Nooo, pleeeaseee. Let's not write a unit tests to test regular expressions :< They are designed to work, we don't need to double check them with unit tests.
Second thing: it can actually remove it, this logic is overcomplicated 🤔 Done!
Third: it is prefix, we are matching `incontent_boxad_` :D

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: It's not used anywhere else, but sure, why not. It's easier to understand thanks to the variable name. But I left it on the method level :)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
dmnsobczak: Oh, you're right. Refactored!

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -28,7 +24,7 @@ export class F2DynamicSlotsSetup implements DiProcess {
 	private injectSlots(): void {
 		const topLeaderboardDefinition = this.slotsDefinitionRepository.getTopLeaderboardConfig();
```
nandy-andy, line 25: Why do we assign it to a variable if we don't use it later on in more than one place? I think we can freely add `this.slotsDefinitionRepository.getTopLeaderboardConfig()` to the array passed below to `insertSlots()`.

platforms/shared/dynamic-slots/sports-dynamic-slots-setup.service.ts
```diff
@@ -0,0 +1,29 @@
...
+	private uapFirstCallSlotName = 'cdm-zone-01';
+
+	constructor(private slotsDefinitionRepository: SportsSlotsDefinitionRepository) {}
+
+	execute(): void {
+		this.configureUap();
+
+		const slotsIdsToInsert = context.get('state.isMobile') ? [1, 2, 3, 6] : [1, 2, 3, 4, 6];
+		const slotsToInsert = slotsIdsToInsert.map((id) =>
+			this.slotsDefinitionRepository.getCdmZoneConfig(id, id === 4),
```
nandy-andy, line 18: I think it all looks clean and clever now :) However, in the future the context why `cdm-zone-04` is the only lazy-loaded slot on desktop and does not appear on mobile will be missing. I can think of four ways of not loosing the context here:
1. move the information about lazy-loading to the slot context (in `sports-slots-context-setup.service.ts`) and read it from there within `slotsDefinitionRepository.getCdmZoneConfig()`)
2. have two slot context for sports (mobile and desktop) + 1.
3. write unit tests for the parts that the context will be missing in the future - unit tests are quite good documentation of such decision and are really close to the code ;) 
4. write an [ADR](https://medium.com/fandom-engineering/keep-programmers-decisions-documentation-up-to-date-and-simple-with-architecture-decision-records-8ef5f1761ba) - a new trend also quite good as it's as close to the code as possible - in the code repository.

platforms/shared/utils/insert-slots.ts
```diff
@@ -0,0 +1,27 @@
...
+import { slotsContext } from '../slots/slots-context';
+
+export interface SlotSetupDefinition {
+	slotCreatorConfig?: SlotCreatorConfig;
+	slotCreatorWrapperConfig?: SlotCreatorWrapperConfig;
+	activator?: () => void;
+}
+
+const logGroup = 'insert-slot';
+const slotCreator = new SlotCreator();
```
nandy-andy, line 11: Does it mean there is only one `SlotCreator` instance in the run-time?

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
nandy-andy: Oh, actually we can't as in this method we reuse the variable, so it's totally OK to have it 👍

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,72 @@
...
```
nandy-andy, line 66: We don't need this variable. Let's remove it.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
nandy-andy: Damn it! It means we missed it when it got introduced 🙈 Sure, we can try and clean it up in the future.

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
nandy-andy: Oh, my comment wasn't connected to [the style guide](https://github.com/airbnb/javascript). In fact, I can't find there anything about indentation that is closely connected to my thoughts ;-) I was thinking more about the indentation here as a red flag, a signal that something could be written better since the more indentation the less readable a code becomes. It reminds me the callback hells that were a nightmare years ago ;) Then promises came to save us and promises hell has been created :D

So, at the end it's connected with readability and code style but I don't think we have any strict rules about it in the style guide.

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
dmnsobczak: We are using it in L34 to check whether `top_leadeboard` is allowed on this page and slot repository methods are supposed to be executed only once. I'll leave it :)

platforms/shared/dynamic-slots/sports-dynamic-slots-setup.service.ts
dmnsobczak: Two context or expanded array with configuration sounds like a good idea, but let's rethink that in https://fandom.atlassian.net/browse/ADEN-11431, ok? This task can refactor this bundle even more and imho: we should drop the usage of `cdm-zone-slots` and use Fandom slots on Sports, but that's the case for another ticket :<

platforms/shared/utils/insert-slots.ts
dmnsobczak: No, two: https://github.com/Wikia/ad-engine/blob/e520296596701fa05508776001ef0e997ffb9f95/src/ad-engine/services/slot-injector.ts#L11
This instance is only used in this place, but it's fine to have a multiple SlotCreators: it's a worker class without a state. In the future we can make it Injectable() 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
dmnsobczak: Done!

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-dynamic-slots.setup.ts
```diff
@@ -1,112 +1,84 @@
...
+			this.slotsDefinitionRepository.getNativoIncontentAdConfig(),
+			this.slotsDefinitionRepository.getNativoFeedAdConfig(),
+			this.slotsDefinitionRepository.getTopLeaderboardConfig(),
+			this.slotsDefinitionRepository.getBottomLeaderboardConfig(),
+			this.slotsDefinitionRepository.getIncontentPlayerConfig(),
+			this.slotsDefinitionRepository.getFloorAdhesionConfig(),
+			this.slotsDefinitionRepository.getInvisibleHighImpactConfig(),
+		]);
+
+		merge([communicationService.action$.pipe(ofType(railReady)), timer(2000)])
```
nandy-andy, line 52: `UcpDesktopDynamicSlotsSetup.injectSlots()` now has two responsibilities: inserting the slots right away with a call `insertSlots()` and registering inserting TB and ICB on particular two events. Consider splitting the two responsibilities to two different methods.

PR #1207, opened 2021-12-22
ADEN-11431 Joining Muthead and Futhead into Sports

dmnsobczak:


.prettierignore
```diff
@@ -9,6 +9,8 @@ src/ad-services/facebook-pixel/facebook-pixel-script.ts
 src/ad-services/nielsen/static-queue-script.ts
 src/ad-services/permutive/index.ts
 platforms/*/fallback-config.json
+platforms/*/futhead-fallback-config.json
+platforms/*/muthead-fallback-config.json
```
nandy-andy, line 13: It won't catch the files created by the jenkins job: https://github.com/Wikia/jenkins-jobs/pull/1141/files#diff-be5e7664617bd57076f7da06f65430e7003c95d2f4b54138bb88d990c9cc36c4R20-R24

dmnsobczak: Shieeet I forgot about it. Fixed!

.prettierignore
```diff
@@ -9,6 +9,8 @@ src/ad-services/facebook-pixel/facebook-pixel-script.ts
 src/ad-services/nielsen/static-queue-script.ts
 src/ad-services/permutive/index.ts
 platforms/*/fallback-config.json
+platforms/*/fallback-config-futhead.json
+platforms/*/fallback-config-muthead.json
```
nandy-andy, line 13: I also wonder if `platforms/*/fallback-config*.json` or `platforms/*/fallback-config-*.json` work or if it's too much and it's better to be more explicit in terms of futhead/muthead 🤔 What do you think?

platforms/sports/ad-context.ts
```diff
@@ -45,18 +61,41 @@ export const basicContext = {
...
+			endpoint: 'https://services.fandom.com/icbm/api/config?app=muthead',
+			fallback: mutheadFallbackInstantConfig,
+		},
 	},
-	templates: {},
 };
+
+export function getBasicContext(app): object {
+	if (app === 'futhead') {
+		Object.assign(basicContext, futheadContext);
```
nandy-andy, line 93: Don't we want to use here early return? Or assign `Object.assign()` results to the `basicContext` that we return in line no. 100?

platforms/sports/templates/logo-replacement-template.ts
```diff
@@ -1,14 +1,18 @@
...
 
 export function registerLogoReplacementTemplate(
 	registry: TemplateRegistry,
 ): Observable<TemplateAction> {
 	return registry.register(
 		'logoReplacement',
 		{
-			initial: [LogoReplacementFutheadHandler],
+			initial:
+				context.get('application') === 'futhead'
```
nandy-andy, line 13: We've used it 4 times in this PR - it's a good signal to create `isFuthreadApp()` helper function ;)

.prettierignore
dmnsobczak: Agree. Changed!

platforms/sports/ad-context.ts
dmnsobczak: `Object.assign()` updates the first object passed to it, so it's good, but I made a one liner out of it :D

platforms/sports/templates/logo-replacement-template.ts
dmnsobczak: Created!

platforms/sports/setup/context/targeting/sports-targeting.setup.ts
```diff
@@ -32,7 +32,23 @@ export class FutheadTargetingSetup implements DiProcess {
 	}
 
 	private isSquadPage(): boolean {
+		if (context.get('application') === 'muthead') {
```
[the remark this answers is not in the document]
dmnsobczak: I refactored the code a little bit. I created a helper method `selectApplication()` :)

.prettierignore
nandy-andy: Is it? 🤔

platforms/sports/setup/context/prebid/sports-prebid-config.setup.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getRubiconMutheadContext } from '../../../bidders/prebid/rubicon-muthead';
+import { selectApplication } from '../../../utils/application-helper';
+
+@Injectable()
+export class SportsPrebidConfigSetup implements DiProcess {
+	execute(): void {
+		const mode: DeviceMode = getDeviceMode();
+
+		context.set(
+			'bidders.prebid.rubicon_display',
```
nandy-andy, line 14: What would happen when`null` will be returned and set to `bidders.prebid.rubicon_display`?

platforms/sports/utils/application-helper.ts
```diff
@@ -0,0 +1,23 @@
...
+export function selectApplication(futhead: any = null, muthead: any = null): any | null {
+	if (getApplication() === 'futhead') {
+		return futhead;
+	}
+
+	if (getApplication() === 'muthead') {
+		return muthead;
+	}
+
+	return null;
```
nandy-andy, line 22: Let's throw here an error to the console or at least warning just-in-case in some weird setup (localhost or devbox, etc.) `selectApplication()` will start returning `null`. I think in many cases we may not even notice it (like adding class to the `body` element or not extending base context), so a nice red error or orange warning will be a clear signal something is off ;)

platforms/sports/setup/context/prebid/sports-prebid-config.setup.ts
dmnsobczak: Hmm, I'll follow your suggestion and I will throw an error for this path 🤔

platforms/sports/utils/application-helper.ts
dmnsobczak: Done!

nandy-andy (approved):

PR #1208, opened 2021-12-22
ADEN-11446 types and events

preludek:


src/ad-services/nativo/index.ts
```diff
@@ -19,6 +19,7 @@ export class Nativo {
 	call(): Promise<void> {
 		if (!this.isEnabled()) {
 			utils.logger(logGroup, 'disabled');
+			this.sendEvent(false);
```
nandy-andy, line 22: What do you think of renaming the method so it's more descriptive. For example to `sendLoadedEvent()` because `sendCommunicationServiceLoadedEvent()` is probably too long? 🙈 

While reading the code up to this point I thought: "Oh, OK, we're sending an event but what event? And what does the boolean value suppose to mean?". I needed to scroll down and I got it - it's easy now to just pass the boolean and place it in the literal object passed to `nativoLoadedEvent()` but without scrolling down it was hard to guess and I chose to suggest just renaming as a first step to make it more clear. Maybe also passing `nativoLoadedEvent({ isLoaded })` instead of boolean will make it even more clear for somebody who'll be reading just the `call()` method 🤔

preludek: @nandy-andy  renaming this is definitely a good idea.  👍 

I was thinking of either:
- `sendLoadStatusEvent` and passing a boolean
- generalizing the whole event to `sendNativoStatus`. `[AdEngine] Nativo` and pass everything in props as config. 

Any opinions? ;)

nandy-andy: Second option sounds like a really clean solution 👍

preludek: 👍  I went with something similar that we are doing in UAP sending it's load status.

nandy-andy (approved):

nandy-andy (approved):

dmnsobczak (approved):

src/ad-services/nativo/index.ts
```diff
@@ -98,8 +98,8 @@ export class Nativo {
 		return false;
 	}
 
-	private sendEvent(): void {
-		communicationService.dispatch(nativoLoadedEvent({ isLoaded: true }));
+	private sendNativoLoadStatus(event): void {
+		communicationService.dispatch(adSlotEvent({ event, adSlotName: logGroup }));
```
nandy-andy, line 102: Are we sure we want to assign `logGroup` to the `adSlotName`? 🤔

nandy-andy (changes requested):
Something seems off in the `sendNativoLoadStatus(event)`

spec/ad-services/nativo/nativo.spec.ts
```diff
@@ -1,12 +1,13 @@
...
 import { communicationService } from '@wikia/communication';
 import { expect } from 'chai';
 import { createSandbox, SinonSpy } from 'sinon';
-import { context, utils } from '../../../src/ad-engine';
+import { AdSlot, adSlotEvent, context, utils } from '../../../src/ad-engine';
 import { createHtmlElementStub } from '../../spec-utils/html-element.stub';
 
 describe('Nativo service', () => {
 	const sandbox = createSandbox();
+	const slotName = 'nativo';
```
nandy-andy, line 10: I think this is not enough. We'd need to distinguish Nativo in-content slots and in the Fan Feed.

src/ad-services/nativo/index.ts
preludek: it's conveniently the same string 🤷‍♂️

nandy-andy: That's a horrible code smell ;p

spec/ad-services/nativo/nativo.spec.ts
preludek: done

src/ad-services/nativo/index.ts
preludek: 🙈  I've changed it

nandy-andy (approved):

nandy-andy:
What do you think of changing the base to [the "hold-on" branch](https://github.com/Wikia/ad-engine/tree/ADEN-AdEngine-100)? ;)

PR #1209, opened 2021-12-22
ADEN-11394 Remove Permutive integration

preludek:


spec/ad-bidders/prebid/adapters/appnexus-ast.spec.ts
```diff
@@ -54,44 +53,4 @@ describe('AppnexusAst bidder adapter', () => {
 			},
 		]);
 	});
-
-	it('prepareAdUnits returns data in correct shape with additional key-vals', () => {
```
preludek, line 58: This unit test is identical to the one above.

src/ad-bidders/prebid/adapters/appnexus.ts
```diff
@@ -34,7 +34,6 @@ export class Appnexus extends PrebidAdapter {
 						placementId: placementId || this.getPlacement(position),
 						keywords: {
 							...this.getTargeting(code),
-							p_standard: context.get('bidders.permutiveKeys.appnexus') || [],
```
preludek, line 37: I'm not 100% sure if this can be removed. It seems it can, since we are not setting anything there without permutive. @dmnsobczak can you confirm? :)

dmnsobczak: Yes it can :)

dmnsobczak (approved):

PR #1214, opened 2021-12-28
ADEN-11351 Replacing EventEmitter with Post-Quecast

dmnsobczak:


src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
 		this.connectReduxDevtools(reduxDevtoolsName);
 	}
 
+	communicate(event: EventOptions, payload?: object): void {
+		this.dispatch(this.getGlobalAction(event)(payload));
+	}
+
+	listen(event: EventOptions, callback: (payload?: any) => void, once: boolean = true): void {
+		this.action$
+			.pipe(ofType(this.getGlobalAction(event)), once ? take(1) : skip(0))
```
preludek, line 51: We can do this conditional with `&&` operator and won't need skip

src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
...
+		once: boolean = false,
+	): void {
+		this.action$
+			.pipe(
+				ofType(this.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)),
+				filter(
+					(action) =>
+						action.event === eventName.toString() && (!slotName || action.adSlotName === slotName),
+				),
+				once ? take(1) : skip(0),
```
preludek, line 68: ditto

src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
...
+	listenSlotEvent(
+		eventName: string | symbol,
+		callback: (payload?: any) => void,
+		slotName: string = '',
+		once: boolean = false,
+	): void {
+		this.action$
+			.pipe(
+				ofType(this.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)),
+				filter(
```
preludek, line 64: prettier pls

src/ad-products/templates/outstream/porvata-template.ts
```diff
@@ -100,10 +100,6 @@ export class PorvataTemplate {
 
 		this.handleSlotStatus(player);
 
-		eventService.once(events.PAGE_CHANGE_EVENT, () => {
```
preludek, line 103: Shouldn't we set a listener here to destroy the player?

preludek (changes requested):
Did you consider writing some unit tests for new methods in `communication-service`?

package.json
```diff
@@ -19,7 +19,6 @@
 		"core-decorators": "^0.20.0",
 		"current-device": "^0.7.8",
 		"design-system": "git+https://github.com/Wikia/design-system.git#22.0.0",
-		"eventemitter3": "^3.0.1",
```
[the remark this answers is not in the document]
dmnsobczak: <3

src/communication/event-types.ts
```diff
@@ -0,0 +1,218 @@
...
+	adProduct: string;
+}
+
+export const eventsRepository = {
+	// AdEngine life cycle events //
+	AD_ENGINE_BAB_DETECTION: {
+		category: '[Ad Engine]',
+		name: 'BAB detection finished',
+		payload: props<{ detected: boolean }>(),
+	} as EventOptions,
```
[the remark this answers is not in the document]
dmnsobczak: Yes, but when we mark it as `Dictionary<EventOptions>` or `{ [key: string]: EventOptions }` then IDE won't treat it as a constant and won't give any suggestions if you will type `eventsRepository.` :/ 
I reverted this commit, but we can work on those types in the future.

src/communication/event-types.ts
```diff
@@ -0,0 +1,218 @@
+import { AdSlot } from '@ad-engine/core';
+// tslint:disable-next-line:import-blacklist
```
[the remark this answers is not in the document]
dmnsobczak: Actually I don't know. Maybe it's because this module is imported from npm dependency and that's why it can't be added to the `.tsconfig`? 🤔 @Bielik20 ca you provide some details?

src/communication/event-types.ts
```diff
@@ -0,0 +1,218 @@
...
+	size: string;
+	timeToRespond: number;
+}
+
+export interface UapLoadStatus {
+	isLoaded: boolean;
+	adProduct: string;
+}
+
+export const eventsRepository = {
```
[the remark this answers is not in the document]
dmnsobczak: Agree, but I consider this PR as a first step. I wanted to make it as simple as possible, because removing EventEmitter and switching fully to Post-Quecast is, as you can see, already a big change. Once we test ad merge it we can improve this mechanism in smaller steps in the future. For now I think `eventRepository` fits it's limit and it definitely will be splitted with the future events or some more sophisticated events control system.

src/communication/communication-service.ts
dmnsobczak: xD

src/ad-products/templates/outstream/porvata-template.ts
dmnsobczak: Destroying a player was used on mobile-wiki legacy where between pageviews there was a single AdEngine instance which had to be reconfigured for a different slots. Now all our apps are not a Single Page Apps and nothing is emitting a Page_Change_Event, so I removed all SPA-related logic, including this line :)

src/communication/event-types.ts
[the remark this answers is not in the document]
Bielik20: I used it for 2 things:

1. Enforce separation between AdEngine submodules. For example everyone can import from "core" module, but it cannot import from other modules. This is to prevent circular dependencies between modules and set clear boundaries. 
2. (this example) We do not want to use `post-quecast` directly in the code. We have our own wrapper (I don't remember the name... CommunicationService?) and we want to use that instead. So to prevent users from accidentally importing `post-quecast` itself, this rule was introduced.

src/communication/communication-service.ts
dmnsobczak: Great tip! Changed!

src/communication/communication-service.ts
dmnsobczak: Done!

src/communication/event-types.ts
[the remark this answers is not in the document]
dmnsobczak: Got it! Thank you 🙇

preludek (approved):

src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
...
+		this.dispatch(this.getGlobalAction(event)(payload));
+	}
+
+	listen(event: EventOptions, callback: (payload?: any) => void, once: boolean = true): void {
+		this.action$
+			.pipe(ofType(this.getGlobalAction(event)), once ? take(1) : skip(0))
+			.subscribe(callback);
+	}
+
+	listenSlotEvent(
```
mfarul, line 55: listenOnSlotEven

src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
 		this.connectReduxDevtools(reduxDevtoolsName);
 	}
 
+	communicate(event: EventOptions, payload?: object): void {
+		this.dispatch(this.getGlobalAction(event)(payload));
+	}
+
+	listen(event: EventOptions, callback: (payload?: any) => void, once: boolean = true): void {
```
bckatarzyna, line 49: Wouldn't `listenOn` be better? Or just `on`?

bckatarzyna (approved):

package.json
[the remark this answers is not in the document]
nandy-andy: ❤️

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -32,14 +30,12 @@ export class F2DynamicSlotsSetup implements DiProcess {
 		]);
 
 		if (!topLeaderboardDefinition) {
-			utils.listener(events.AD_STACK_START, () => {
+			communicationService.listen(eventsRepository.AD_ENGINE_STACK_START, () => {
```
nandy-andy, line 33: What did motivate you to rename the event from `AD_STACK_START` to `AD_ENGINE_STACK_START`?

platforms/shared/setup/ad-engine-runner.setup.ts
```diff
@@ -8,9 +14,13 @@ export class AdEngineRunnerSetup implements DiProcess {
...
 
 	execute(): void {
-		eventService.on(events.AD_SLOT_CREATED, (slot) => {
-			utils.logger(logGroup, `Created ad slot ${slot.getSlotName()}`);
-			bidders.updateSlotTargeting(slot.getSlotName());
-		});
+		communicationService.listen(
+			eventsRepository.AD_ENGINE_SLOT_ADDED,
+			({ slot }) => {
+				utils.logger(logGroup, `Created ad slot ${slot.getSlotName()}`);
```
nandy-andy, line 20: I'd change the log message to `'Added ad slot...'` since we don't listen to `AD_SLOT_CREATE` anymore.

platforms/shared/utils/collapsed-messages/message-box.ts
```diff
@@ -51,12 +51,10 @@ export class MessageBox {
 	};
 
 	sendTrackingEvent = (adSlot: AdSlot, ad_status: string) => {
-		communicationService.dispatch(
-			messageBoxTrackingEvent({
-				ad_status,
-				adSlotName: adSlot.getSlotName(),
-			}),
-		);
+		communicationService.communicate(eventsRepository.AD_ENGINE_MESSAGEBOX_EVENT, {
```
nandy-andy, line 54: There is no word like messagebox, so I'd go with: `r/AD_ENGINE_MESSAGEBOX_EVENT/AD_ENGINE_MESSAGE_BOX_EVENT`

platforms/sports/templates/handlers/logo-replacement/logo-replacement-futhead-handler.ts
```diff
@@ -34,8 +31,6 @@ export class LogoReplacementFutheadHandler implements TemplateStateHandler {
 					parentElement.appendChild(trackingPixel);
 					customLogoAnchorElement.appendChild(customLogo);
 				}
-
-				this.adSlot.emitEvent(events.LOGO_REPLACED);
```
nandy-andy, line 38: Don't we want to replace it with some event from the `eventRepository`?

platforms/sports/templates/handlers/logo-replacement/logo-replacement-muthead-handler.ts
```diff
@@ -32,8 +29,6 @@ export class LogoReplacementMutheadHandler implements TemplateStateHandler {
 				parentElement.appendChild(trackingPixel);
 				customLogoAnchorElement.appendChild(customLogo);
 			}
-
-			this.adSlot.emitEvent(events.LOGO_REPLACED);
```
nandy-andy, line 36: Ditto.

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -169,12 +169,16 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 			},
 		};
 
-		eventService.on(events.AD_SLOT_CREATED, (slot) => {
-			context.onChange(`slots.${slot.getSlotName()}.audio`, () => this.setupSlotParameters(slot));
-			context.onChange(`slots.${slot.getSlotName()}.videoDepth`, () =>
-				this.setupSlotParameters(slot),
-			);
-		});
+		communicationService.listen(
```
nandy-andy, line 172: This is the same code as in f2 setup: https://github.com/Wikia/ad-engine/pull/1214/files#diff-ccba3e1af310afba5c74877c647d6d03e0b930d8721f2ae3010e041677c739c1R185 What do you think of moving it to shared?

platforms/ucp-mobile/setup/context/slots/ucp-mobile-slots-context.setup.ts
```diff
@@ -256,12 +256,16 @@ export class UcpMobileSlotsContextSetup implements DiProcess {
 			},
 		};
 
-		eventService.on(events.AD_SLOT_CREATED, (slot) => {
-			context.onChange(`slots.${slot.getSlotName()}.audio`, () => this.setupSlotParameters(slot));
-			context.onChange(`slots.${slot.getSlotName()}.videoDepth`, () =>
-				this.setupSlotParameters(slot),
-			);
-		});
+		communicationService.listen(
```
nandy-andy, line 259: This is the same logic we use in f2 and ucp-desktop. I think it's a good candidate to move it to shared :)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -84,9 +80,13 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
 	private configureInterstitial(): void {
 		const slotName = 'interstitial';
```
nandy-andy, line 81: We don't need this variable here ;)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -115,31 +115,45 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
-		slotService.on('floor_adhesion', AdSlot.STATUS_SUCCESS, () => {
-			this.CODE_PRIORITY.floor_adhesion.active = true;
+		const slotName = 'floor_adhesion';
+
+		communicationService.listenSlotEvent(
+			AdSlot.STATUS_SUCCESS,
+			() => {
+				this.CODE_PRIORITY.floor_adhesion.active = true;
+
+				const disableFloorAdhesionWithStatus = (status: string) => {
```
nandy-andy, line 125: What do you think of moving the definition of `disableFloorAdhesionWithStatus()` to the line 119? It should make the lines 120-148 a bit more readable.

spec/ad-engine/models/ad-slot.spec.ts
```diff
@@ -229,64 +229,42 @@ describe('ad-slot', () => {
...
-			adSlot.emit(symbol, 'foo', 'bar');
-
-			expect(emitSpy.callCount).to.equal(1);
-			expect(emitSpy.firstCall.args[0]).to.equal(symbol);
-			expect(emitSpy.firstCall.args[1]).to.equal(adSlot);
-			expect(emitSpy.firstCall.args[2]).to.equal('foo');
-			expect(emitSpy.firstCall.args[3]).to.equal('bar');
-		});
-
 		it('should call eventService.communicator.dispatch with string event', () => {
```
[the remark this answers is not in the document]
nandy-andy: Yep, consider replacing this and following unit tests with something general for the `communicationService` integration.

spec/ad-engine/ad-slot-fake.ts
```diff
@@ -98,4 +97,11 @@ export default {
 	getSlotsToInjectAfterRendered() {
 		return [];
 	},
+	emit(event) {
```
nandy-andy, line 100: Shouldn't we have a similar method in the real AdSlot, so it's not confusing for anyone reading the unit tests?

src/ad-engine/models/ad-slot.ts
```diff
@@ -189,15 +146,66 @@ export class AdSlot extends EventEmitter {
...
+
+					this.updateOnRenderEnd(event, adType);
+
+					resolve();
+				},
+				this.getSlotName(),
+				true,
+			);
+		});
+		this.viewed = new Promise<void>((resolve) => {
```
nandy-andy, line 190: What do you think of defining the promises elsewhere and also adding `catch()` to them with some logging in the browser's console?

src/ad-engine/models/ad-slot.ts
```diff
@@ -440,16 +448,16 @@ export class AdSlot extends EventEmitter {
 			templateNames.forEach((templateName: string) => templateService.init(templateName, this));
 		}
 
-		this.emit(AdSlot.TEMPLATES_LOADED, ...templateNames);
+		this.emit(AdSlot.TEMPLATES_LOADED, templateNames);
```
nandy-andy, line 451: I assume we were copying `templateNames` in order to avoid unexpected mutations. Are you sure it's safe to just pass it as an argument?

spec/ad-engine/ad-slot-fake.ts
nandy-andy: Nevermind, I've just found it: https://github.com/Wikia/ad-engine/pull/1214/files#diff-18f1c6cfa4982c404779ce1c6c6021cf2da725f5f31b9d165393f4ef5d8816f8R596 :D

src/ad-engine/models/ad-slot.ts
```diff
@@ -583,17 +591,17 @@ export class AdSlot extends EventEmitter {
...
+	 * Pass all events to Post-QueCast
 	 */
-	emit(event: string | symbol, ...args: any[]): boolean {
-		const result = super.emit(event, ...args);
-
-		eventService.emit(event, this, ...args);
-		this.emitPostQueueCast(event, args);
-
-		this.logger(this.getSlotName(), event, result, ...args);
+	emit(event: string | symbol, data: any = {}): void {
```
nandy-andy, line 596: Are you sure we want to just pass the `data` object? We used to copy it with `JSON.parse(JSON.stringify())` in the previous `emitPostQueueCast()` implementation 🤔

src/ad-engine/services/template-service.ts
```diff
@@ -71,7 +65,7 @@ class TemplateService {
 	subscribeCommunicator(): void {
 		communicationService.action$
 			.pipe(
-				ofType(loadTemplate),
+				ofType(communicationService.getGlobalAction(eventsRepository.GAM_LOAD_TEMPLATE)),
```
nandy-andy, line 68: Nice idea of informing that the template comes from GAM 👌 I'd suggest just changing its place, so it sounds better: `LOAD_GAM_TEMPLATE` or `LOADED_GAM_TEMPLATE`. What do you think?

src/ad-engine/tracking/intervention-tracker.ts
```diff
@@ -24,7 +21,7 @@ class InterventionTracker {
 
 		communicationService.action$
 			.pipe(
-				ofType(adIntervention),
+				ofType(communicationService.getGlobalAction(eventsRepository.GAM_AD_INTERVENTION)),
```
nandy-andy, line 24: The GAM part here seems misleading and unnecessary. It's not GAM that blocks the ad - it's the browser.

src/communication/communication-service.ts
nandy-andy: I had similar thoughts ;) My thoughts were leaning towards leaving just `on` as when reading the code we read "on something do something" or going with `listenTo` as it's more grammatically correct, I guess.

src/communication/event-types.ts
```diff
@@ -0,0 +1,218 @@
...
+	size: string;
+	timeToRespond: number;
+}
+
+export interface UapLoadStatus {
+	isLoaded: boolean;
+	adProduct: string;
+}
+
+export const eventsRepository: Dictionary<EventOptions> = {
```
nandy-andy, line 53: Hooray! One place for all the events definitions ❤️

src/communication/communication-service.ts
```diff
@@ -40,6 +42,34 @@ export class CommunicationService {
 		this.connectReduxDevtools(reduxDevtoolsName);
 	}
 
+	communicate(event: EventOptions, payload?: object): void {
```
nandy-andy, line 45: Shouldn't we make `dispatch()` private if we want to use `communicate()` now?

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
dmnsobczak: I wanted all internal AdEngine events to start with `AD_ENGINE_...`, because their name also starts like that: `[AdEngine] Something...`. It looks cleaner and this event is only used by us, so we can rename it however we like :D

platforms/shared/setup/ad-engine-runner.setup.ts
dmnsobczak: I wanted to limit the amount of non-event related code changes to minimum to limit the testing scope, especially I don't want to change log messages, because somebody might be used to them like that. Creating an ad slot is a synonym of adding if to the slotService or to ad stack.

platforms/sports/templates/handlers/logo-replacement/logo-replacement-futhead-handler.ts
dmnsobczak: No, this event was not used anywhere - there's no need to emit events that are not listened anywhere 🤔

platforms/sports/templates/handlers/logo-replacement/logo-replacement-muthead-handler.ts
dmnsobczak: Replied above.

src/ad-engine/services/template-service.ts
dmnsobczak: It's not a event about loading a template. It's a event with a data from GAM - it is dispatched by GAM Post-Quecast creative wrapper and it starts with `[GAM] Something...` - that's why `GAM` is on the first place. I want to keep the first word in the event identifier a uppercase event category: `[GAM]` === `GAM_`, `[AdEngine]` === `AD_ENGINE_` etc :)

src/ad-engine/tracking/intervention-tracker.ts
dmnsobczak: Yes, but the event comes from GAM Post-Quecast instance - that's what `GAM_` states. I would like to keep it that way.

src/communication/event-types.ts
dmnsobczak: <3

dmnsobczak: But this file will be splitted in the future when it will grow up :D

platforms/shared/utils/collapsed-messages/message-box.ts
dmnsobczak: Done! :grammar: 🙇

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
dmnsobczak: Good idea. Moved to `slotsContext`!

platforms/ucp-mobile/setup/context/slots/ucp-mobile-slots-context.setup.ts
dmnsobczak: Done!

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
dmnsobczak: Removed!

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
dmnsobczak: But then this method (or actually "decorator") will be created always, not only when FloorAdhesion gets `success` and there is a real chance to use this code prio logic. We don't need to allocate a memory for that. Small improvement, but still :D I'll keep it in the local scope.

spec/ad-engine/models/ad-slot.spec.ts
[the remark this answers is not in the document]
dmnsobczak: Renamed!
Also: `communicationService` tests are spread across multiple modules since events are really important mechanism. I know that coverage is never enough, but I think it's high enough for now 🤔

spec/ad-engine/models/ad-slot.spec.ts
```diff
@@ -229,64 +229,42 @@ describe('ad-slot', () => {
...
+				communicationService.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)({
 					payload: ['foo', 'bar'],
 					event: AdSlot.TEMPLATES_LOADED,
 					adSlotName: slotName,
+					slot: adSlot,
 				}),
 			);
 		});
 
 		it('should call eventService.communicator.dispatch with Symbol event', () => {
```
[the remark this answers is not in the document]
dmnsobczak: Done!

src/ad-engine/models/ad-slot.ts
dmnsobczak: Those promises are like a semaphores and they are ultra fragile - I spent a lot of time fixing them after changing event system to make all ads load and track correctly. I would rather create a separate ticket for that. Tomek has a solid ideas and motivation to split AdSlot Mega Class into smaller chunks - let's postpone this idea until then :)
Also nothing is emitting any exception there and an exception would lead to a massive AdEngine crash - this can't be just logged and forget, this part needs to be designed with an idea to omit any exceptions 🤔

src/ad-engine/models/ad-slot.ts
dmnsobczak: Good idea. I don't think anybody will edit this event payload which, like you observed, will be a context reference, but a simple `{ ...var }` usage is a solid protection :D

dmnsobczak: Changing mind: fixed in the `emit()` method.

src/ad-engine/models/ad-slot.ts
dmnsobczak: Hmm, good thinking 🤔 I will restore a serialization-deserialization!

dmnsobczak: One place where I can't use it is a renderEnded GPT event - there are some important data sent there. I added optional, default parameter.

src/communication/communication-service.ts
dmnsobczak: `communicate()` is just a wrapper for a 95% of events, but some of them are more complex, have complicated types and use some edge-case RxJS stream parameters. For those I left `dispatch()` for manual use. We probably will get rid of them in the future: for example types in `tracking-opt-in-wrapper` will be removed after moving Post-Quecast events to TrackingOptIn repository in different ticket :)

dmnsobczak: I changed a method name to `.on()`

src/communication/communication-service.ts
dmnsobczak: Less letters is better - changed to `.on()` :)

src/communication/communication-service.ts
dmnsobczak: Changed to `.onSlotEvent()`!

src/communication/event-types.ts
```diff
@@ -0,0 +1,218 @@
...
+		category: '[Video]',
+		name: 'Video event',
+		payload: props<{ videoEvent: object }>(),
+	},
+	VIDEO_PLAYER_TRACKING: {
+		category: '[Video]',
+		name: 'Video player tracking',
+		payload: props<{ eventInfo: object }>(),
+	},
+	// Google Ad Manager events //
```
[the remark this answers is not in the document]
dmnsobczak: I changed the comment to: `// Events emitted from Google Ad Manager //`, because that's their origin :)

LICENSE
```diff
@@ -38,7 +38,7 @@
       (an example is provided in the Appendix below).
 
       "Derivative Works" shall mean any work, whether in Source or Object
-      form, that is based on (or derived from) the Work and for which the
+      form, that is based emit (or derived from) the Work and for which the
```
nandy-andy, line 41: I don't think you need to change the LICENSE :p

README.md
```diff
@@ -8,7 +8,7 @@ npm install github:Wikia/ad-engine
 
 ## Platforms
 
-* `npm run dev:platforms` - to serve ad-engine platforms code (UCP, Futhead, Muthead) on port 9000
+* `npm run dev:platforms` - to serve ad-engine platforms code (UCP, Futhead, Muthead) emit port 9000
```
nandy-andy, line 11: Nope, `on` was fine here ;)

platforms/shared/setup/ad-engine-runner.setup.ts
nandy-andy: But it's not inconsistent and I'm afraid only you treat creating and ad slot and adding an ad slot as synonyms 🙈 I'd say adding is a synonym for injecting ;) So, we need to figure it out first but in the spirit of consistency the log message should be consistent with the event. Otherwise, it's misleading.

src/ad-engine/models/ad-slot.ts
nandy-andy: If such massive crash happens with a log message we'd now the place when it happens, won't we? :) Please, create a follow-up ticket and let's groom it this or next week.

src/communication/event-types.ts
```diff
@@ -204,7 +204,7 @@ export const eventsRepository: Dictionary<EventOptions> = {
 		name: 'Video player tracking',
 		payload: props<{ eventInfo: object }>(),
 	},
-	// Google Ad Manager events //
+	// Events emitted from Google Ad Manager //
```
nandy-andy, line 207: That's not true. The events are emitted from the iframe created by GPT ;p GAM does not emit anything on our pages ;)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
nandy-andy: Code readability > premature optimalisation ;)

README.md
dmnsobczak: Arrrghhhhm fixed :<

LICENSE
dmnsobczak: Damn, fixed too.

platforms/shared/setup/ad-engine-runner.setup.ts
dmnsobczak: Ooook, changed :D

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
dmnsobczak: Ok, done :D

src/communication/event-types.ts
dmnsobczak: How about `Events emitted by Google Ad Manager creatives`? :D It's just a comment :>

nandy-andy (approved):

nandy-andy (approved):

PR #1215, opened 2021-12-28
ADEN-11413 Add newsletter signup form when collapse appears on page

szymon-nawrocki:
### Ticket:
https://fandom.atlassian.net/browse/ADEN-11413


platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,80 @@
...
+			vars: {
+				sign_up_date: new Date().toISOString(),
+				email: emailValue,
+			},
+			lists: { 'Fandom Account Registration - MASTERLIST': 1 },
+			source: 'homepage_hero_unit',
+			onSuccess: () => {
+				this.showFormMessage('Thanks for signing up!');
+				submitBtn.disabled = false;
+				this.sendTrackingEvent(adSlot, status_clicked);
```
szymon-nawrocki, line 72: I wonder if sending _adengadinfo_ event is necessary here, because these clicks can be probably tracked in _Sailthru_  🤔

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,24 @@
+import { utils } from '@ad-engine/core';
+
+const ACCT_ID = 'c9322e218da57e71a4965ae18fecbefa';
```
szymon-nawrocki, line 3: I am also wondering if I can use this _ACCT_ID_, because I took it straight from Newsletter Form implementation on f2, so maybe it can distort their data or something.
@preludek Do you think that I should consult it with someone from f2 first?

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,80 @@
...
+			<button class="newsletter-submit wds-button cm-button" type="submit">${this.buttonText}</button>
+		`;
+
+		return form;
+	};
+
+	private createFormMessage = (): HTMLElement => {
+		const message = document.createElement('div');
+		message.className = 'newsletter-message';
+		message.innerHTML = '';
```
preludek, line 42: is this necessary?

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,80 @@
...
+		return message;
+	};
+
+	private showFormMessage = (text: string) => {
+		const messageArea = document.querySelector('.newsletter-message');
+		messageArea.innerHTML = text;
```
preludek, line 48: Can we use innerText instead?

preludek (changes requested):

src/ad-services/sailthru/index.ts
preludek: We should ask somebody if we can use that. Also, ACCT_ID may sometimes differ for different platforms.

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
szymon-nawrocki: You're right, I don't think we have any access there.
And since we are tracking other message boxes with these events I will leave it as it is

src/ad-services/sailthru/index.ts
szymon-nawrocki: @mklucsarits
I see that you were implementing Newsletter sign-up in f2 platform (https://github.com/Wikia/f2/pull/2413)

I've got the task of copying this Newsletter sign-up and serve it into collapsed ad slots, which I did.
So far everything seems to be working:
- Sailthru script is correctly added to the site,
- we're getting 201 responses on sign ups with valid email address
But I am not really sure about this _ACCT_ID_ variable.

Could you tell me if I can use the same `ACCT_ID` that is being used in f2 platform?

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,24 @@
...
+const logGroup = 'sailthru';
+export const packageUrl = 'https://ak.sail-horizon.com/spm/spm.v1.min.js';
+
+function loadScript(path: string): Promise<Event> {
+	return utils.scriptLoader.loadScript(path, 'text/javascript', true, 'first');
```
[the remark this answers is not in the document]
szymon-nawrocki: You're right, we don't really need "first" parameter here

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
szymon-nawrocki: not really

src/ad-services/sailthru/index.ts
szymon-nawrocki: It turns out that when AdEngine is loaded Sailthru is already there, so there is no need to initialize it a second time

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+class Sailthru {
+	call(): Promise<void> {
```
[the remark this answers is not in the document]
szymon-nawrocki: When AdEngine is loaded Sailthru is already there, so there is no need for ICBM flag.

src/ad-services/sailthru/index.ts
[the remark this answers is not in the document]
szymon-nawrocki: icVar won't be implemented

src/ad-services/sailthru/index.ts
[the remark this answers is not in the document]
szymon-nawrocki: Turns out that Sailthru is being initiated on the backend, and only one of its method will be used in the AdEngine: `Sailthru.integration()`
So I think that flag is not super necessary in this case.

src/ad-services/sailthru/index.ts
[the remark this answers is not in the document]
szymon-nawrocki: It was decided that icVar won't be implemented

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,84 @@
...
+		const submitBtn: HTMLButtonElement = document.querySelector('.newsletter-submit');
+		submitBtn.disabled = true;
+
+		if (!sailthru.isLoaded()) {
+			this.showFormMessage('An error occurred. Please try again later.');
+			submitBtn.disabled = false;
+			return;
+		}
+
+		window.Sailthru.integration('userSignUp', {
```
preludek, line 65: This should be moved to Sailthru integration. Let's keep all calls to `window.Sailthru` in one place.

preludek (changes requested):

platforms/shared/utils/collapsed-messages/message-box-service.ts
```diff
@@ -1,10 +1,10 @@
 import { AdSlot } from '@wikia/ad-engine';
 import { MessageBoxCreator } from './message-box-creator';
 
-export type MessageBoxType = 'REGISTER' | 'FANLAB' | 'NEWSLETTER';
+export type MessageBoxType = 'REGISTER' | 'FANLAB' | 'NEWSLETTER_LINK' | 'NEWSLETTER_FORM';
 
 export class MessageBoxService {
-	private types: MessageBoxType[] = ['REGISTER', 'FANLAB', 'NEWSLETTER'];
+	private types: MessageBoxType[] = ['REGISTER', 'FANLAB', 'NEWSLETTER_FORM', 'NEWSLETTER_LINK'];
```
nandy-andy, line 7: Do we need both versions of `MessageBoxType`?

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,84 @@
...
+			email: emailValue,
+			vars: {
+				sign_up_date: new Date().toISOString(),
+				email: emailValue,
+			},
+			lists: { 'Fandom Account Registration - MASTERLIST': 1 },
+			source: 'homepage_hero_unit',
```
nandy-andy, line 72: That's not true. Let's make the source here `adengine_in_content_ad`

platforms/shared/utils/collapsed-messages/message-box-service.ts
szymon-nawrocki: I think that you mistook "type" with "types" :)
The second one is an array of types that I am using to iterate over message boxes

platforms/shared/utils/collapsed-messages/message-box.ts
```diff
@@ -3,26 +3,36 @@ import { MessageBoxType } from './message-box-service';
...
 	protected redirectUrl: string;
 	protected buttonText: string;
 	protected messageText: string;
 
-	create = (placeholder: HTMLElement, adSlot: AdSlot) => {
+	constructor(adSlot: AdSlot) {
+		this.adSlot = adSlot;
+	}
+
+	create = () => {
```
preludek, line 15: While reviewing this PR I see that you are using arrow functions in classes which is undesirable. For example, arrow functions will be created every time while calling a new MessageBox and will not be shared in a prototype. 

This is already merged, should we make a separate ticket for refactoring and just ignore that here? @nandy-andy @szymon-nawrocki

preludek (changes requested):

platforms/shared/utils/collapsed-messages/message-box.ts
nandy-andy: What holds us from changing it within this PR? 🤔

szymon-nawrocki: Sorry, I didn't realize that using arrow functions inside classes is undesirable. Thanks for pointing this up to me!

@preludek
I think that I can change it in this PR, so I will do that in the next commit

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export type SailthruSource = 'adengine_in_content_ad';
+
+class Sailthru {
+	isLoaded(): boolean {
+		return !!window.Sailthru;
+	}
+
+	userSignup(payload: UserSignupPayload, source: string): void {
+		const { email, onSuccess, onError } = payload;
```
preludek, line 15: You can destructure it directly in line 14.

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,31 @@
+export interface UserSignupPayload {
+	email: string;
+	onSuccess: () => void;
+	onError: () => void;
+}
+
+export type SailthruSource = 'adengine_in_content_ad';
```
preludek, line 7: I wouldn't export that type, and use it in line 14.

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,95 @@
...
+		submitBtn.disabled = true;
+
+		if (!sailthru.isLoaded()) {
+			this.showFormMessage('An error occurred. Please try again later.');
+			submitBtn.disabled = false;
+			return;
+		}
+
+		const payload: UserSignupPayload = this.createPayload(emailValue, submitBtn);
+		const source: SailthruSource = 'adengine_in_content_ad';
```
preludek, line 91: I wouldn't specify that type here. If it's typed correctly in Sailthru class, it will throw an error anyway. Same with the line above. You specified what createPayload returns and what should be passed into `sailthru.userSignup`.

preludek (approved):

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -81,6 +83,11 @@ export class NewsletterFormBox extends MessageBox {
 		const submitBtn: HTMLButtonElement = document.querySelector('.newsletter-submit');
 		submitBtn.disabled = true;
 
+		if (!EMAIL_REGEX.test(emailValue)) {
```
szymon-nawrocki, line 86: I've added additional validation to avoid submitting form with empty email address

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,105 @@
...
+			return;
+		}
+
+		if (!sailthru.isLoaded()) {
+			this.showFormMessage('An error occurred. Please try again later.');
+			submitBtn.disabled = false;
+			return;
+		}
+
+		sailthru.init();
```
szymon-nawrocki, line 98: Turns out Sailthru has to be initialized in AdEngine in order to work correctly, so I've added logic for that.
I've added call for `init()` here, because this service will be used only when Newsletter Form will appear in place of every third collapsed ad slot, so very rarely.
@preludek how does it look to you?

@types/sailthru/index.d.ts
```diff
@@ -0,0 +1,16 @@
+interface SailthruIntegrationData {
+	email: string;
+	vars: {
+		sign_up_date: string;
+		email: string;
+	};
+	lists: { 'Fandom Account Registration - MASTERLIST': number };
```
nandy-andy, line 7: I wonder if we shouldn't go with just `object` or `any` here... but the second isn't really welcome 🙈 It just feels weird to see a string as a property ;p

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,109 @@
...
+	private createPayload(email: string, submitBtn: HTMLButtonElement): UserSignupPayload {
+		const status_clicked = `cm_${this.type.toLowerCase()}_clicked`;
+
+		const onSuccess = (): void => {
+			this.showFormMessage('Thanks for signing up!');
+			submitBtn.disabled = false;
+			this.sendTrackingEvent(status_clicked);
+		};
+
+		const onError = (): void => {
```
nandy-andy, line 63: I think we should be consistent and resign from this arrow functions (`onSuccess`, `onError`) as well as we did after the previous conversation: https://github.com/Wikia/ad-engine/pull/1215#discussion_r780108758

platforms/shared/utils/collapsed-messages/message-box-service.ts
```diff
@@ -1,40 +1,48 @@
 import { AdSlot } from '@wikia/ad-engine';
 import { MessageBoxCreator } from './message-box-creator';
 
-export type MessageBoxType = 'REGISTER' | 'FANLAB' | 'NEWSLETTER';
+export type MessageBoxType = 'REGISTER' | 'FANLAB' | 'NEWSLETTER_LINK' | 'NEWSLETTER_FORM';
 
 export class MessageBoxService {
-	private types: MessageBoxType[] = ['REGISTER', 'FANLAB', 'NEWSLETTER'];
+	private messageBoxCreator: MessageBoxCreator;
+	private types: MessageBoxType[] = ['REGISTER', 'FANLAB', 'NEWSLETTER_FORM', 'NEWSLETTER_LINK'];
```
nandy-andy, line 8: Should we have `NEWSLETTER_FORM` in types when Sailthru is not enabled? 🤔

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,56 @@
+import { context, utils } from '@ad-engine/core';
+
+export interface UserSignupPayload {
+	email: string;
+	onSuccess: () => void;
+	onError: () => void;
+}
+
+type SailthruSource = 'adengine_in_content_ad';
```
nandy-andy, line 9: I think this type is unnecessary. Let's just use `string` in the defintion of `userSignup()`.

platforms/shared/utils/collapsed-messages/message-box-service.ts
nandy-andy: Yeah, I think we shouldn't allow a collapse message with the form to appear on pages where Sailthru isn't enabled nor loaded.

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const logGroup = 'sailthru';
+const packageUrl = 'https://ak.sail-horizon.com/spm/spm.v1.min.js';
```
nandy-andy, line 12: It's not necessary since we still treat it as an experiment but a nice-to-have is to move `packageUrl` and `ACCT_ID` to our `instant-config` service, so in the future we could change them without releasing code changes.

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
szymon-nawrocki: I've used arrow functions here, because with "normal" functions there is a problem with `this`

nandy-andy: We can figure something out. Use biding to bind to the right context or pass whatever we need as an argument. I'll think about it once I get to my laptop :-)

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,96 @@
...
+		super(adSlot);
+		this.type = 'NEWSLETTER_FORM';
+		this.messageText = 'THE LATEST TRENDS, DELIVERED STRAIGHT TO YOUR INBOX.';
+		this.buttonText = 'Sign up';
+		this.status_impression = 'cm_newsletter_form_impression';
+		this.status_clicked = 'cm_newsletter_form_clicked';
+	}
+
+	create(): void {
+		sailthru.init();
```
nandy-andy, line 16: What happens if Sailthru has been already initialised by silver-surfer and we re-initialise it here? 🤔

platforms/shared/utils/collapsed-messages/message-box-service.ts
```diff
@@ -1,40 +1,55 @@
...
 	private currentType = 0;
 
-	getCurrentTypeIndex = (): number => {
+	constructor() {
+		this.messageBoxCreator = new MessageBoxCreator();
+		this.init();
+	}
+
+	init(): void {
+		this.types = sailthru.isEnabled()
```
nandy-andy, line 17: Shouldn't we also check if Sailthru loaded?

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,96 @@
...
+
+		const placeholder = this.adSlot.getPlaceholder();
+		if (!placeholder) {
+			throw new Error(`No placeholder to insert '${this.type}' Message Box`);
+		}
```
nandy-andy, line 21: It seems this logic is the same in [the base class](https://github.com/Wikia/ad-engine/pull/1215/files#diff-f79ceb05f281f21ad82180a59c54153209f87af26b2a3189e6bf007e215f2c08R29-R31), so we could extract it to a helper method that we'd just reuse here.

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,96 @@
...
+
+		const wrapper = this.createBoxWrapper();
```
nandy-andy, line 23: If I'm correct this method differs from the base class `create()` method with two things: (1) Sailthru initialisation and (2) adding different things to the wrapper and then to the placeholder. Consider simplifying the base class method to a simple one that receives things to add to the placeholder, adds them and sends the tracking pixel. Then in the `NewsletterFormBox`'s version you'll just additionally init Sailthru and add different elements 🤔

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
nandy-andy: It's also a promise that's returned from `init()`, so I think we should do everything that's below in the `then()` callback 🤔

platforms/shared/utils/collapsed-messages/message-box-service.ts
nandy-andy: I think we don't need to since if it's enabled we'll call `NewsletterFormBox::create()` that will call `Sailthru::init()` that checks if it's loaded and if not it's loading the script.

platforms/shared/utils/collapsed-messages/message-box-service.ts
```diff
@@ -1,40 +1,55 @@
...
+	private types: MessageBoxType[];
 	private currentType = 0;
 
-	getCurrentTypeIndex = (): number => {
+	constructor() {
+		this.messageBoxCreator = new MessageBoxCreator();
+		this.init();
+	}
+
+	init(): void {
```
nandy-andy, line 16: Do we need this method? I can't find its usage anywhere else and I think setting `this.types` fits perfectly in the `constructor()` 🤔

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,96 @@
...
+		placeholder.append(wrapper);
+
+		this.sendTrackingEvent(this.status_impression);
+	}
+
+	private createForm(): HTMLElement {
+		const form = document.createElement('form');
+		form.className = 'newsletter-form';
+
+		form.innerHTML = `
```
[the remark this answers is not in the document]
szymon-nawrocki: I created it this way because it was easier for me, but I did not know that it can cause potential security issues

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
szymon-nawrocki: Sailthru will be initialized from scratch, only if it is not already loaded:
https://github.com/Wikia/ad-engine/blob/93541d646bf79f6f204b5667bd254aed1da7fdd4/src/ad-services/sailthru/index.ts#L29

szymon-nawrocki: I've made a change to the NewsletterForm, and I removed completely `create()` method, instead of this method, I created `getElementsToAppend()` where we can specify what elements should be added to _Message Box wrapper_.

I was thinking where will be the best place to put `sailthru.init()` method, so I put it to the constructor, @nandy-andy do you think it is a good place for it?

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,99 @@
...
+		this.status_impression = 'cm_newsletter_form_impression';
+		this.status_clicked = 'cm_newsletter_form_clicked';
+		sailthru.init();
+	}
+
+	getElementsToAppend(): HTMLElement[] {
+		const elements: HTMLElement[] = [];
+		elements.push(this.createForm(), this.createFormMessage());
+
+		return elements;
```
nandy-andy, line 20: I think in this case, where there is no additional logic you can simplify the method to:
```js
getElementsToAppend(): HTMLElement[] {
  return [this.createForm(), this.createFormMessage()]
}
```

platforms/shared/utils/collapsed-messages/newsletter-form-box.ts
```diff
@@ -0,0 +1,99 @@
...
+		const submitBtn: HTMLButtonElement = document.querySelector('.newsletter-submit');
+		submitBtn.disabled = false;
+	}
+
+	private doEmailSignUp(event): void {
+		event.preventDefault();
+
+		const emailInput: HTMLInputElement = document.querySelector('.newsletter-email');
+		const emailValue = emailInput.value;
+		const submitBtn: HTMLButtonElement = document.querySelector('.newsletter-submit');
```
nandy-andy, line 85: You can create a small method `disableSubmitButton()`, move there this line and the next line and reuse the new method by replacing lines 76, 77 above :)

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		if (!this.isLoaded()) {
+			utils.logger(logGroup, 'loading');
+
+			return utils.scriptLoader
+				.loadScript(packageUrl)
+				.then(() => {
+					window.Sailthru.init({ customerId: ACCT_ID });
+					utils.logger(logGroup, 'ready');
```
szymon-nawrocki, line 36: Here return `Promise.resolve()`

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,59 @@
...
+				})
+				.catch(() => {
+					throw new Error('Error occurred while loading Sailthru.');
```
szymon-nawrocki, line 39: And here `Promise.reject()` instead of error?

@nandy-andy  what do you think?

src/ad-services/sailthru/index.ts
```diff
@@ -0,0 +1,59 @@
...
+				});
+		}
```
nandy-andy, line 41: @szymon-nawrocki and here one more `Promise.resolved()` if Sailthru is enabled and loaded 🤔

nandy-andy (approved):

dmnsobczak (approved):

PR #1217, opened 2021-12-30
ADEN-11477 Replacing clip() with dynamic resize

dmnsobczak:


platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -75,6 +75,12 @@ export class UapDomManager {
 		this.manipulator.element(this.adSlot.getElement()).setProperty('height', height);
 	}
 
+	setSlotHeightClipping(): void {
+		this.manipulator
+			.element(this.adSlot.getElement())
+			.setProperty('clipPath', `inset(0 0 ${this.reader.getSlotHeightClipping()}px)`);
```
preludek, line 81: Did you check inset on `caniuse`, cause it looks worse than clip 🙈 

https://caniuse.com/mdn-css_properties_inset

platforms/ucp-mobile/templates/bfaa-3-template.ts
```diff
@@ -1,111 +0,0 @@
...
-			UapDomManager,
-			UapDomReader,
-			VideoDomReader,
-			VideoDomManager,
-			CloseButtonHelper,
-			StickinessTimeout.provide(universalAdPackage.BFAA_UNSTICK_DELAY),
-			registerUcpMobileUapDomElements(),
-		],
-	);
-}
```
preludek, line 111: That had to be so satisfying :D

platforms/shared/templates/helpers/uap-dom-reader.ts
```diff
@@ -98,30 +94,21 @@ export class UapDomReader {
...
-		return this.calculateProgress(offset);
+	private calculateSlotHeight(ratio: number): number {
+		return (1 / ratio) * this.adSlot.element.offsetWidth;
 	}
 
-	private calculateProgress(offset: number): number {
-		if (offset >= 1) {
-			return 1;
-		}
+	getSlotHeightClipping(): number {
```
preludek, line 101: Nit-pick: This is not following the name convention and the description in line 71. What do you think of changing `Progress` in line 71 to `Clipping`?

platforms/shared/templates/handlers/bfaa-new/bfaa-new-bootstrap-handler.ts
```diff
@@ -1,61 +0,0 @@
...
-
-	async onLeave(): Promise<void> {
-		this.adSlot.show();
-		document.body.classList.add('has-new-uap');
-	}
-
-	async onDestroy(): Promise<void> {
-		document.body.classList.remove('has-new-uap');
-	}
-}
```
preludek, line 61: 💎

preludek (changes requested):

platforms/ucp-mobile/templates/bfaa-3-template.ts
dmnsobczak: Yeees, it was <3

platforms/shared/templates/handlers/bfaa-new/bfaa-new-bootstrap-handler.ts
dmnsobczak: 🙇

platforms/shared/templates/helpers/uap-dom-manager.ts
dmnsobczak: Based on our Supported Browsers page, we're good to go. Fandom requires updated browsers and IE is no longer supported:
https://community.fandom.com/wiki/Help:Supported_browsers
Additionally: `inset` CSS property is widely used in Unified Platforms CSS files, mostly for `box-shadow` property. That's why I decided that we don't need to worry about that :D

platforms/shared/templates/helpers/uap-dom-reader.ts
dmnsobczak: Git is making a mess in the history: I removed `calculateProgress()` completely, because this method was unused and I added completely different `getSlotHeightClipping()` which follows `getSlotHeightImpact()` or `getSlotHeightResolved()` convention 🤔

preludek (approved):

PR #1218, opened 2021-12-30
ADEN-11483 Fix existing TypeScript warnings

szymon-nawrocki:
### Ticket:
https://fandom.atlassian.net/browse/ADEN-11483


src/ad-services/nielsen/index.ts
```diff
@@ -8,7 +8,7 @@ const nlsnConfig: Dictionary = {};
 /**
  * Creates Nielsen Static Queue Snippet
  */
-function createInstance(nielsenKey) {
+function createInstance(nielsenKey): void {
```
szymon-nawrocki, line 11: This function returns a value so it will be better to use 'any' instead of 'void'

spec/ad-engine/listeners/porvata-listener.spec.ts
```diff
@@ -3,10 +3,10 @@ import { PorvataListener } from '@wikia/ad-products/video/porvata/porvata-listen
 import { expect } from 'chai';
 import { createSandbox, spy } from 'sinon';
 
-function getListener() {
+function getListener(): any {
```
Tomasz-at-Fandom, line 6: nit: we could be more specific here then any

spec/communication/is-type.spec.ts
```diff
@@ -18,9 +18,9 @@ describe('isType', () => {
 	});
 
 	it('should pass local for local', () => {
-		actions.forEach((action) => {
-			if (isType(action, localExample)) {
-				results.push(action);
+		actions.forEach((act) => {
```
Tomasz-at-Fandom, line 21: Did the word "action" cause some conflict?

Full words are better for readability, let's not make funny shortcuts if not necessary.

src/ad-engine/utils/flow-control.ts
```diff
@@ -29,14 +29,14 @@ export function once(
 	const isObject: boolean = typeof emitter === 'object';
 	const hasAddEventListener: boolean =
 		isObject && typeof (emitter as HTMLElement).addEventListener === 'function';
+	let testOptions = { ...options };
 
 	return new Promise((resolve, reject) => {
 		if (typeof options === 'boolean') {
-			options = { capture: options };
+			testOptions = { capture: options };
 		}
```
Tomasz-at-Fandom, line 37: We don't have to overwrite `testOptions` it can be a const. Let's make this simpler e.g.

```TypeScript
	const testOptions = typeof options === 'boolean' ? { capture: options } : { ...options };

	return new Promise((resolve, reject) => {
		if (hasAddEventListener) {
			(emitter as HTMLElement).addEventListener(eventName, resolve, { ...testOptions, once: true });
		} else {
			reject(new Error('Emitter does not have `addEventListener` nor `once` method.'));
		}
	});
```

spec/communication/is-type.spec.ts
szymon-nawrocki: Not a conflict per se, but it is shadowing "action" variable which is imported at the beginning of the file, so it is also not the best for readibility

Tomasz-at-Fandom: OK that make sense 👍

src/ad-products/templates/interface/video/ui-template.ts
```diff
@@ -75,7 +69,7 @@ const getTemplates = (params, videoSettings?: PorvataSettings | UapVideoSettings
 export function selectTemplate(
 	videoSettings: PorvataSettings | UapVideoSettings,
 	params?: Partial<UapParams>,
-) {
+): any {
```
Tomasz-at-Fandom, line 72: Currently this will always return a list [], so let's declare that.

spec/ad-engine/listeners/porvata-listener.spec.ts
nandy-andy: We could and yet it does not match `PovataListener` interface :/ I think we'd need a custom type just for the unit test 🤔

Tomasz-at-Fandom: Sounds like more problems than it's worth, let's leave it as it is.

dmnsobczak (approved):

PR #1221, opened 2022-01-11
ADEN-11452 Restore FMR rotation for AdBlock

dmnsobczak:


src/ad-products/utils/fmr-rotator.ts
```diff
@@ -77,7 +77,13 @@ export class FmrRotator {
 
 	private initializeBTRotation(): void {
 		this.pushNextSlot();
-		this.swapRecirculation(false);
+
+		let recirculationVisible = false;
```
nandy-andy, line 81: Shouldn't it be a `document.querySelector`, so the `true`/`false` is based on what's on the page now: if there is a recirculation and if it's visible? 🤔

nandy-andy: Discussed "offline" ;)

nandy-andy (approved):

src/ad-products/utils/fmr-rotator.ts
```diff
@@ -77,7 +77,13 @@ export class FmrRotator {
...
+
+		setInterval(() => {
+			this.swapRecirculation(recirculationVisible);
+			recirculationVisible = !recirculationVisible;
```
szymon-nawrocki, line 85: I think this line of code is redundant, because you're changing `recirculationVisible` variable and not using it later anywhere

dmnsobczak: `setInterval()` is executing callback every 12 seconds so it will reuse this variable - it a semaphore :)

szymon-nawrocki: Never mind, I mistook `setInterva()` l for `setTimeout()` xd

szymon-nawrocki (approved):

PR #1223, opened 2022-01-12
ADEN-11307 Replace tslint with eslint

nandy-andy:


@types/pbjs/index.d.ts
```diff
@@ -22,26 +22,52 @@ interface PrebidAdUnit {
 	labelAll?: string[];
 }
 
+interface PrebidNativeMediaType {
+	image: {
+		required: boolean,
+		sizes: Array<number>
```
preludek, line 28: It seems that in AdEngine we use different convention when typing arrays, so I'd change that to `number[]`

@types/pbjs/index.d.ts
```diff
@@ -22,26 +22,52 @@ interface PrebidAdUnit {
...
+	};
+	clickUrl: {
+		required: boolean,
+	};
+	body: {
+		required: boolean,
+	};
+	icon: {
+		required: boolean,
+		sizes: Array<number>
```
preludek, line 45: ditto

@types/pbjs/index.d.ts
```diff
@@ -114,7 +144,7 @@ interface Pbjs {
 
 	aliasBidder(bidderCode: string, alias: string): void;
 
-	registerBidAdapter(bidderAdaptor: () => {}, bidderCode: string): void;
+	registerBidAdapter(bidderAdaptor: unknown, bidderCode: string): void;
```
preludek, line 147: `() => void` ? 🤔

src/ad-engine/services/local-cache.ts
```diff
@@ -14,29 +8,29 @@ class LocalCache {
 	}
 
 	// @TODO: Should not return boolean if item expired
-	get<T = any>(key: string): T {
-		const cacheItem: CacheItem = this.storage.getItem<CacheItem>(key);
+	get(key: string): boolean | unknown {
+		const cacheItem: Record<string, unknown> = this.storage.getItem<Record<string, unknown>>(key);
```
preludek, line 12: We have an interface Dictionary in `src/ad-engine/models/dictionary` that can be used instead of Record<string, T>. I'd go with Dictionary everywhere it's possible for consistency.

bckatarzyna (approved):

@types/pbjs/index.d.ts
bckatarzyna: Or https://github.com/Wikia/ad-engine/pull/1223#discussion_r786506166 ? (I've just added)

nandy-andy: Adaptor :D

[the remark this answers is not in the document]
nandy-andy: It works for linter, it doesn't work for webpack/babel ;p There is something weird going on anytime we try to import something in `ad-engine/@types`. I'll need to think about it how to do it properly, so we don't import anything we shouldn't in there since `PrebidAdapter` is defined in `src/ad-bidders/prebid/prebid-models`. Maybe it should be also moved to `ad-engine/@types`  🤔 Since it's failing the built I went with TJ's idea and I'll think about it in the follow-up ticket 👌

@types/google/index.d.ts
```diff
@@ -0,0 +1 @@
+import '../../node_modules/@alugha/ima/typings/ima';
```
dmnsobczak, line 1: 💎

src/ad-engine/utils/targeting.ts
```diff
@@ -44,7 +44,7 @@ class Targeting {
 		return [];
 	}
 
-	private matchesTargetingBundle(bundle: Dictionary<string[]>): boolean {
+	private matchesTargetingBundle(bundle: Dictionary<string | string[]>): boolean {
```
dmnsobczak, line 47: Why `string`? Bundle element should always be an array:
https://services.fandom.com/icbm/dashboard/feature_flag/icTargetingBundles

src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -85,7 +85,7 @@ class IasPublisherOptimization {
 			iasPETSlots.push(config);
 		});
 
-		// @ts-ignore
+		// @ts-ignore the __iasPET API is not under our control
```
dmnsobczak, line 88: Can't we handle it like `NOLBUNDLE?: any;`?

src/communication/event-types.ts
```diff
@@ -1,5 +1,4 @@
 import { AdSlot, Dictionary } from '@ad-engine/core';
-// tslint:disable-next-line:import-blacklist
```
dmnsobczak, line 2: Is that mean that rule restricting cross-module import is unavailable now? It was checking if you don't for example import `ad-bidders` inside `ad-services` module - you can only import from `ad-engine` 🤔

dmnsobczak (approved):
Everything is fine, few questions left.
@nandy-andy Please move to QA after answering them :)

src/communication/event-types.ts
nandy-andy: This line was disabling this kind of check but generally yeah, when switching from `tslint` to `eslint` the custom `import-blacklist` rule was removed and I'm trying to recreate it or some similar solution in a follow-up ticket: [ADEN-11503](https://fandom.atlassian.net/browse/ADEN-11503)

src/ad-engine/utils/targeting.ts
nandy-andy: Good catch. I connected it with GAM key-vals (that can be either string or array) not our instant-config var 🙈

@types/google/index.d.ts
nandy-andy: Kudos to @preludek :)

src/ad-services/ias-publisher-optimization/index.ts
nandy-andy: Right! We do it already, this ts-ignore is not needed here, thanks 🙇 https://github.com/Wikia/ad-engine/pull/1223/files#diff-ad751592e52c8e1020bd2cc0cee8fbd493a0796ee39d50587eeb8626234d8cbaL10

nandy-andy:
Thanks @dmnsobczak I've addressed your comments in https://github.com/Wikia/ad-engine/pull/1230

preludek (approved):

PR #1229, opened 2022-01-19
SITE-2038 | Update adMarketplace selector for UCP mobile

latata:


platforms/ucp-mobile/ad-context.ts
```diff
@@ -73,8 +73,8 @@ export const basicContext = {
 	services: {
 		adMarketplace: {
 			enabled: false,
-			insertMethod: 'prepend',
-			insertSelector: '.wds-global-navigation__search-suggestions.wds-dropdown__content',
+			insertMethod: 'after',
```
nandy-andy, line 76: Is the change only needed for FandomMobile? The "AdMarketplace" ads are enabled on both skins: FandomMobile and FandomDesktop :)

nandy-andy: Nvm, [its inserting method](https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp-desktop/ad-context.ts#L75) is set to "after" there 👌

nandy-andy (approved):

preludek (approved):

preludek (approved):

src/ad-services/ad-marketplace/index.ts
```diff
@@ -127,22 +139,34 @@ class AdMarketplace {
 		);
 	}
 
-	private getInstantSearchAdsWrapper(): HTMLElement | null {
+	private getElement(selector: string) {
+		const element = document.querySelector(selector);
+
+		if (!element) {
+			return null;
+		}
```
latata, line 147: It seems to be not needed as querySelector returns `null`, if the element has not been found.

PR #1231, opened 2022-01-19
ADEN-11492 | Identify sequential UAP ad

Tomasz-at-Fandom:


platforms/shared/sequential-messaging/data-structures/sequential-messaging-cookie.ts
```diff
@@ -0,0 +1,3 @@
+export interface SequentialMessagingCookie {
+	[lineItemId: number]: { length: number };
+}
```
Tomasz-at-Fandom, line 3: I know this looks the same as `IcSequentialMessaging` but it will change in the next step so no point in merging them now.

platforms/shared/sequential-messaging/domain/sequence-handler.ts
```diff
@@ -0,0 +1,49 @@
+import { InstantConfigServiceInterface } from '@wikia/ad-engine';
+import { IcSequentialMessaging } from './data-structures/ic-sequential-messaging';
+import { SequentialMessagingCookie } from './data-structures/sequential-messaging-cookie';
+import { DetectorFactory } from './factories/detector-factory';
+import { CookieJarInterface } from './interfaces/cookie-jar-interface';
+
+export class SequenceHandler {
```
mfarul, line 7: Not sure what we handle here? is this supposed to store the sequence number per user ? 
If so the handleMethod should just detect if slot is filled with sequence Ad and call an internal method  storeSequenceNumberForUser. As now this is our product: 
1) detect if we are getting sequenced Ad (line item config)
2) store this info per user (in cookie)
3) use this info for targeting Ad (this will be implemented further)

If this is the exact flow we should see it in method names and all being called from main handle method.

mfarul (changes requested):

platforms/shared/sequential-messaging/domain/data-structures/sequential-messaging-cookie.ts
```diff
@@ -0,0 +1,3 @@
+export interface SequentialMessagingCookie {
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: Well this goes in to the cookie and IcSequentialMessaging come from ICBM (or maybe GAM in the future). I expect trouble if I'll represent two different "external" data stores with one data structure. E.g. in the next step I expect we will probably store additionally information about the ad slot that this came in, that we will not be defining in the configuration.

platforms/shared/sequential-messaging/domain/data-structures/ic-sequential-messaging.ts
```diff
@@ -0,0 +1,3 @@
+export interface IcSequentialMessaging {
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: Well the jury is still out on this one, I expect that this will change as we progress through the development of this feature, so for now I just labeled it with the icbm var name that this data comes from.

platforms/ucp-desktop/ucp-desktop-platform.ts
```diff
@@ -39,6 +40,7 @@ export class UcpDesktopPlatform {
 		this.pipeline.add(
 			() => context.extend(basicContext),
 			parallel(InstantConfigSetup, () => bootstrapAndGetConsent()),
+			SequentialMessagingSetup,
```
preludek, line 43: Is this the right place for it? I feel like those SequentialMessaging is more like an ad-product 🤔

platforms/shared/sequential-messaging/domain/sequence-handler.ts
```diff
@@ -0,0 +1,49 @@
...
+
+		const sequenceDetector = new DetectorFactory(icSequentialMessaging).makeSequenceDetector();
+
+		if (sequenceDetector.isAdSequential(lineItemId.toString())) {
+			this.storeCookie(lineItemId, icSequentialMessaging);
+		}
+	}
+
+	private validateInstantConfigInput(icSequentialMessaging: IcSequentialMessaging): boolean {
+		if (typeof icSequentialMessaging !== 'object') {
```
preludek, line 32: I have some doubts about that line. If we plug in `null` this will go through and break on line 34.

platforms/shared/sequential-messaging/domain/factories/detector-factory.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@wikia/dependency-injection';
+import { IcSequentialMessaging } from '../data-structures/ic-sequential-messaging';
+import { SequenceDetector } from '../sequence-detector';
+
+@Injectable()
+export class DetectorFactory {
+	constructor(private icSequentialMessaging: IcSequentialMessaging) {}
```
preludek: Because we are naming icVars this way the naming starting with ic feels really odd.

platforms/shared/sequential-messaging/domain/factories/detector-factory.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@wikia/dependency-injection';
+import { IcSequentialMessaging } from '../data-structures/ic-sequential-messaging';
+import { SequenceDetector } from '../sequence-detector';
+
+@Injectable()
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: Oh good catch, this is a leftover from previous implementation where I was injecting ICBM here. That's why I created a separate Factory class in the fist place but yea this could go back to the Sequence Detector all together.

platforms/shared/sequential-messaging/domain/interfaces/cookie-jar-interface.ts
```diff
@@ -0,0 +1,5 @@
+import { SequentialMessagingCookie } from '../data-structures/sequential-messaging-cookie';
+
+export interface CookieJarInterface {
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: Well I guess it's possible to use something else then cookies so I'll change the name of the interface and will make an adapter to hide the cookie `name` required by `js-cookie` package. I should also rename the `SequentialMessagingCookie` to something like `UserSequentialMessageState`

platforms/shared/sequential-messaging/domain/sequence-handler.ts
```diff
@@ -0,0 +1,49 @@
+import { InstantConfigServiceInterface } from '@wikia/ad-engine';
+import { IcSequentialMessaging } from './data-structures/ic-sequential-messaging';
+import { SequentialMessagingCookie } from './data-structures/sequential-messaging-cookie';
+import { DetectorFactory } from './factories/detector-factory';
+import { CookieJarInterface } from './interfaces/cookie-jar-interface';
+
+export class SequenceHandler {
+	constructor(
+		private instantConfig: InstantConfigServiceInterface,
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: That can get tricky down the line if every feature would establish its own interface that the ICBM class would have to abide by to meet its requirements. It would be easier to discuss this on a call.

platforms/ucp-desktop/ucp-desktop-platform.ts
Tomasz-at-Fandom: It's the best place I could find for a one point per platform plug in.

platforms/shared/sequential-messaging/domain/factories/detector-factory.ts
Tomasz-at-Fandom: This was refactored now, please see the new implementation.

platforms/shared/sequential-messaging/domain/data-structures/user-sequential-message-state.ts
```diff
@@ -0,0 +1,3 @@
+export interface UserSequentialMessageState {
+	[sequentialAdIs: number]: { length: number };
```
preludek, line 2: typo :D

preludek (approved):

platforms/shared/sequential-messaging/domain/data-structures/user-sequential-message-state.ts
Tomasz-at-Fandom: Good catch, like sniper 🧐

PR #1241, opened 2022-01-25
ADEN-11496 | Sequential Messaging

Tomasz-at-Fandom:


platforms/shared/sequential-messaging/domain/sequence-continuation-handler.ts
```diff
@@ -0,0 +1,48 @@
+import { UserSequentialMessageStateStoreInterface } from './interfaces/user-sequential-message-state-store.interface';
+import { UserSequentialMessageState } from './data-structures/user-sequential-message-state';
+import { SequentialMessagingConfigStoreInterface } from './interfaces/sequential-messaging-config-store.interface';
+import { TargetingManagerInterface } from './interfaces/targeting-manager-interface';
+
+type adConfig = { length: number | string; targeting: Record<string, unknown> };
```
preludek, line 6: nit pick - I think all type names should start with a capital letter

platforms/shared/sequential-messaging/infrastructure/sequential-messaging-config-store.ts
```diff
@@ -31,7 +31,9 @@ export class SequentialMessagingConfigStore implements SequentialMessagingConfig
 		for (const val of Object.values(sequentialMessagingConfig)) {
```
preludek, line 31: `validateSequentialMessagingConfigInput` method will break if we call it with null, out of curiosity is this a possible scenario?

preludek (approved):

platforms/shared/sequential-messaging/infrastructure/sequential-messaging-config-store.ts
Tomasz-at-Fandom: Good catch! 👏  If someone would remove or clear the var in ICBM then this function would probably receive null or undefined. I've patched it and supplemented the test with a null sample.

PR #1246, opened 2022-01-31
ADEN-11426 Add ExCo player to our outstream players integrations

nandy-andy:


platforms/shared/setup/base-context.setup.ts
```diff
@@ -178,6 +179,19 @@ export class BaseContextSetup implements DiProcess {
 		);
 	}
 
+	private setupOutstreamPlayers(): void {
+		const distroscalePlayerEnabled = this.instantConfig.get('icDistroScale');
+		const exCoPlayerEnabled = this.instantConfig.get('icExCoPlayer');
+
+		if (exCoPlayerEnabled === true) {
```
nandy-andy, line 186: The conditions here would grow with 2 more players to test. I wonder if we shouldn't introduce here some strategy or something that will simplify the code. On the other hand, it's just an experiment to validate which player performs the best and we'll end up with the one chosen and Porvata 🤔

platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
```diff
@@ -108,7 +109,12 @@ export class UcpDesktopAdsMode implements DiProcess {
 				const incontentPlayer = slotService.get('incontent_player');
 				if (!isLoaded && incontentPlayer) {
 					slotDataParamsUpdater.updateOnCreate(incontentPlayer);
-					distroScale.call();
+
+					if (distroScale.isEnabled()) {
```
nandy-andy, line 113: This conditions could also be replaced with something more sophisticated and easy to read 🤔 Like an abstraction of `EnabledOutstreamPlayer` that goes through players defined in the code: `[distroScale, exco]`, returns just the one enabled and calls `call()` on it. `Porvata` player acts differently here, though 🤔

platforms/shared/setup/base-context.setup.ts
dmnsobczak: Yeah, I think we can live with some if's for a brief amount of time. We should focus to not mess the logic up :D

platforms/shared/setup/base-context.setup.ts
```diff
@@ -178,6 +179,19 @@ export class BaseContextSetup implements DiProcess {
...
+			context.set('services.exco.enabled', exCoPlayerEnabled);
+			context.set('services.distroScale.enabled', false);
+		} else {
+			context.set('services.exco.enabled', false);
+			context.set('services.distroScale.enabled', distroscalePlayerEnabled);
+		}
+	}
```
dmnsobczak, line 193:
```suggestion
	private setupOutstreamPlayers(): void {
		if (this.instantConfig.get('icExCoPlayer')) {
			context.set('services.exco.enabled', true);
			context.set('services.distroScale.enabled', false);
			return;
		}
		
		context.set('services.distroScale.enabled', this.instantConfig.get('icDistroScale'));
	}
```

src/ad-services/exco/index.ts
```diff
@@ -0,0 +1,53 @@
+import { context, utils } from '@ad-engine/core';
+
+const logGroup = 'exco';
+
+class ExCo {
+	private readonly id: string;
+
+	constructor() {
+		this.id = 'f6c04939-d96e-4bc6-850e-d0e6e6cf9701';
```
dmnsobczak, line 9: In the future this value might be different, depending the platform, and it should go to the context. But it's totally fine for now :)

src/ad-services/exco/index.ts
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	getId(): string {
```
dmnsobczak, line 12: You can create a getter while creating a variable

src/ad-services/exco/index.ts
```diff
@@ -0,0 +1,53 @@
...
+		const exCoPlayerDiv = document.createElement('div');
+		exCoPlayerDiv.id = this.getId();
+
+		const incontentPlayerContainer = document.getElementById('incontent_player');
+		incontentPlayerContainer.appendChild(exCoPlayerDiv);
+		incontentPlayerContainer.classList.remove('hide');
+	}
+}
+
+export const exco = new ExCo();
```
dmnsobczak, line 53: I think it should be `exCo` if we want to strictly follow our camelcase service names rules 🤔

dmnsobczak (approved):
Everything fine, few comments, no bugs though. Address whatever you like and move the ticket to QA :)

src/ad-services/exco/index.ts
nandy-andy: Yeah, I kept it here since it's an experiment and we don't know if we'd stay with ExCo 🤷‍♂️

src/ad-services/exco/index.ts
nandy-andy: I think I did what you had thought about. But I'm not sure if I like it. It's less verbose, isn't it? ;) Or maybe I misunderstood you 🙈

platforms/shared/setup/base-context.setup.ts
```diff
@@ -178,6 +179,16 @@ export class BaseContextSetup implements DiProcess {
 		);
 	}
 
+	private setupOutstreamPlayers(): void {
```
preludek, line 182: That's the place I have concerns with. We will be adding 2 more partners to test soon. Maybe we can refactor it? Since we know for sure we are running an A/B/C/D test for 4 partners maybe we can have it behind a variable already like icVideoTest or something? And just the test pick a partner to load. If the we are not in the test is just default  to DistroScale since we are running with them ATM.

preludek (approved):

PR #1249, opened 2022-02-01
ADEN-11546 Add tracking to the replay button in Porvata player

nandy-andy:


src/ad-products/templates/interface/video/replay-overlay.ts
```diff
@@ -73,6 +74,11 @@ export class ReplayOverlay {
 		) {
 			const replayIcon = addReplayIcon(overlay);
 
+			communicationService.emit(eventsRepository.AD_ENGINE_VIDEO_REPLAY_DISPLAYED, {
```
nandy-andy, line 77: Another way of doing it but maybe a bit challenging is to send the tracking data within the listener we setup in: https://github.com/Wikia/ad-engine/blob/f3466c06901613a7eec2d57cb6c8bd1556246171/src/ad-products/templates/interface/video/replay-overlay.ts#L59 🤔

src/ad-products/templates/interface/video/replay-overlay.ts
```diff
@@ -92,6 +93,11 @@ export class ReplayOverlay {
 			container.parentElement.insertBefore(overlay, container);
 		}
 
+		communicationService.emit(eventsRepository.AD_ENGINE_VIDEO_REPLAY_OVERLAY_DISPLAYED, {
```
nandy-andy, line 96: Another way of doing it but maybe a bit challenging is to send the tracking data within the listener we setup in: https://github.com/Wikia/ad-engine/blob/f3466c06901613a7eec2d57cb6c8bd1556246171/src/ad-products/templates/interface/video/replay-overlay.ts#L59 🤔

szwajkowski (approved):
Well, it looks like a copy of Learn more tracking, so it is gonna work :)

jakubbanas (approved):

platforms/shared/templates/handlers/video/video-bootstrap-handler.ts
```diff
@@ -59,6 +59,7 @@ export class VideoBootstrapHandler implements TemplateStateHandler {
 
 			fromEvent(player, 'wikiaAdStarted').pipe(
 				mergeMap(() => fromEvent(player, 'wikiaAdCompleted')),
+				debounceTime(10),
```
szwajkowski, line 62: nit pick - 10 is a magic number - would be nice to make it a constant with indication whether its seconds or microseconds

szwajkowski (approved):
nice hack :D

PR #1253, opened 2022-02-04
ADEN-11565 UAP improvements

nandy-andy:


src/ad-products/templates/interface/video/replay-overlay.ts
```diff
@@ -57,7 +57,16 @@ export class ReplayOverlay {
 		const overlay = document.createElement('div');
 
 		overlay.classList.add(replayOverlayClass);
-		overlay.addEventListener('click', () => video.play());
+		overlay.addEventListener('click', () => {
+			video.play();
+
+			if (video.getPlayCounter() > 1) {
```
szwajkowski, line 63: tbh it is a bit confusing, since it is a "REPLAY" overlay it means that all of the clicks should mean that it is a replay

szwajkowski: is it possible to add an unittest for this if?

nandy-andy: Yeah, unfortunately, we use replay overlay as a player overlay for yet another version of Fan Takeover - click-to-play :/ Let me check how hard it is to unit test it 👌 I can also try and rename it to "player overlay" 🤔

szwajkowski (approved):
lgtm, thx for all the additional changes :)

PR #1254, opened 2022-02-04
ADEN-11425 add Anyclip partner

preludek:


platforms/ucp-desktop/ad-context.ts
```diff
@@ -75,6 +75,10 @@ export const basicContext = {
 			insertMethod: 'after',
 			insertSelector: '.search-modal__content > form',
 		},
+		anyclip: {
+			pubname: 'mavenio',
```
nandy-andy, line 79: Have you considered adding `enabled: false` as default. It'll make it consistent with DistroScale and ExCo configs and it'll keep AnyClip disabled on other platforms for sure (as we change it only in the `setupOutstreamPlayers()` for ucp-desktop :)

src/ad-services/anyclip/index.ts
```diff
@@ -0,0 +1,43 @@
+import { context, utils } from '@ad-engine/core';
+import { isEmpty } from 'lodash';
```
nandy-andy, line 2: Personally I'd resign from coupling it with 3rd-party dependency because it's only two config parameters we want to verify. But it's like... my opinion, man :D And we're still in the experiment phase with the players. So, think about it and it's up to you to keep or get rid of lodash ;)

src/ad-services/anyclip/index.ts
```diff
@@ -0,0 +1,43 @@
...
+		const libraryUrl = 'https://player.anyclip.com/anyclip-widget/lre-widget/prod/v1/src/lre.js';
+		const incontentPlayerContainer = document.getElementById('incontent_player');
+		incontentPlayerContainer.classList.remove('hide');
+
+		utils.logger(logGroup, 'loading Anyclip asset', libraryUrl);
+
+		return utils.scriptLoader
+			.loadScript(libraryUrl, 'text/javascript', true, incontentPlayerContainer, this.params)
+			.then(() => {
+				utils.logger(logGroup, 'ready');
```
nandy-andy, line 38: Would unhiding `#incontent_player` here break anything in the integration? If not, I'd suggest to place it here as it could improve UX in these rare situations when for some reasons we fail loading the script ;)

src/ad-engine/utils/script-loader.ts
```diff
@@ -7,28 +7,32 @@ class ScriptLoader {
...
-				? document.getElementsByTagName('script')[0]
-				: (node as ChildNode) || document.body.lastChild;
 
 		script.async = isAsync;
 		script.type = type;
 		script.src = src;
 
 		Object.keys(parameters).forEach((parameter) => {
-			script[parameter] = parameters[parameter];
+			script.setAttribute(parameter, parameters[parameter]);
```
nandy-andy, line 20: 💎

src/ad-engine/utils/script-loader.ts
```diff
@@ -7,28 +7,32 @@ class ScriptLoader {
...
 		});
 
 		Object.keys(datasets).forEach((dataset) => {
 			script.dataset[dataset] = datasets[dataset];
 		});
 
-		temp.parentNode.insertBefore(script, temp);
+		if (typeof node === 'string') {
```
nandy-andy, line 27: Shouldn't it be `typeof node === 'string'` and `node === 'first'` to keep the previous logic? Although, it's odd too ;)

platforms/ucp-desktop/ad-context.ts
preludek: Forgot about setting enabled to false 🙈

src/ad-engine/utils/script-loader.ts
preludek: The old implementation was kind of weird. It wasn't setting the attributes properly, `node` had HTMLElement as a possible type but it never was so it was just to shut typescript up. I like how it is now given the incoming test but also I'm thinking about refactoring this class in a separate ticket :)

src/ad-services/anyclip/index.ts
preludek: I agree 🙈 I went with lodash since we are already using it in AdEngine anyway.

nandy-andy (approved):

PR #1259, opened 2022-02-07
ADEN-11427 Connatix integration

preludek:


platforms/ucp-desktop/ad-context.ts
```diff
@@ -84,6 +84,12 @@ export const basicContext = {
 			enabled: false,
 			propertyId: 'd-aIf3ibf0cYxCLB1HTWfBQOFEA',
 		},
+		connatix: {
+			enabled: false,
+			cid: '016551d5-7095-47c0-a46b-fd0cb9bf4c72',
+			playerId: '039a9ead-fb3b-4afc-bcfb-ed241bbaa8d1',
+			renderId: 'da3ec3e41d7e41209e5963444de28591',
```
Tomasz-at-Fandom, line 91: Why are we putting all of this in the context? Does it need to be reachable by some 3rd party in the `window`?
If not then why not just keeping it as params of the Connatix class

Tomasz-at-Fandom, line 91: Why are we putting all of this in the context? 
Does it need to be reachable by some 3rd party in the `window` or something?
If not then why not just keeping it as params of the Connatix class, except for the `enable` param, because it's entangled in some other logic that's better left alone for now.

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	private async loadPlayerAsset() {
+		const libraryUrl = `//cd.connatix.com/connatix.player.js?cid=${this.cid}`;
+
+		utils.logger(logGroup, 'loading Connatix asset', libraryUrl);
+
+		return utils.scriptLoader.loadScript(libraryUrl).then(() => {
+			utils.logger(logGroup, 'Connatix player is ready');
+		});
```
Tomasz-at-Fandom, line 40: In case the script doesn't load will this fail gracefully somehow?

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,72 @@
+import { context, utils } from '@ad-engine/core';
+
+const logGroup = 'connatix';
+
+class Connatix {
```
Tomasz-at-Fandom, line 5: It's a brand new class, let's test it! :D

platforms/ucp-desktop/ad-context.ts
preludek: We are keeping this in context since I remember so this one is for consistency sake 🙈  generally credentials are shared between skins, but sometimes for example mobile and desktop may have different id's etc.

src/ad-services/connatix/index.ts
preludek: AdEngine shouldn't crash. We will get an error calling the script and the player won't load and that's pretty much it.

platforms/ucp-desktop/ad-context.ts
dmnsobczak: Ticket to start context refactor: https://fandom.atlassian.net/browse/ADEN-11165

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,49 @@
+import { context, utils } from '@ad-engine/core';
+import { ConnatixPlayer } from './connatix-player';
+import { PlayerInjector, PlayerInjectorInterface } from './player-injector';
+import { ScriptLoaderInterface } from '../../ad-engine/utils';
```
dmnsobczak, line 4: You can export `ScriptLoaderInterface` in `utils/index.ts` and then add it to the `@ad-engine/core` import 🤔

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,49 @@
...
+		await this.loadPlayerAsset();
+		this.playerInjector.insertPlayerContainer(this.cid);
+	}
+
+	private async loadPlayerAsset(): Promise<void> {
+		const libraryUrl = `//cd.connatix.com/connatix.player.js?cid=${this.cid}`;
+
+		utils.logger(logGroup, 'loading Connatix asset', libraryUrl);
+
+		// TODO test if this would work if instead of .then we would use await
```
dmnsobczak, line 38: ToDo left here will create an issue in AdEngine repository: replace with follow up?

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,49 @@
...
+			utils.logger(logGroup, 'Connatix player is ready');
+		});
+	}
+}
+
+const cnxPlayer = new ConnatixPlayer();
+
+const playerInjector = new PlayerInjector(cnxPlayer);
+
+export const connatix = new Connatix(utils.scriptLoader, playerInjector);
```
dmnsobczak, line 49: 🙈

src/ad-services/connatix/player-injector.ts
```diff
@@ -0,0 +1,46 @@
+import { context, utils } from '@ad-engine/core';
+import { logGroup } from './index';
+import { ConnatixPlayerInterface } from './connatix-player';
+
+export interface PlayerInjectorInterface {
+	insertPlayerContainer(cid: string): void;
+}
+
+// TODO test this thing! >=D
```
dmnsobczak, line 9: Left accidentally? 🤔

dmnsobczak (approved):

src/ad-services/connatix/player-injector.ts
Tomasz-at-Fandom: Yea it was a reminder where to continue the next day, I'll remove this.

src/ad-services/connatix/index.ts
Tomasz-at-Fandom: Let's not bother with this, I was just curious if we could get rid of few brackets, but it turn out that not really so I'll just remove this comment.

src/ad-services/connatix/index.ts
Tomasz-at-Fandom: What? You like singletons don't you? :trollface:

src/ad-services/connatix/player-injector.ts
preludek: removed

src/ad-services/connatix/index.ts
preludek: removed

PR #1261, opened 2022-02-08
ADEN-11562 Adjusting NoAds mode

jakubbanas:


platforms/shared/modes/no-ads/no-ads.mode.ts
```diff
@@ -1,22 +1,23 @@
...
+		this.noAdsDetector.addReasons(window.ads.context.opts.noAdsReasons);
 		this.dispatchJWPlayerSetupAction();
 		this.trackAdEngineStatus();
 	}
 
 	private removeAdSlotsPlaceholders(): void {
 		const placeholders = document.querySelectorAll('.ad-slot-placeholder');
 		placeholders.forEach((placeholder) => {
-			placeholder.remove();
+			placeholder.parentElement.remove();
```
dmnsobczak, line 20: I'm worried about that. On mobile `bottom_leaderboard` ad placeholder don't have a wrapper, so this will remove it's parent, which is footer :/ And what about many other platforms? This should be heavy tested if NoAds layouts won't be corrupted.

dmnsobczak (changes requested):
Comment left regarding placeholders

platforms/shared/modes/no-ads/no-ads.mode.ts
jakubbanas: Good point, fixed 👍

dmnsobczak (approved):

PR #1262, opened 2022-02-08
ADEN-11576 Add internal tracking for nativo ads

nandy-andy:
This changes are supposed to deliver without breaking anything basic internal tracking for Nativo's slots: collapse and success statuses. This branch is a base branch for the big refactor in [ADEN-11578](https://fandom.atlassian.net/browse/ADEN-11578) in which we want to change Nativo from ad-service to ads provider but it requires rebuilding the AdEngine's core.


src/ad-services/nativo/index.ts
```diff
@@ -27,6 +27,18 @@ export class Nativo {
 			});
 	}
 
+	private handleNativoNativeEvent(e, slot: AdSlot, adStatus: string) {
+		utils.logger(logGroup, 'Nativo native event fired', e, adStatus);
+
```
dmnsobczak, line 32: This empty line suggest that something is missing here. Maybe some comment for future code? :D

src/ad-services/nativo/index.ts
```diff
@@ -52,11 +64,28 @@ export class Nativo {
...
+		slot.setConfigProperty('trackEachStatus', true);
+
+		slotService.add(slot);
 
 		window.ntv.Events?.PubSub?.subscribe('noad', (e) => {
-			this.sendNativoLoadStatus(AdSlot.STATUS_COLLAPSE, e);
+			this.handleNativoNativeEvent(e, slot, AdSlot.STATUS_COLLAPSE);
 		});
+
+		window.ntv.Events?.PubSub?.subscribe('adRenderingComplete', (e) => {
```
dmnsobczak, line 83: 💎

src/ad-services/nativo/index.ts
```diff
@@ -99,17 +128,20 @@ export class Nativo {
...
 		communicationService.dispatch(
-			communicationService.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)({
-				event: status,
-				payload: {
-					adLocation,
-					provider: 'nativo',
-				},
-				adSlotName: adLocation,
-			}),
+			communicationService.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)(payload),
```
dmnsobczak, line 144: This seems hacky, I hope that in the future this will be dispatched by the AdSlot class instance itself.

src/ad-services/nativo/index.ts
```diff
@@ -52,11 +64,28 @@ export class Nativo {
...
-		utils.logger(logGroup, 'Sending an ad request to Nativo');
+		utils.logger(logGroup, 'Sending an ad request to Nativo', placeholder.id);
+
+		this.createAdSlot(placeholder.id);
+		this.pushNativoQueue();
+	}
+
+	private createAdSlot(slotName: string): void {
+		const slot = new AdSlot({ id: slotName });
+		slot.setConfigProperty('trackEachStatus', true);
```
dmnsobczak, line 75: This can be set in the slots ad-context under `placeholder.id` key 🤔

dmnsobczak (approved):
Just some foods for thoughts and rhetorical questions.

src/ad-services/nativo/index.ts
nandy-andy: Darn, leftover after cleaning up a bit in https://github.com/Wikia/ad-engine/pull/1262/commits/41149c278fb50f326dd699fdf8d9e6ac7131c4e5 ;)

src/ad-services/nativo/index.ts
nandy-andy: Yes, this can be done like that or like this ;) I went with this solution as it's all in one place and with the refactor we can move it to the slot setup/base context.

src/ad-services/nativo/index.ts
nandy-andy: I didn't want to touch it as there is a risk the Pathfinder part will stop working. Definitely, we can change it in the future.

dmnsobczak:
Also @nandy-andy: please add labels :D

PR #1263, opened 2022-02-09
ADEN-11577 RCA for Nativo as an ad provider

nandy-andy:


docs/rca/a-sample-template.md
```diff
@@ -0,0 +1,25 @@
+# Title
```
nandy-andy, line 1: @ritaly FYI we're trying it out ;) Thanks for inspiration 🙇

Tomasz-at-Fandom (approved):

docs/rca/a-sample-template.md
ritaly: May it serve you well 🚀

PR #1267, opened 2022-02-11
ADEN-11531 Adding Load Time tracking

dmnsobczak:


platforms/shared/tracking/tracking.setup.ts
```diff
@@ -273,6 +274,30 @@ export class TrackingSetup {
 		}
 	}
 
+	private loadTimeTracker(): void {
+		const trackerConfig = context.get('options.loadTimeTracking');
+
+		if (!trackerConfig || !trackerConfig.enabled) {
```
nandy-andy, line 280: Good idea! This way we could extend it with icVar and just enable it for sampled traffic! 👍

platforms/ucp-mobile/ad-context.ts
```diff
@@ -122,6 +122,11 @@ export const basicContext = {
 		customAdLoader: {
 			globalMethodName: 'loadCustomAd',
 		},
+		loadTimeTracking: {
+			enabled: true,
+			timing: ['responseEnd', 'domContentLoadedEventStart'],
```
nandy-andy, line 127: Consider if we really need it and if we can't use anything from the data [we're already gathering already](https://github.com/Wikia/unified-platform/blob/4b62b122156ec189444c0e1117f598812a79f713/front/scripts/shared/performance.js) that is stored in DW in `statsdb.performance_metrics`. Removing it we make two requests less on the page load but maybe it's worth keeping it here as it supports all the platforms 🤔 Worth discussing within the team, I guess.

nandy-andy (approved):
Looks like good point to start with, check data and see if we need less/more 👍 Consider releasing it for a sampled traffic.

PR #1268, opened 2022-02-11
ADEN-11578 Change Nativo from a service to a proper ad provider

nandy-andy:


platforms/shared/modes/start-ad-engine.ts
```diff
@@ -1,13 +1,51 @@
...
+	);
+}
+
 export function startAdEngine(inhibitors: Promise<any>[] = []): void {
 	if (!adEngineInstance) {
 		const GPT_LIBRARY_URL = '//www.googletagservices.com/tag/js/gpt.js';
-
 		utils.scriptLoader.loadScript(GPT_LIBRARY_URL);
 
+		if (context.get('services.nativo.enabled') && context.get('wiki.opts.enableNativeAds')) {
```
nandy-andy, line 33: We could have GPT loader and Nativo loader that would keep additional logic and load the assets when some conditions are met.

src/ad-engine/ad-engine.ts
```diff
@@ -64,29 +64,53 @@ export class AdEngine {
...
 
+	private fillWithProvidersChain(adSlot) {
+		const providersChain = context.get(`slots.${adSlot.getSlotName()}.providers`);
+
+		while (providersChain.length > 0) {
+			const providerName = providersChain.shift();
+
+			const provider = this.createProvider(providerName);
+
+			if (provider.fillIn(adSlot)) {
```
nandy-andy, line 94: Right now only Nativo returns `true` other providers (GPT and Prebidium) are more complicated and I'm not sure but probably we'll need to resign from this idea. We could replace it with having all the providers to respond with something and we pick at this point what gets rendered ;)

Tomasz-at-Fandom:
I could not make up my mind where to start this review in code, so I'm starting with this general comment.
Two things that worry me about the direction of these changes are:
1. We remove tests of the old implementation without replacing them with tests for the new one.
2. The old implementation encapsulated Nativo better than this one, which is spread wider across the codebase.
I'll follow this up with comments to specific points in code, so please keep in mind when reading them that what I'm trying to do is to find a way to better encapsulate this provider and his features and make the code easier to test at a granular level.

platforms/shared/modes/start-ad-engine.ts
Tomasz-at-Fandom: So how about moving all this newly added code to a separate class within the Nativo Module to reduce Nativos spread across the existing codebase?

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -202,39 +200,49 @@ export class UcpDesktopSlotsDefinitionRepository {
 		return !this.instantConfig.get('icFloorAdhesion') && !context.get('custom.hasFeaturedVideo');
 	}
 
+	private isNativoEnabled() {
+		return context.get('services.nativo.enabled') && context.get('wiki.opts.enableNativeAds');
+	}
```
Tomasz-at-Fandom, line 205: This is something that could live in the Nativo Module

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -202,39 +200,49 @@ export class UcpDesktopSlotsDefinitionRepository {
...
 
 		return {
 			activator: () => {
-				fanFeedNativeAdListener((uapLoadStatusAction: any = {}) =>
-					nativo.requestAd(document.getElementById(Nativo.FEED_AD_SLOT_NAME), uapLoadStatusAction),
-				);
+				fanFeedNativeAdListener(() => context.push('state.adStack', { id: slotName }));
 			},
 		};
 	}
```
Tomasz-at-Fandom, line 248: Actually all of this could live in the Nativo Module, let's not bloat this "repository" with  every provider we want to add.

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -92,45 +90,51 @@ export class UcpMobileSlotsDefinitionRepository {
...
+				classList: ['ntv-ad', 'hide'],
 			},
 			activator: () => {
-				fanFeedNativeAdListener((uapLoadStatusAction: any = {}) =>
-					nativo.replaceAndShowSponsoredFanAd(uapLoadStatusAction),
-				);
+				fanFeedNativeAdListener(() => context.push('state.adStack', { id: slotName }));
 			},
 		};
 	}
```
Tomasz-at-Fandom, line 140: Like in the desktop version, perhaps we could move this to a Nativo Module.

src/ad-engine/providers/nativo-provider.ts
```diff
@@ -0,0 +1,42 @@
+import { AdSlot } from '../models';
+import { Provider } from './provider';
+import { logger } from '../utils';
+
+const logGroup = 'nativo-provider';
+
+export class NativoProvider implements Provider {
+	constructor() {
+		window.ntv = window.ntv || {};
+		window.ntv.cmd = window.ntv.cmd || [];
```
Tomasz-at-Fandom, line 10: If we inject this through constructors parameters instead of fetching from window we could easily test this class and not even bother with instantiating it in case `window.ntv` is not even set

@types/window/index.d.ts
```diff
@@ -29,7 +29,7 @@ interface Window {
 	moatjw?: MoatJW;
 	moatPrebidApi?: MoatPrebidApi;
 	mw?: MediaWiki;
-	ntv?: NativoQueue;
+	ntv?: NativoSdk;
```
dmnsobczak, line 32: It's more like an API than SDK 🤔

src/ad-engine/ad-engine.ts
```diff
@@ -64,29 +64,51 @@ export class AdEngine {
...
+				break;
+			}
+		}
+	}
+
+	private createProvider(providerName: string) {
```
dmnsobczak, line 100: This seems over-splitted, but it's fine :D

src/ad-engine/ad-engine.ts
```diff
@@ -64,29 +64,51 @@ export class AdEngine {
...
+					this.fillWithProvidersChain(adSlot);
+				} else {
+					this.defaultProvider.fillIn(adSlot);
+				}
 			});
 		}
 	}
 
+	private fillWithProvidersChain(adSlot) {
+		const providersChain = context.get(`slots.${adSlot.getSlotName()}.providers`);
```
dmnsobczak, line 87: This context key is used twice - we could pass providers like an argument 🤔

dmnsobczak: Also: `... || []`?

src/ad-engine/ad-engine.ts
```diff
@@ -64,29 +64,51 @@ export class AdEngine {
...
+
+		while (providersChain.length > 0) {
+			const providerName = providersChain.shift();
+
```
dmnsobczak, line 91: Unneeded line break :>

src/ad-engine/ad-engine.ts
dmnsobczak: I was surprised by the `chain` logic while we do not have any use case where actual chain is used. My proposal is to resign from `chain` for now and re-add it in the future if needed.

platforms/shared/modes/start-ad-engine.ts
```diff
@@ -1,13 +1,18 @@
-import { AdEngine, AdSlot, communicationService, utils } from '@wikia/ad-engine';
+import { AdEngine, AdSlot, context, communicationService, Nativo, utils } from '@wikia/ad-engine';
 
 let adEngineInstance: AdEngine;
 
 export function startAdEngine(inhibitors: Promise<any>[] = []): void {
 	if (!adEngineInstance) {
+		const nativo = new Nativo(context);
```
dmnsobczak, line 7: This makes `Nativo` class as important as `AdEngine` class. Maybe let's move it to `setupProviders()` inside AdEngine class? It would be executed on `AdEngine::init` and logically it will be in the better place. Also `Nativo` class constructor could be executed only once 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+	eventsRepository,
+	Nativo,
+	nativoLazyLoader,
+	UapLoadStatus,
+} from '@wikia/ad-engine';
+import { fanFeedNativeAdListener, SlotSetupDefinition } from '@platforms/shared';
+
+@Injectable()
+export class UcpDesktopNativoSlotsDefinitionRepository {
+	constructor(protected domListener: DomListener, protected nativo: Nativo) {
```
dmnsobczak, line 16: `, protected nativo: Nativo` is redundant

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,68 @@
...
+	eventsRepository,
+	Nativo,
+	nativoLazyLoader,
+	UapLoadStatus,
+} from '@wikia/ad-engine';
+import { fanFeedNativeAdListener, SlotSetupDefinition } from '@platforms/shared';
+
+@Injectable()
+export class UcpMobileNativoSlotsDefinitionRepository {
+	constructor(protected nativo: Nativo, protected domListener: DomListener) {
```
dmnsobczak, line 16: `protected nativo: Nativo, ` can be removed

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+		this.nativo = new Nativo(context);
```
dmnsobczak, line 17: Actually: can we make `Nativo` class instance Injectable? 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+		nativoLazyLoader.scrollTrigger(slotName, scrollThreshold, this.domListener, () =>
+			this.nativo.scrollTriggerCallback(action, slotName),
+		);
+	}
+
+	getNativoIncontentAdConfig(): SlotSetupDefinition {
+		if (!this.nativo.isEnabled()) {
+			return;
+		}
+		const slotName = Nativo.INCONTENT_AD_SLOT_NAME;
```
dmnsobczak, line 30: Assigning a static variable to a variable - why don't we use a static variable directly? :D

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+
+		return {
+			slotCreatorConfig: {
+				slotName,
+				anchorSelector: '.mw-parser-output > h2:nth-of-type(2)',
+				insertMethod: 'before',
+				classList: ['ntv-ad', 'ad-slot'],
+			},
+			activator: () => {
+				const scrollThreshold = context.get('events.pushOnScroll.nativoThreshold') || 200;
```
dmnsobczak, line 40: This variable was already defaulted as `200` multiple times, ` || 200;` is reduntant

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+				);
+			},
+		};
+	}
+
+	getNativoFeedAdConfig(): SlotSetupDefinition {
+		if (!this.nativo.isEnabled()) {
+			return;
+		}
+		const slotName = Nativo.FEED_AD_SLOT_NAME;
```
dmnsobczak, line 54: `slotName` is used only once :trollface:

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		return {
+			slotCreatorConfig: {
+				slotName: slotName,
+				anchorSelector: '.mw-parser-output > h2:nth-of-type(4n)',
+				insertMethod: 'before',
+				classList: ['ntv-ad', 'ad-slot'],
+			},
+			activator: () => {
+				const scrollThreshold = context.get('events.pushOnScroll.nativoThreshold') || 200;
```
dmnsobczak, line 40: Some lines here as well as definition of `onAdEngineUapLoaded()` method are duplicated - some refactor of a refactor? Code can be always stored as ucp-shared util inside `platforms/shared/` directory.

src/ad-engine/providers/nativo/nativo-provider.ts
```diff
@@ -0,0 +1,28 @@
...
+		logger(logGroup, `Filling ${slot.getSlotName()} by pushing Nativo's command queue`);
+
+		this.getQueue().push(() => {
+			window.PostRelease.Start();
+		});
+
+		return true;
+	}
+
+	getQueue(): Array<Dictionary> {
```
dmnsobczak, line 25: `cmd?: any[];` - can you choose the same type? It can be just `Dictionary[]` 🤔

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,92 @@
...
+		1142669: Nativo.FEED_AD_SLOT_NAME,
+	};
+
+	constructor(protected context: Context) {}
+
+	isEnabled() {
+		const isEnabled =
+			this.context.get('services.nativo.enabled') && this.context.get('wiki.opts.enableNativeAds');
+
+		logger(logGroup, 'Is Nativo enabled?', isEnabled);
```
dmnsobczak, line 25: This log with unchanged message will be console.logged multiple times 🤔

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		return isEnabled;
+	}
+
+	load() {
+		logger(logGroup, 'Loading Nativo SDK...');
+		const NATIVO_LIBRARY_URL = '//s.ntv.io/serve/load.js';
```
dmnsobczak, line 32: Usually we make variables like that a variable outside the class, not a local variable.

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,92 @@
...
+				this.sendNativoLoadStatus(AdSlot.SLOT_ADDED_EVENT);
+			});
+	}
+
+	scrollTriggerCallback(action: UapLoadStatus, slotName: string) {
+		if (!action.isLoaded && action.adProduct !== 'ruap') {
+			this.context.push('state.adStack', { id: slotName });
+		} else {
+			logger(logGroup, "UAP or UAP:Roadblock on page - don't display Nativo");
+		}
```
dmnsobczak, line 48:
```suggestion
		if (action.isLoaded || action.adProduct === 'ruap') {
			logger(logGroup, "UAP or UAP:Roadblock on page - don't display Nativo");
			return;
		}
			
		this.context.push('state.adStack', { id: slotName });
```

@types/window/index.d.ts
nandy-andy: What's the difference? ;)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
nandy-andy: You mean redundant because of the line below? 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
nandy-andy: Damian once told me he likes the pattern of having `slotName` defined at the beginning of the definition repository methods ;)

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
nandy-andy: But just-in-case... ;)

src/ad-engine/ad-engine.ts
nandy-andy: It's used now in two places instead of one. However, if we want to get rid off of the chain we can make it as it used to be 👌 ;)

src/ad-engine/providers/nativo/nativo.ts
nandy-andy: Yep.

@types/window/index.d.ts
dmnsobczak: I always understand SDK like a set of tools that helps to create an application, but are not present in the finished build, while the API allows two apps to communicate: AdEngine and Nativo library in this example 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
dmnsobczak: Yes, `protected nativo` creates an class object and usually it should be assigned to a DI injected value. Line below can also create a class field with `private` access level and while nothing is inheriting from this class, then `protected` is actually a `private` 🤔

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
dmnsobczak: Dependency Injection is available only in `platforms/` so `Nativo` class should be moved here, but that's not strictly required. I was thinking about binding it to container via `DiProcess` interface. You can use InstantConfig as an example :D 
https://github.com/Wikia/ad-engine/blob/3b90faf7958f9f09885e17f7c88c1ff8d92ff4bb/platforms/shared/utils/instant-config-setup.ts#L23

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-nativo-slots-definition-repository.ts
dmnsobczak: Yes, but then he admitted that reviewing person was right and he abandoned that idea :trollface:

@types/window/index.d.ts
nandy-andy: Sure, we can go this way :) My understanding was that whatever comes from Nativo's lib is a set of tools - there is a queue, an API to request something from the ad server and an API with events. But I have no strong feelings about SDK vs. API. There are two votes for changing it to API, so let's do it :)

src/ad-engine/providers/nativo/lazy-loader.ts
```diff
@@ -0,0 +1,46 @@
...
+			.pipe(
+				filter(() => this.isThresholdExceeded(slotName, threshold)),
+				take(1),
+				tap(callback),
+			)
+			.subscribe();
+	}
+
+	private isThresholdExceeded(slotName: string, threshold: number): boolean {
+		const scrollPosition =
```
preludek, line 27: nitpick - you are declaring a type for 2 out of 3 consts :D since those are all primitive types and typescript is inferring them, I'd just skip them for all 3.

platforms/shared/dynamic-slots/nativo-slots-definition-repository.ts
```diff
@@ -0,0 +1,67 @@
...
+	context,
+	DomListener,
+	eventsRepository,
+	Nativo,
+	nativoLazyLoader,
+	UapLoadStatus,
+} from '@wikia/ad-engine';
+
+import { fanFeedNativeAdListener } from './fan-feed-native-ad-listener';
+import { SlotSetupDefinition } from './../utils/insert-slots';
```
preludek, line 15: nitpick - `'../utils/insert-slots'`

spec/ad-engine/providers/nativo/nativo.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { expect } from 'chai';
+import { spy } from 'sinon';
+
+import { Context, Nativo, UapLoadStatus } from '../../../../src';
+
+describe('Nativo', () => {
+	describe('module', () => {
+		it('is enabled in instant-config and (by default) per wiki', () => {
+			const contextMock = new Context();
```
preludek, line 9: I think this will leak between units. We need to clear that after this spec.

spec/ad-engine/providers/nativo/nativo.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+		});
+	});
+
+	describe('scrollTriggerCallback', () => {
+		let context;
+		let contextPushSpy;
+		let nativo;
+
+		before(() => {
+			context = new Context();
```
preludek, line 42: ditto

src/ad-engine/ad-engine.ts
```diff
@@ -80,13 +77,33 @@ export class AdEngine {
...
 		}
 	}
 
+	private createProvider(providerName: string) {
+		switch (providerName) {
+			case 'prebidium':
+				return new PrebidiumProvider();
+			case 'nativo':
+				return new NativoProvider(window.ntv);
+			case 'gpt':
```
preludek, line 101: since `gpt` is default this line seems redundant 🤔

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,93 @@
...
+	scrollTriggerCallback(action: UapLoadStatus, slotName: string) {
+		if (action.isLoaded || action.adProduct === 'ruap') {
+			logger(logGroup, "UAP or UAP:Roadblock on page - don't display Nativo");
+			return;
+		}
+
+		this.context.push('state.adStack', { id: slotName });
+	}
+
+	sendNativoLoadStatus(status: string, event?: any): void {
```
preludek, line 52: This method can be private

preludek: And we never pass the second argument so it seems we can 🔪  it

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,93 @@
...
+			this.handleNtvNativeEvent(e, slotName, AdSlot.STATUS_COLLAPSE); // init or collapse
+		});
+
+		window.ntv.Events?.PubSub?.subscribe('adRenderingComplete', (e: NativoCompleteEvent) => {
+			const slotName = Nativo.AD_SLOT_MAP[e.data.placement];
+			this.handleNtvNativeEvent(e, slotName, AdSlot.STATUS_SUCCESS);
+		});
+	}
+
+	private handleNtvNativeEvent(e, slotName: string, adStatus: string) {
```
preludek, line 80: e has any type here, we can be more specific

spec/ad-engine/providers/nativo/nativo-provider.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { expect } from 'chai';
+import { adSlotFake } from '../../ad-slot-fake';
+import { NativoProvider } from '../../../../src/ad-engine/providers/nativo/nativo-provider';
+
+describe('Nativo Provider', () => {
+	beforeEach(() => {
```
preludek, line 6: I think this will also leak if we didn't reset it in afterEach.

preludek (changes requested):

spec/ad-engine/providers/nativo/nativo.spec.ts
nandy-andy: we only spy here and then we reset the spy, so I think it's safe 🤔

nandy-andy: @Tomasz-at-Fandom I think that reseting the spy's history in `afterEach()` prevents the case you described from happening 🤔 Let me know if my understanding is incorrect 😊

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -44,15 +49,20 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 			this.slotsDefinitionRepository.getFloorAdhesionConfig(),
 			this.slotsDefinitionRepository.getInvisibleHighImpactConfig(),
 			this.slotsDefinitionRepository.getInterstitialConfig(),
-			this.slotsDefinitionRepository.getNativoFeedAdConfig(),
+			this.nativoSlotDefinitionRepository.getNativoFeedAdConfig({
+				slotName: Nativo.FEED_AD_SLOT_NAME,
+				anchorSelector: '.recirculation-prefooter',
+				insertMethod: 'before',
+				classList: ['ntv-ad', 'hide'],
+			}),
```
szymon-nawrocki, line 65: Can't we set the config in `getNativoFeedAdConfig` function to be consistent with another slots?

nandy-andy: Not really or more: not right now. Not in this scope change. We need to pass this configuration in order to create slot in DOM for FandomMobile.

szymon-nawrocki (approved):

spec/ad-engine/providers/nativo/nativo.spec.ts
Tomasz-at-Fandom: Oh I did not notice the after each, sorry!

dmnsobczak (approved):

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -0,0 +1,118 @@
...
+		logger(logGroup, 'Nativo native event fired', e, adStatus, slotName, slot);
+
+		if (!slot || slot.getSlotName() !== slotName) return;
+
+		if (adStatus === AdSlot.STATUS_COLLAPSE) {
+			slot.hide();
+		}
+
+		if (adStatus === AdSlot.STATUS_SUCCESS && slotName === Nativo.FEED_AD_SLOT_NAME) {
+			this.replaceAndShowSponsoredFanAd();
```
nandy-andy, line 109: Follow-up: https://fandom.atlassian.net/browse/ADEN-11626

spec/ad-engine/providers/nativo/nativo.spec.ts
```diff
@@ -68,21 +69,36 @@ describe('Nativo', () => {
...
 			expect(contextPushSpy.calledWith('state.adStack', { id: 'mocked_slot' })).to.be.false;
 		});
 
-		it('does not push slot when UAP:Roadblock is on the page', () => {
+		it('does not push slot when UAP is on the page', () => {
 			const uapLoadStatusActionMock: UapLoadStatus = {
-				isLoaded: false,
-				adProduct: 'ruap',
+				isLoaded: true,
+				adProduct: 'nothing important',
```
preludek, line 101: 🙈

preludek (approved):

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -44,15 +49,20 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 			this.slotsDefinitionRepository.getTopBoxadConfig(),
 			this.slotsDefinitionRepository.getIncontentBoxadConfig(),
 			this.slotsDefinitionRepository.getBottomLeaderboardConfig(),
 			this.slotsDefinitionRepository.getMobilePrefooterConfig(),
 			this.slotsDefinitionRepository.getFloorAdhesionConfig(),
 			this.slotsDefinitionRepository.getInvisibleHighImpactConfig(),
 			this.slotsDefinitionRepository.getInterstitialConfig(),
-			this.slotsDefinitionRepository.getNativoFeedAdConfig(),
+			this.nativoSlotDefinitionRepository.getNativoFeedAdConfig({
+				slotName: Nativo.FEED_AD_SLOT_NAME,
```
dmnsobczak, line 61: I would rather move it to the Slot Definition Repository file - placing it here looks weird 🤔 Alternatively it can be stored in the context or something.

src/ad-engine/ad-engine.ts
```diff
@@ -86,14 +86,11 @@ export class AdEngine {
...
+		this.defaultProvider = this.createProvider(providerName);
 
-		switch (providerName) {
-			case 'prebidium':
-				this.provider = new PrebidiumProvider();
-				break;
-			case 'gpt':
-			default:
-				this.provider = new GptProvider();
+		const nativo = new Nativo(context);
```
dmnsobczak, line 91: Can we move it to `createProvider()` method below `case 'nativo':` line or after the switch? Initializing Nativo has nothing to do with setting up providers 🤔

dmnsobczak (approved):

spec/ad-engine/providers/nativo/nativo.spec.ts
nandy-andy: It's a test. What's the matter?

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
nandy-andy: There is a ticket for dealing with it: https://fandom.atlassian.net/browse/ADEN-11626 a bit out of scope for this PR but good catch 👍

src/ad-engine/ad-engine.ts
nandy-andy: Right. In the future it's something that the ProvidersManager/AdLayoutBuilder/LIS will take care of 👌

nandy-andy: Darn, I can't move it because moving it `createProvider()` means we'll load the Nativo API every time we create a provider for a slot. Maybe making `isEnabled()` public static will be a solution - no need to create an instance of Nativo class 🤔

nandy-andy: Ugh, it's complicated. We'll change it while working on AdLayoutBuilder hopefully 🤞

PR #1270, opened 2022-02-15
Implement Identity Hub Script

jakubbanas:


src/ad-bidders/prebid/identity-hub/identity-hub.ts
```diff
@@ -0,0 +1,41 @@
...
+const logGroup = 'IdentityHub';
+
+class IdentityHub {
+	private isLoaded = false;
+	private identityHubScriptSrc = `https://ads.pubmatic.com/AdServer/js/pwt/${pubId}/${profileId}/pwt.js`;
+
+	call(): Promise<void> {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+		}
```
szwajkowski, line 15: i guess we should return here if it is disabled

src/ad-bidders/prebid/identity-hub/identity-hub.ts
```diff
@@ -0,0 +1,41 @@
...
+
+		utils.logger(logGroup, 'enabled');
+
+		if (!this.isLoaded) {
+			window.PWT = {
+				jsLoaded: () => {
```
szwajkowski, line 21: I don't get it, what's the purpose of it? It just logs that it is loaded. And does it need to be a part of `window` ?

src/ad-bidders/prebid/identity-hub/identity-hub.ts
jakubbanas: Right, good catch!

src/ad-bidders/prebid/identity-hub/identity-hub.ts
jakubbanas: It comes with the vendor script, after the code is executed and Identity Hub is "installed" it calls this function from the window so we know that we could proceed with GAM.

I'll put a comment there, it's confusing.

szwajkowski: nice, that's better

szwajkowski (approved):
LGTM :)

PR #1271, opened 2022-02-15
ADEN-11545 Disable StickyTLB if AdBlock is enabled

dmnsobczak:


szwajkowski (approved):
LGTM :)

preludek:
Did you consider doing it via communication service 🤔 ?  Dispatch event on ad block detection and consume it here. I feel it would be a bit cleaner. @dmnsobczak

dmnsobczak:
> Did you consider doing it via communication service 🤔 ? Dispatch event on ad block detection and consume it here. I feel it would be a bit cleaner. @dmnsobczak

But in that case I would need to catch the event and store it's value in some new variable like `private isAdBlock`, while this value is already stored in `BabDetection` class 🤔

PR #1274, opened 2022-02-17
ADEN-11560 implemented outstream experiment logic

szwajkowski:


szwajkowski:
Need to merge https://github.com/Wikia/ad-engine/pull/1259 first !!!11

bckatarzyna (approved):

platforms/shared/experiments/outstream-experiment.ts
```diff
@@ -0,0 +1,27 @@
+import { InstantConfigService } from '@wikia/ad-engine';
+
+export class OutstreamExperiment {
```
mfarul, line 3: that's nice class, i would love to have a generic base class for experiment f.e.
`abstract class MultivariantExperiment {
     private readonly enabled: boolean;
     private readonly selectedVariantName: string;
    
     constructor(protected instantConfig: InstantConfigService, expEnabledVariableName: string, expDefaultVariantVariableName: string) { ... }

     public isVariantSelected(variantName) : boolean { ... } 
`

szwajkowski: done, added to cases for single and multi variant experiments

platforms/shared/experiments/experiment.ts
```diff
@@ -0,0 +1,38 @@
+import { InstantConfigServiceInterface } from '@wikia/ad-engine';
+
+export interface Experiment {
+	enabled: boolean;
+}
+
+export abstract class SingleVariantExperiment implements Experiment {
+	private readonly _enabled: boolean;
```
dmnsobczak, line 8: Finally first interface which is actually an interface xD It looks weird, because we don't use variables like that in AdEngine project, but it's not wrong 🤷

dmnsobczak (approved):

PR #1276, opened 2022-02-17
ADEN-11585 Rafactoring custom companion sizes

dmnsobczak:


platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
```diff
@@ -27,14 +23,10 @@ export class BfaaF2ConfigHandler implements TemplateStateHandler {
...
-				: universalAdPackage.UAP_ADDITIONAL_SIZES.desktop;
+		const bfaSize = context.get('state.isMobile')
+			? universalAdPackage.UAP_ADDITIONAL_SIZES.bfaSize.mobile
+			: universalAdPackage.UAP_ADDITIONAL_SIZES.bfaSize.desktop;
 
-		slotsContext.setupSlotSizeOverwriting(additionalSizes);
-		slotsContext.addSlotSize('top_boxad', additionalSizes.companionSize);
-		slotsContext.addSlotSize('incontent_boxad', additionalSizes.companionSize);
-		slotsContext.setSlotSize('bottom_leaderboard', additionalSizes.bfaSize);
+		slotsContext.setSlotSize('bottom_leaderboard', bfaSize);
```
szymon-nawrocki, line 30: Won't it be better to set it in _configureTopLeaderboardAndCompanions()_ to keep setting slot sizes in one place?

szymon-nawrocki: Then it will be good to split _configureTopLeaderboardAndCompanions_ into 2 separate functions:
- configureTopLeaderboard
- configureCompanions

platforms/sports/templates/handlers/bfaa/bfaa-sports-config-handler.ts
```diff
@@ -23,10 +23,11 @@ export class BfaaSportsConfigHandler implements TemplateStateHandler {
 		);
 		context.set('slots.cdm-zone-04.viewportConflicts', []);
 
-		const additionalSizes = universalAdPackage.UAP_ADDITIONAL_SIZES.desktop;
-
-		slotsContext.setupSlotSizeOverwriting(additionalSizes);
-		slotsContext.addSlotSize('cdm-zone-02', additionalSizes.companionSize);
-		slotsContext.setSlotSize('cdm-zone-04', additionalSizes.bfaSize);
+		slotsContext.setSlotSize(
```
szymon-nawrocki, line 26: Same as above, won't it be better to set it in _configureUap()_ method?

platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
dmnsobczak: No, because BLB additional size should be set only if UAP is loaded - that's why this code is in `bfaa-config-handler`. While the rest of the companions should have additional sizes always :D

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -107,21 +107,18 @@ function getAdType(
 }
 
 function adjustIframeSize(adSlot: AdSlot): void {
-	const map: Dictionary<[number, number]> = context.get('templates.sizeOverwritingMap');
+	const map = context.get('templates.sizeOverwritingMap');
```
szymon-nawrocki, line 110: How about using more descriptive name like "sizeMap" ?

dmnsobczak: Good idea! Changed!

szymon-nawrocki (approved):

PR #1278, opened 2022-02-18
ADEN-11598 Disable slot repeating if it was force_collapsed

dmnsobczak:


szymon-nawrocki (approved):
Very clever solution to the problem

PR #1286, opened 2022-02-23
ADEN-11434 Adding LIS: Layout Initializer Slot

dmnsobczak:


platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
```diff
@@ -20,6 +20,24 @@ import { Injectable } from '@wikia/dependency-injection';
 export class UcpDesktopSlotsDefinitionRepository {
 	constructor(protected instantConfig: InstantConfigService, protected domListener: DomListener) {}
 
+	getLayoutInitializerConfig(): SlotSetupDefinition {
```
nandy-andy, line 23: I'm thinking of getting rid-off of the code changes in `*-dynamic-slots.setup.ts` and `*-slots-definition-repository.ts` are not needed. We just want to call it [in `runInitStack()`](https://github.com/Wikia/ad-engine/pull/1286/files#diff-83cc4163d820af6fdec14e8344e30f59a2ce8946fa5953ffcd7ad47b5711a5baR108) when [it's enabled](https://github.com/Wikia/ad-engine/pull/1286/files#diff-d05eb3a1d334df7aadbc295389a14c993486c2688408a449d8bd20cb02f42338R162).

src/ad-engine/models/ad-slot.ts
```diff
@@ -377,6 +378,10 @@ export class AdSlot {
 		return this.enabled;
 	}
 
+	isInitCall(): boolean {
```
nandy-andy, line 381: I suggest renaming it to `isInitCallSlot()` a bit longer yet not confusing that a slot could be a call ;)

src/ad-engine/utils/tagless-request-url-builder.ts
```diff
@@ -108,5 +124,19 @@ export function buildVastUrl(
 
 	params.push(`rdp=${trackingOptIn.isOptOutSale() ? 1 : 0}`);
 
-	return baseUrl + params.join('&');
+	return vastBaseUrl + params.join('&');
+}
+
+export function buildTaglessRequestUrl(
```
nandy-andy, line 130: Perfect for a unit test <3 You just mock `adSlot` (we have a `fakeAdSlot` in `spec` directory), mock options and check if the outcome is what we expect :)

src/ad-engine/ad-engine.ts
```diff
@@ -87,6 +95,36 @@ export class AdEngine {
...
+
+		return await scriptLoader.loadAsset(buildTaglessRequestUrl(adSlot), 'text').then((response) => {
+			if (!response) {
+				return inhibitors;
+			}
+
+			try {
+				const layoutPayload = JSON.parse(response);
+
+				if (layoutPayload.layout === 'uap') {
```
nandy-andy, line 116: Does it mean we'll be returning a custom code creative in GAM for calls to LIS with a JSON? We could create a LayoutPayload type and define how we think it'll look like. Currently it looks like:
```
{
  layout: string,
  data: {
    lineItemId: number,
    creativeId: number
  }
}
```

src/ad-engine/ad-engine.ts
```diff
@@ -87,6 +95,36 @@ export class AdEngine {
...
 
+	private async runInitStack(inhibitors: Promise<any>[] = []): Promise<Promise<any>[]> {
+		if (!context.get('options.initCall')) {
+			return inhibitors;
+		}
+
+		const slotName = context.get('state.initSlot');
+		const adSlot = new AdSlot({ id: slotName });
+
+		slotService.add(adSlot);
```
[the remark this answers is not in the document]
dmnsobczak: `slotService` is not only a slot manager, but also a slot repository. By adding a slot to it we can easily access it by requesting `slotService.get(slotName)`. I would like to treat it like a normal slot, not a direct HTTPRequest hacky-something :D

platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts
dmnsobczak: From GAM perspective LIS is a slot with `pos`, ad unit and impressions, so I decided to treat it the same way in AdEngine. I know that this method looks very simple, but will be expanded in the future, because slots repository is a great place to store slot-related custom logic. It also have a whole bunch of existing methods, which I don't need to put somewhere (like `activator()`) and it's exactly in the place I want it to be from the RxJS pipeline perspective :)

src/ad-engine/ad-engine.ts
dmnsobczak: Done!

src/ad-engine/models/ad-slot.ts
dmnsobczak: I don't understand :( We use the same convention for `isFirstCall()`

src/ad-engine/services/btf-blocker-service.ts
```diff
@@ -91,6 +91,10 @@ class BtfBlockerService {
 	}
 
 	push(adSlot: AdSlot, fillInCallback: FillInCallback): void {
+		if (adSlot.isInitCall()) {
```
nandy-andy, line 94: Per our discussion offline: we can get rid off this condition because we're safe as long as LIS is not added to any queue.

src/ad-engine/models/ad-slot.ts
nandy-andy: Yep, doesn't make sense either after the years 🤷‍♂️ A slot is not a first or init call, right? A slot is an object, possibly a representation of a div on a page and it can be connected with an init or first call to an ad server. Since it's unified with the previous methods we can leave it if you don't feel it. Maybe we'll feel it better in the future and then change it ;)

src/ad-engine/utils/tagless-request-url-builder.ts
dmnsobczak: Done!

szymon-nawrocki (approved):

platforms/shared/setup/base-context.setup.ts
```diff
@@ -159,9 +159,11 @@ export class BaseContextSetup implements DiProcess {
 	private setMiscContext(): void {
 		this.instantConfig.get('icLABradorTest');
 
+		context.set('options.initCall', this.instantConfig.get('icLayoutInitializerSlot'));
```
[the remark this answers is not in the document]
dmnsobczak: Agree :< Ticket is ready, we just need to take it into the sprint 💪 
https://fandom.atlassian.net/browse/ADEN-11165

src/ad-engine/ad-engine.ts
```diff
@@ -87,6 +95,36 @@ export class AdEngine {
...
+
+				if (layoutPayload.layout === 'uap') {
+					context.set('targeting.uap', layoutPayload.data.lineItemId);
+					context.set('targeting.uap_c', layoutPayload.data.creativeId);
+				}
+			} catch (e) {
+				return inhibitors;
+			}
+
+			return [];
```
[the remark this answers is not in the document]
dmnsobczak: Inhibitors were designed to return a simple Promise, which will be fulfilled or timed out based on the inhibitor type - for example if all bidders will finish bidding or after 2000ms. The array of those Promises then can be easily processed later by `Promise.all()` or `Promise.race()` methods. Passing the array to the initCall logic allow to return unchanged array of pending/finished Promises or empty array, which will not delay the AdEngine. If this logic should be separated then we would need to expand the inhibitors logic to allow terminating the promises on some signal. We could do this using Post-QUecast events but that would be only shifting the logic. That's why I decided to do this like that without expanding the scope of this ticket 🤔

PR #1288, opened 2022-03-01
ADEN-11438 bump webpack

preludek:


src/ad-engine/runner/runner.ts
```diff
@@ -10,6 +11,8 @@ export class Runner {
 	) {}
 
 	async waitForInhibitors(): Promise<void> {
+		logVersion();
```
preludek, line 14: For whatever reason this didnt get called in it's previous place. Not sure if this is the right place for it 🤔

spec/ad-services/nativo/nativo.spec.ts
```diff
@@ -5,7 +5,7 @@ import { createSandbox, SinonSpy } from 'sinon';
 import { AdSlot, context, utils } from '../../../src/ad-engine';
 import { createHtmlElementStub } from '../../spec-utils/html-element.stub';
 
-describe('Nativo service', () => {
+describe.skip('Nativo service', () => {
```
preludek, line 8: Those tests are leaking and are going to get deleted anyway in https://github.com/Wikia/ad-engine/pull/1268

src/ad-engine/runner/runner.ts
dmnsobczak: Place should be fine as long as it is executed at AdEngine start 🤔

dmnsobczak (approved):

PR #1290, opened 2022-03-02
ADEN-11433 Introducing Prebid.js multi-level auctions

dmnsobczak:


platforms/shared/utils/instant-config-setup.ts
```diff
@@ -23,5 +25,14 @@ export class InstantConfigSetup implements DiProcess {
...
 		this.container.bind(InstantConfigCacheStorage).value(InstantConfigCacheStorage.make());
 		communicationService.dispatch(setInstantConfig({ instantConfig }));
+
+		this.preloadLibraries(instantConfig);
+	}
+
+	private preloadLibraries(instantConfig: InstantConfigService) {
+		if (instantConfig.get('icPrebid')) {
+			context.set('bidders.prebid.libraryUrl', instantConfig.get('icPrebidVersion'));
+			pbjsFactory.init();
```
nandy-andy, line 35: Don't we need `await` here in order to be sure `window.pbjs` is ready? We could also handle better the promise returned by `pbjsFactory.init()`.

platforms/ucp-desktop/ad-context.ts
```diff
@@ -21,10 +21,14 @@ export const basicContext = {
 		},
 		prebid: {
 			enabled: false,
-			lazyLoadingEnabled: false,
 			bidsRefreshing: {
 				slots: ['incontent_boxad_1'],
 			},
+			stagesSlots: {
```
nandy-andy, line 27: `r/stagesSlots/stagedSlots` maybe? :)

platforms/ucp-desktop/ad-context.ts
```diff
@@ -21,10 +21,14 @@ export const basicContext = {
...
+				init: ['top_leaderboard', 'featured'],
+				main: ['top_boxad', 'incontent_player'],
+				lazy: ['incontent_boxad_1', 'bottom_leaderboard'],
```
nandy-andy, line 30: What about the rest of GPT slots like `invisible_high_impact_2` and `interstitial`?

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -73,6 +73,7 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 				},
 			},
 			incontent_boxad_1: {
+				lazyCall: true,
```
nandy-andy, line 76: What do you think of adding a note in the name that it's lazy-loading connected to Prebid? On the other hand, at the end it's not only lazy calling Prebid but also lazy-calling GAM, isn't it? 🤔

src/ad-bidders/prebid/index.ts
```diff
@@ -224,6 +226,36 @@ export class PrebidProvider extends BidderProvider {
 		(pbjs.adUnits || []).forEach((adUnit) => pbjs.removeAdUnit(adUnit.code));
 	}
 
+	private registerStages(): void {
+		communicationService.on(eventsRepository.BIDDERS_INIT_STAGE_DONE, () => {
+			utils.logger(logGroup, 'multi auction init stage - done');
+
+			setTimeout(() => {
```
nandy-andy, line 233: Why is the `setTimeout()` necessary here? 🤔

src/ad-engine/ad-engine.ts
```diff
@@ -100,21 +100,37 @@ export class AdEngine {
...
+						() => {
+							this.pushSlot(ad, adSlot);
+						},
+					);
 				}
 			});
 		}
 	}
 
+	private pushSlot(ad: AdStackPayload, adSlot: AdSlot): void {
```
nandy-andy, line 120: You don't need `ad` if you've already have `adSlot`, do you? 

I was thinking if in the line below you can write: 
```ts
context.get(`slots.${adSlot.getSlotName()}.providers`)
```
instead of:
```ts
context.get(`slots.${ad.id}.providers`)
```

src/platforms-shared/bidders-state-setup.ts
```diff
@@ -14,7 +14,12 @@ export function setupBidders(context: Context, instantConfig: InstantConfigServi
 
 	if (instantConfig.get('icPrebid')) {
 		context.set('bidders.prebid.enabled', true);
-		context.set('bidders.prebid.libraryUrl', instantConfig.get('icPrebidVersion'));
+
+		const customPrebidConfig: object = instantConfig.get('icPrebidConfiguration');
```
nandy-andy, line 18: This is a bit scary without more validation ;) On the other hand: we own it and if we forget about a key in the config it will break loudly, won't it? :P

src/ad-bidders/prebid/index.ts
```diff
@@ -224,6 +226,36 @@ export class PrebidProvider extends BidderProvider {
 		(pbjs.adUnits || []).forEach((adUnit) => pbjs.removeAdUnit(adUnit.code));
 	}
 
+	private registerStages(): void {
```
nandy-andy, line 229: There are very similar lines in this method - maybe you can extract some logic to a reusable function 🤔

src/ad-bidders/prebid/index.ts
```diff
@@ -183,39 +179,45 @@ export class PrebidProvider extends BidderProvider {
...
+			).then(() => {
+				ats.call();
+			});
+		} else {
+			this.requestBids(this.adUnits, () => {
+				bidsBackHandler();
+				communicationService.emit(eventsRepository.BIDDERS_MAIN_STAGE_DONE);
+				communicationService.emit(eventsRepository.BIDDERS_LAZY_STAGE_DONE);
+			}).then(() => {
+				ats.call();
```
nandy-andy, line 212: Why did w have to move it and couldn't keep it in `requestBids()`?

src/ad-bidders/prebid/index.ts
```diff
@@ -183,39 +179,45 @@ export class PrebidProvider extends BidderProvider {
...
-			communicationService.on(
-				eventsRepository.BIDDERS_PREBID_LAZY_CALL,
+		this.applySettings();
+		this.removeAdUnits();
+
+		if (context.get('bidders.prebid.multiAuction')) {
+			utils.logger(logGroup, 'multi auction request enabled');
+
+			this.registerStages();
+			this.requestBids(
```
nandy-andy, line 196: You could extract it to `requestInitBids()` and what's below the line #207 could be extracted to `requestAllBids()` and hopefully it'll make the if-else more readable 🤞

platforms/shared/utils/instant-config-setup.ts
dmnsobczak: No. We don't need to wait for Prebid to load, because we don't need it at this point. We just want to start loading it as fast as possible - internal Promise will pause other chunks of code which use await there :)

platforms/ucp-desktop/ad-context.ts
dmnsobczak: Noo, it's a list of slots for every stage, so that sound ok. In ICBM we also have `stagesTimeouts` and `stagesIntervals`. I know it should be `...prebid.stages.slots` but this way it's easier to use a destruction syntax `{ ...config }`

platforms/ucp-desktop/ad-context.ts
dmnsobczak: Those slots are not using bidders, at least on UCP-Desktop. They can be requested normally with the second call.

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
dmnsobczak: Yes, this is a additional flag, but not the requirement. Lazy slots are loaded some time after second call start. For ICB1 it's 10 seconds, for BLB it's a long scroll. So indeed - they are GAM lazy :)

dmnsobczak: Also while adding this I found an issue that BLB on mobile is not lazy loaded. We want to change it and @sseyda is working on that.

src/ad-bidders/prebid/index.ts
```diff
@@ -183,39 +179,45 @@ export class PrebidProvider extends BidderProvider {
 
 	protected callBids(bidsBackHandler: (...args: any[]) => void): void {
```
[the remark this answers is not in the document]
dmnsobczak: Yes, it's hard to testi bidders because there is a lot to mock - whole Prebid, bids, queues. Probably that's why we created `wikia_adapter` and `wikia_video_adapter` and we strongly test it with synthetic tests 🤔

src/ad-bidders/prebid/index.ts
dmnsobczak: Isn't it too much? AdBidders is executing `requestBids()` which call `provider.call()`, then `this.callBids()`, then `requestInitBids()`, then `this.requestBids()` and finally `pbjs.requestBids()`? This chain is already very long :<

src/ad-bidders/prebid/index.ts
```diff
@@ -183,39 +179,45 @@ export class PrebidProvider extends BidderProvider {
...
-			);
+				context.get('bidders.prebid.stagesTimeouts.init') || this.timeout,
+			).then(() => {
+				ats.call();
+			});
+		} else {
+			this.requestBids(this.adUnits, () => {
+				bidsBackHandler();
+				communicationService.emit(eventsRepository.BIDDERS_MAIN_STAGE_DONE);
+				communicationService.emit(eventsRepository.BIDDERS_LAZY_STAGE_DONE);
```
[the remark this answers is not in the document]
dmnsobczak: Actually those are not needed there - thank you pointing this out :)

src/ad-bidders/prebid/index.ts
dmnsobczak: `ats.call()` was called multiple times when we were refreshing the ICB and that's a bug. Now it's called only once :) Also: small refactor applied.

src/ad-bidders/prebid/index.ts
dmnsobczak: Yeah, but almost every line is different - this might impact readability. But done: please review now :)

src/ad-bidders/prebid/index.ts
```diff
@@ -224,6 +226,36 @@ export class PrebidProvider extends BidderProvider {
 		(pbjs.adUnits || []).forEach((adUnit) => pbjs.removeAdUnit(adUnit.code));
 	}
 
+	private registerStages(): void {
+		communicationService.on(eventsRepository.BIDDERS_INIT_STAGE_DONE, () => {
```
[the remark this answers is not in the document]
dmnsobczak: After post-CR refactor it should be a bit easier to understand :)

src/ad-bidders/prebid/index.ts
dmnsobczak: I added a possibility to set additional interval before next auction to reduce network and CPU usage in users browsers. We might add something like 100ms here to see if timeout like that improves ad load times :)

src/ad-engine/ad-engine.ts
dmnsobczak: This won't be true if you use this context feature:
`this.config.slotName = this.config.slotName || ad.id;`
But since it's not used anywhere I can apply your suggestion :)

src/platforms-shared/bidders-state-setup.ts
dmnsobczak: Yep, this is an `adeng-config-editor` variable. It's fine for the experiment phase - we can add validation in follow ups if needed, but I predict that we're going to move it to context or even hardcode the winning values :D

platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
nandy-andy: That was by design AFAIK. BLB on desktop and mobile were lazy-loaded only with UAP 🤔 But if you guys want to change it'll make it unified with the UAP behaviour 👍

platforms/ucp-desktop/ad-context.ts
nandy-andy: If it's a list of every stage maybe it should be just called `stages`? ;) Or `adsLoadingStages`? :) My point is that if I read the code the name doesn't tell me anything.

platforms/ucp-desktop/ad-context.ts
nandy-andy: Weird, I thought we had a ticket to bring bidders to interstitial. Maybe it got blocked 🤷‍♂️

src/ad-bidders/prebid/index.ts
```diff
@@ -224,6 +228,22 @@ export class PrebidProvider extends BidderProvider {
 		(pbjs.adUnits || []).forEach((adUnit) => pbjs.removeAdUnit(adUnit.code));
 	}
 
+	private registerStage(name: MultiAuctionStep, trigger: MultiAuctionStep): void {
```
nandy-andy, line 231: We should pass here whole events' keys instead of partial strings. It'll make it easier in the future to find any occurrences of `BIDDERS_MAIN_STAGE_DONE` and `BIDDERS_LAZY_STAGE_DONE`.

dmnsobczak: Done!

platforms/ucp-desktop/ad-context.ts
dmnsobczak: Only on mobile :)

nandy-andy (approved):
Since it's an experiment let's QA it really hard 💪 And discuss what to monitor after the release. 

The thing that worries me the most is that it's not only lazy-calling bidder but also slows down calls to GAM (because it waits for the Prebid responses).

src/platforms-shared/bidders-state-setup.ts
preludek, line 18: We can do better typing here than just `object` 😛 . I think `InstantConfigGroup` will work

preludek (approved):
Are all of those changes somewhat tested (units/ synthetic)? If not, we should create a follow-up for tests assuming the experiment will show those changes are improving ad performance.

src/ad-bidders/prebid/index.ts
```diff
@@ -183,39 +179,47 @@ export class PrebidProvider extends BidderProvider {
...
 
-		if (this.isLazyLoadingEnabled) {
-			communicationService.on(
-				eventsRepository.BIDDERS_PREBID_LAZY_CALL,
+		this.applySettings();
+		this.removeAdUnits();
+
+		let firstBidRequest: Promise<void>;
+
+		if (context.get('bidders.prebid.multiAuction')) {
```
mfarul, line 194: we could use here strategy pattern or simply inheritance - instead of the big if here, you could create an instance based on the config and split those logic into two separate classes for better readability.

dmnsobczak: Once we confirm that this experiment is worth implementing we should definitely do some refactor :) For now I suggest leaving it as it is, because it's easier to revert or disable.

src/platforms-shared/bidders-state-setup.ts
dmnsobczak: Sad that context is not keeping typings 😭  I'll add those in the next ticket or to this in the meantime :)

PR #1292, opened 2022-03-03
ADEN-11595 - sequential messaging with size based creatives targeting 

Tomasz-at-Fandom:


platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
```diff
@@ -0,0 +1,25 @@
+import { TargetingManagerInterface } from '../domain/interfaces/targeting-manager.interface';
+import { ContextInterface } from '@wikia/ad-engine';
+import { SlotsContextInterface } from '../../slots/slots-context';
+
+export class GamTargetingManager implements TargetingManagerInterface {
+	private baseSize = 10;
+	private sizeMapping = {
+		'11x11': { originalSize: [970, 250] },
```
mfarul, line 8: what about other slots? we are only handling this for TLB on UAP here? If so, i would encode this information in the field name e.g. topLeaderboardSizeMapping ?

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
```diff
@@ -1,82 +1,47 @@
...
-
-@Injectable()
-export class SequentialMessagingSetup implements DiProcess {
-	constructor(private instantConfig: InstantConfigService) {}
+import { GamTargetingManager } from './infrastructure/gam-targeting-manager';
+import { slotsContext } from '../slots/slots-context';
+import { SequenceEndHandler } from './domain/sequence-end-handler';
+import { sequenceEventsTypes } from './infrastructure/sequence-event-types';
 
+export class SequentialMessagingSetup {
```
mfarul, line 11: Here i think we have overuse of classes - we have 3 different sequence handlers, each with very small responsibility that could also be put into a method, e.g. one handler with 3 methods. That should give more readability into how the algorithm works. Moreover we could create the handler once here instead of 3 different objects. Moving forward this class has the same methods, so it could implement the same interface as handler with 3 methods. So it could be async wrapper on the handler.

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
Tomasz-at-Fandom: First campaign is suppose to be TLB only so I limited the implementation to that slot only. Each slot requires separate testing, mapping for each size, template configuration and possibility of multiple parallel sequences comes in to play. That last part especially complicates matters, so not to drag this out I decided to keep the implementation limited for now. Let's run another campaign, make sure that single sequence in TLB works well and is something with enough profitability potential, before investing more time in to a more complex solution, because the complexity will grow exponentially beyond this point, as far as I can see.

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
Tomasz-at-Fandom: I agree, I wanted to implement a common interface for the 3 handler classes at some point, but when we switched from detecting the sequence base on ICBM config, to just being notified by the template a lot of code went out the window because this approach proved to be much simpler. That's how I ended up with 3 tiny handlers because i trimmed down from a more complicated implementation.
However, as I mentioned in the other comment, this is a minimum implementation for the first campaign. I see this getting more complicated as we will start adding more slots and will have to deal with parallel sequences in different slots. That is why I decided not to refactor the little classes in to methods because I expect this to grow again and it seems like a waste of time to polish this implementation so much knowing it's in a intermediate state.

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
Tomasz-at-Fandom: Oh and yea I'll change the name :)

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
```diff
@@ -0,0 +1,28 @@
...
+	private baseSize = 10;
+
+	constructor(private context: ContextInterface, private slotsContext: SlotsContextInterface) {}
+
+	setTargeting(sequenceId: string, userState: SequenceState): void {
+		const sizeSide = this.baseSize + userState.stepNo;
+		this.slotsContext.setSlotSize('top_leaderboard', [sizeSide, sizeSide]);
+		this.context.set(
+			'templates.sizeOverwritingMap',
+			this.generateSizeMapping(userState.width, userState.height),
```
mfarul, line 16: Maybe instead of giving a map for 3 additional steps just a map for this particular step? It will handle then all possible sequence lenght.

platforms/shared/sequential-messaging/infrastructure/sequence-event-types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface SequenceEventOptions {
+	category: string;
+	name: string;
+	payload: any;
+}
+
+export const sequenceEventsTypes: Dictionary<SequenceEventOptions> = {
+	// This event is fired in a GAM template
+	GAM_SEQUENTIAL_MESSAGING_STARTED: {
```
mfarul, line 12: Let's not call it GAM_ event. Even if now GAM template is firing it might not be the only source in the future. 
This should be like a domain type event with a source in a particular implementation, being GAM here.

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
```diff
@@ -0,0 +1,28 @@
...
+		);
+		this.context.set('targeting.uap', sequenceId);
+	}
+
+	private generateSizeMapping(width: number, height: number) {
+		return {
+			'12x12': { originalSize: [width, height] },
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: From 1, I think it will be less confusing that way.12 will be step 2, 13 will be step 3 and so on.

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
Tomasz-at-Fandom: OK will do. I tried it earlier but couldn't find the syntax for it in JS. Now I finally have it's called `computed keys` from ES2015 :)

platforms/shared/sequential-messaging/infrastructure/sequence-event-types.ts
Tomasz-at-Fandom: OK will do!

platforms/shared/sequential-messaging/domain/sequence-continuation-handler.ts
```diff
@@ -1,35 +1,25 @@
...
-		for (const sequentialAdId of Object.keys(userState)) {
-			const adConfig = fullConfig[sequentialAdId];
-
-			this.targetingManager.setTargeting(adConfig.targeting);
-
-			this.onSlotShowedEvent(
-				new SequenceEndStateHandler(this.userStateStore, adConfig.lastStepId.toString()),
-			);
+		for (const sequenceId of Object.keys(userState)) {
+			userState[sequenceId].stepNo++;
```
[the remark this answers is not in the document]
Tomasz-at-Fandom: I don't understand the problem, let's talk about this tomorrow please.

mfarul (approved):

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
Tomasz-at-Fandom: Will do in another ticket as agreed on Zoom

PR #1293, opened 2022-03-04
ADEN-11434-2 Adding impression tracking to LIS

dmnsobczak:


preludek (approved):

src/ad-engine/ad-engine.ts
```diff
@@ -37,6 +37,7 @@ type LayoutPayload = {
 
 interface FanTakeoverLayoutPayload extends LayoutPayload {
 	data: {
+		impression: string;
```
nandy-andy, line 40: `impressionPixelUrl` is more descriptive 🤔

PR #1299, opened 2022-03-09
ADEN-11498 | Add click tracking to Celtra interstitial

bckatarzyna:


src/ad-tracking/click-tracker.ts
```diff
@@ -2,6 +2,7 @@ import { communicationService, eventsRepository } from '@ad-engine/communication
 import { AdSlot, FuncPipeline, FuncPipelineStep, slotService, utils } from '@ad-engine/core';
 
 const logGroup = 'ad-click-tracker';
+const celtraInterstitialBannerId = 'celtra-object-37';
```
nandy-andy, line 5: I'm worried it'll work for the test "Best Buy" interstitial but for new ones it won't work. Have we checked couple of Celtra interstitial and verified it's always `celtra-object-37`? 😊

src/ad-tracking/click-tracker.ts
```diff
@@ -33,6 +34,10 @@ class AdClickTracker {
 			false,
 		);
 
+		communicationService.on(eventsRepository.GAM_INTERSTITIAL_LOADED, () => {
+			this.addCeltraInterstitialClickTrackingListener(callback);
```
nandy-andy, line 38: Great idea to minimise the impact and only add the listener for pageviews with interstitials. I hope it's just a % of our traffic 😊

src/ad-tracking/click-tracker.ts
```diff
@@ -87,6 +92,29 @@ class AdClickTracker {
...
+				const celtraBanner = celtraIframe.contentWindow.document.getElementById(
+					celtraInterstitialBannerId,
+				);
+				if (celtraBanner) {
+					celtraBanner.addEventListener('click', (event) => {
+						this.handleClickEvent(callback, slotService.get('top_leaderboard'), event);
+					});
+					observer.disconnect();
+				}
+			}
```
nandy-andy, line 110: Consider adding some `utils.logger()` calls in the `else` cases just-in-case you'd like to debug it on production 😈 in order to check for example why we have so little data or something.

PR #1302, opened 2022-03-11
ADEN-11544 | Add Live Connect service

bckatarzyna:


src/ad-services/live-connect/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	call(): void {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+			return;
+		}
+
+		if (!this.isLoaded) {
+			utils.logger(logGroup, 'loading');
+			context.set('targeting.AU_SEG', '-1');
```
nandy-andy, line 27: Looks like a left-over after the code you were inspired by ;) Unless, LiveConnect has anything to do with Audigent ;)

src/ad-services/live-connect/index.ts
```diff
@@ -0,0 +1,40 @@
+import { context } from '@ad-engine/core';
+import * as utils from '../../ad-engine/utils';
```
nandy-andy, line 2: What do you think of importing `utils` in the line above from `@ad-engine/core`? 🤔

src/ad-services/live-connect/index.ts
bckatarzyna: Hahaha, thanks!

PR #1303, opened 2022-03-11
ADEN-11654 Remove old StickyTLB template and adjustments for mobile

dmnsobczak:


platforms/f2/styles.scss
```diff
@@ -165,6 +165,10 @@
 
 					#top_leaderboard {
 						height: unset;
+
+						html.mobile & {
+							z-index: 5000102;
```
nandy-andy, line 170: Can't we use some SCSS variable? 😊 We can mention it in a follow-up ticket ;)

nandy-andy (approved):

platforms/f2/styles.scss
dmnsobczak: Actually I think that News&Stories navbar shouldn't have such high z-index, but it's too risky to change it now. Let's follow-up it :D

PR #1310, opened 2022-03-22
ADEN-11314 AdEngine stops working when the page is loaded in an iframe

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11314


nandy-andy (approved):

platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -26,7 +26,13 @@ class TrackingOptInWrapper {
 			communicationService.on(eventsRepository.AD_ENGINE_CONSENT_READY, (payload) => {
```
szymon-nawrocki, line 26: The communicationService does not work properly when site is run inside iframe, so the `AD_ENGINE_CONSENT_READY` most of the time is not send from the backend, so I've added the lines below to go around this obstacle.
I know that it is far from perfect solution, but we need it only for a week or two to collect a data of how often AdEngine is being loaded in iframe, and I couldn't think of a better idea of how to do it.

platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -26,7 +26,13 @@ class TrackingOptInWrapper {
...
 				this.setConsents(payload);
 				resolve();
+				return;
 			});
+
+			const pageInIframe = document.documentElement.classList.contains('in-iframe');
```
szymon-nawrocki, line 32: So I've added an additional promise resolve possibility, which is:
if `AD_ENGINE_CONSENT_READY` event is not sent from the backend -> check if site is loaded in iframe -> if it is -> resolve the promise

@nandy-andy
Do you think this solution could fly for a week or two, to collect some data of how often AdEngine is being loaded in iframe? Then we could probably remove it.

szymon-nawrocki:
Reopening PR due to closing by mistake

platforms/shared/consent/tracking-opt-in-wrapper.ts
nandy-andy, line 32: how about `r/pageInIframe/isPageInIframe/`? :)

platforms/shared/consent/tracking-opt-in-wrapper.ts
```diff
@@ -26,7 +26,13 @@ class TrackingOptInWrapper {
...
+			if (pageInIframe) {
+				resolve();
```
nandy-andy, line 34: Don't we need to use `return` here as well as we do in line no. 29? Or maybe we don't need it in line no. 29 either? Or maybe we need to return `resolve();`? 🤔

platforms/shared/setup/iframe-detector.setup.ts
```diff
@@ -0,0 +1,18 @@
+import { DiProcess } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { PageTracker } from '../tracking/page-tracker';
+import { DataWarehouseTracker } from '../tracking/data-warehouse';
+
+@Injectable()
+export class IframeDetectorSetup implements DiProcess {
+	execute(): void {
+		const elementFound = document.documentElement.classList.contains('in-iframe');
```
nandy-andy, line 9: element or class found? ;-)

PR #1311, opened 2022-03-22
ADEN-11498 Handle situation with many Celtra iframes

nandy-andy:


nandy-andy:
Follow-up: https://github.com/Wikia/ad-engine/pull/1312

PR #1317, opened 2022-03-25
ADEN-11651 Sequential messaging v0.3 alpha  

Tomasz-at-Fandom:


platforms/shared/sequential-messaging/domain/sequence-continuation-handler.ts
```diff
@@ -19,7 +20,15 @@ export class SequenceContinuationHandler {
 			userState[sequenceId].stepNo++;
 			const sequenceState = userState[sequenceId];
 			this.targetingManager.setTargeting(sequenceId, sequenceState);
-			this.userStateStore.set(userState);
+			this.onIntermediateStepLoad((loadedStep: number) => {
+				// TODO SM attempts counter can be added for extra safety
+				if (loadedStep !== userState[sequenceId].stepNo) {
+					console.log('[SM] Invalid step loaded by the Provider!');
```
szymon-nawrocki, line 26: I think using logger (`utils/logger`) to log a message would be more in line with AdEngine project convention

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
```diff
@@ -24,26 +38,39 @@ export class SequentialMessagingSetup {
...
 			sequenceHandler.endSequence();
 		});
 	}
+
+	private onIntermediateStepLoad(storeState: (loadedStep: number) => void) {
+		communicationService.on(SequenceEventTypes.SEQUENTIAL_MESSAGING_INTERMEDIATE, (payload) => {
+			// TODO SM extract the 12 and 14 number to a shared parameters
+			//  to be used here and in GamTargetingManager.generateSizeMapping
+			if (payload.height == null || 12 > payload.height || payload.height > 14) {
+				console.log('[SM] Invalid Creative configuration. Creative size ot ouf bounds.');
```
szymon-nawrocki, line 68: Same here with using logger to log a message

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
```diff
@@ -24,26 +38,39 @@ export class SequentialMessagingSetup {
...
+			const sequenceHandler = new SequenceEndHandler(this.userStateStore);
 			sequenceHandler.endSequence();
 		});
 	}
+
+	private onIntermediateStepLoad(storeState: (loadedStep: number) => void) {
+		communicationService.on(SequenceEventTypes.SEQUENTIAL_MESSAGING_INTERMEDIATE, (payload) => {
+			// TODO SM extract the 12 and 14 number to a shared parameters
+			//  to be used here and in GamTargetingManager.generateSizeMapping
+			if (payload.height == null || 12 > payload.height || payload.height > 14) {
```
szymon-nawrocki, line 67: In Javascript I'd recomment using identity operator '===' instead of equality operator '==' to avoid unexpected situations/behaviour.

But in this case it would be even better to use `if (!payload.height)` which is easier to read

spec/platforms/shared/sequential-messaging/domain/sequence-continuation-handler.spec.ts
```diff
@@ -37,4 +50,54 @@ describe('Sequence Continuation Handler', () => {
 		assert.calledOnce(userStateStoreSpy.set);
 		assert.calledWith(userStateStoreSpy.set, userInitialStateAfterSecondStep);
 	});
+
+	it('Handle invalid step loaded', () => {
+		const userStateStoreSpy = makeUserStateStoreSpy();
```
szymon-nawrocki, line 55: Since `userStateStoreSpy` and `targetingManagerSpy` are created in every test, won't it be better to create them in `beforeEach` function?

platforms/shared/sequential-messaging/domain/sequence-continuation-handler.ts
Tomasz-at-Fandom: It's a good idea, we can change this later on, if we deem this message necessary. At the moment it's more of a tests helper, users should never see this message if everything is properly configured. I also don't want to add the logger directly as it would pull in a chain of dependencies in to the "domain layer" of this feature.

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
Tomasz-at-Fandom: OK, I can do that here :)

platforms/shared/sequential-messaging/sequential-messaging.setup.ts
Tomasz-at-Fandom: I'm using double equal sign here because I expect that payload.height might be also undefined.
But the suggested solution covers both cases to I can go with !payload.height as well :)

spec/platforms/shared/sequential-messaging/domain/sequence-continuation-handler.spec.ts
Tomasz-at-Fandom: Yea I can clean it up a little 👍

szymon-nawrocki (approved):

PR #1328, opened 2022-03-29
ADEN-11680 feat: SM UAP state consistency across communities

Tomasz-at-Fandom:


nandy-andy (approved):

platforms/shared/sequential-messaging/infrastructure/gam-targeting-manager.ts
```diff
@@ -18,6 +19,14 @@ export class GamTargetingManager implements TargetingManagerInterface {
...
 		);
 		this.context.set('slots.top_leaderboard.targeting.sequential', sequenceId);
+
+		if (this.isUap(userState)) {
+			this.forceUapResolveState();
+		}
+	}
+
+	private isUap(userState: SequenceState) {
+		return userState.width === 2 || userState.width === 3;
```
nandy-andy, line 29: Just a thought to later assign `2` and `3` to a more descriptive variables. I assume it's like this:
```js
const UAP_MOBILE_WIDTH = 2;
const UAP_DESKTOP_WIDTH = 3;
```

Tomasz-at-Fandom: Good idea, I'll add this.

PR #1329, opened 2022-03-29
ADEN-11697 Load AdEngine in NoAdsMode when page is loaded inside iframe

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11697

### PR:
- UCP: https://github.com/Wikia/unified-platform/pull/9387


spec/platforms/shared/services/no-ads-detector.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { NoAdsDetector } from '../../../../platforms/shared';
+import { expect } from 'chai';
+
+describe('No Ads Detector', () => {
+	let noAdsDetector: NoAdsDetector;
+	beforeEach(() => {
+		noAdsDetector = new NoAdsDetector();
+	});
+
+	it('isAdsMode() return true when reasons array is empty', () => {
```
bckatarzyna, line 10:
```suggestion
	it('isAdsMode() returns true when reasons array is empty', () => {
```

spec/platforms/shared/services/no-ads-detector.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+		expect(noAdsDetector.isAdsMode()).to.equal(true);
+	});
+
+	it('isAdsMode() return false when reasons array is not empty', () => {
```
bckatarzyna, line 14:
```suggestion
	it('isAdsMode() returns false when reasons array is not empty', () => {
```

bckatarzyna (approved):

platforms/shared/services/no-ads-detector.ts
```diff
@@ -17,7 +17,9 @@ export class NoAdsDetector {
 	}
 
 	addReasons(reasons: string[]): void {
-		this.reasons.push(...reasons);
+		if (reasons) {
```
nandy-andy, line 20: What does cause `reasons` not to be an array?

szymon-nawrocki: This line of code:
https://github.com/Wikia/ad-engine/blob/90d8dfc5a1398eae6165696cebb019deaa8418d5/platforms/shared/modes/no-ads/no-ads.mode.ts#L12
was breaking an AdEngine, because `window.ads.context.opts.noAdsReasons` was undefined, so I've added this special check

szymon-nawrocki: I've had another idea, to define `window.ads.context.opts.noAdsReasons` here https://github.com/Wikia/ad-engine/blob/dc2ccd07c53d41f5a00c068c255978b25ad99dd5/platforms/shared/setup/base-context.setup.ts#L35
with: `window.ads.context.opts.noAdsReasons = ['in-iframe']`, without changing `addReasons` function.

But I decided to go with a first option

nandy-andy: Hm, I think the bug is on the backend that we set it to `undefined` instead of an empty array when there are no reasons on the backend for no ads mode 🤔 https://github.com/Wikia/unified-platform/blob/master/extensions/fandom/AdEngine/AdEngine.class.php#L152

szymon-nawrocki: Yes it may not be that easy, so I've added a warning when `reasons=undefined` to not miss any unwanted behaviours of NoAdsDetector

PR #1332, opened 2022-03-31
ADEN-11693 Extend targeting based on a cookie

nandy-andy:


platforms/shared/utils/get-cross-domain-targeting.ts
```diff
@@ -0,0 +1,7 @@
+import { CookieStorageAdapter } from '@wikia/ad-engine';
+
+export function getCrossDomainTargeting(crossDomainStorage: CookieStorageAdapter) {
+	return {
+		'cortex-visitor': crossDomainStorage.getItem('adeng-cortex-rpg-visited') ? 'yes' : 'no',
```
mfarul, line 5: I don't like the adeng- context here. Either it's Fandom or just plain cortex-rpg-visited :)

jakubbanas (approved):

jakubbanas (approved):

PR #1333, opened 2022-04-04
ADEN-11701 fix: put UAP sequence on hold when FV is loaded

Tomasz-at-Fandom:


platforms/shared/sequential-messaging/domain/data-structures/user-sequential-message-state.ts
```diff
@@ -1,7 +1,12 @@
...
-	width: number;
-	height: number;
+export class SequenceState {
+	readonly UAP_MOBILE_WIDTH: number = 2;
+	readonly UAP_DESKTOP_WIDTH: number = 3;
+
+	constructor(public stepNo: number, public width: number, public height: number) {}
+
+	isUap(): boolean {
+		return this.width === this.UAP_MOBILE_WIDTH || this.width === this.UAP_DESKTOP_WIDTH;
```
mfarul, line 8: Hell of a hack here :) Is there any other way to get this info? Setting an explicit JS/DOM variable on the GAM creative level maybe?

Tomasz-at-Fandom: Yea I guess we can define a flag in the Start Creative and send it over to AdEngine in the sequence Start Event and store it in the cookie. One more little thing to configure when creating the LI but it will be explicit in code.

platforms/shared/sequential-messaging/domain/data-structures/user-sequential-message-state.ts
```diff
@@ -1,7 +1,14 @@
...
-	height: number;
+export class SequenceState {
+	constructor(
+		public stepNo: number,
+		public width: number,
+		public height: number,
+		public uap: boolean = false,
+	) {}
+
+	isUap(): boolean {
```
szymon-nawrocki, line 9: You can check if UAP is loaded by checking the event `eventsRepository.AD_ENGINE_UAP_LOAD_STATUS` with `communicationService`.

Example:
https://github.com/Wikia/ad-engine/blob/27f3fec442ca17c49626000c2738623c9d45dfa6/platforms/shared/utils/placeholder-service.ts#L75

Tomasz-at-Fandom: On sequence continuation I need to know that prior to this event being fired, hence this method.

nandy-andy: This way it's possibly even faster ;) Just right with the GAM response https://github.com/Wikia/ad-engine/pull/1333/files#diff-c71f555df036889edcf57f34ebec5cead18f6dfe99ea201fc014914e2703dddfR37

Tomasz-at-Fandom: On sequence continuation I need to know if SM is UAP before the request is sent :) But I just realise I could utilize the event that @szymon-nawrocki mentioned in the sequence detection step. This method however needs to stay.

jakubbanas (approved):

PR #1334, opened 2022-04-04
ADEN-11532 Detailed UAP tracking in DW

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11532


src/ad-tracking/slot-tracking-middleware.ts
```diff
@@ -1,5 +1,6 @@
 import { context, FuncPipelineStep, InstantConfigCacheStorage, utils } from '@ad-engine/core';
 import { AdInfoContext } from './slot-tracker';
+import { universalAdPackage } from '../ad-products';
```
dmnsobczak, line 3: We shouldn't import `universalAdPackage` which is an `ad-products` module inside another `src` submodule :/ And going outside of the module with `../` is bad too, because it breaks tree shaking features and might lead to the circular dependencies. Do you have other idea how to pass UAP type here? It can be set on context and then used here.

platforms/shared/templates/handlers/roadblock/roadblock-handler.ts
```diff
@@ -29,7 +29,7 @@ export class RoadblockHandler implements TemplateStateHandler {
 	) {}
 
 	async onEnter(): Promise<void> {
-		this.params.adProduct = 'ruap';
+		this.params.adProduct = this.params.adProduct || 'ruap';
```
dmnsobczak, line 32: `adProduct: string;`
This field has a general `string` type. I think we can list all possible values in TS mappings :D

src/ad-tracking/slot-tracking-middleware.ts
szymon-nawrocki: Ahh you're right. Thanks for pointing this out.

szymon-nawrocki: I've added UAP type to context ✅

platforms/shared/templates/handlers/roadblock/roadblock-handler.ts
szymon-nawrocki: done ✅

src/ad-tracking/slot-tracking-middleware.ts
```diff
@@ -1,5 +1,6 @@
 import { context, FuncPipelineStep, InstantConfigCacheStorage, utils } from '@ad-engine/core';
 import { AdInfoContext } from './slot-tracker';
+import { AdProductType } from '../ad-products';
```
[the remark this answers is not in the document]
szymon-nawrocki: You're right. Sorry for making the same again :)
I've moved AdProductType to global @types ✅

dmnsobczak (approved):

szymon-nawrocki:
Timeout

PR #1336, opened 2022-04-04
ADEN-11556 load PIH asynchronously

jakubbanas:


spec/ad-bidders/prebid/identity-hub/identity-hub.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+describe('Pubmatic IdentityHub', () => {
+	const sandbox = createSandbox();
+	let loadScriptSpy;
+
+	beforeEach(() => {
+		loadScriptSpy = sandbox.stub(utils.scriptLoader, 'loadScript');
+		loadScriptSpy.resolvesThis();
+
+		context.set('pubmatic.identityHub.enabled', true);
```
szymon-nawrocki, line 13: I wonder if it won't be better to mock a context, and do the operations on the mock to avoid potential issues in other tests caused by changing the global object. @jakubbanas  what do you think?

Example:
https://github.com/Wikia/ad-engine/blob/de7e0e3727b4007d9a1a98615125f30babf5eed2/spec/ad-products/templates/uap/universal-ad-package.spec.ts#L16

spec/ad-bidders/prebid/identity-hub/identity-hub.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	it('IdentityHub is disabled by feature flag', async () => {
+		context.set('pubmatic.identityHub.enabled', false);
+
+		await identityHub.call();
+
+		assert.notCalled(loadScriptSpy);
+	});
+
+	it('IdentityHub is disabled if user has opted out sale', async () => {
```
szymon-nawrocki, line 37: Doesn't `options.optOutSale` mean that we (AdEngine) opted out of sale, not the user?

spec/ad-bidders/prebid/identity-hub/identity-hub.spec.ts
jakubbanas: Good point, I'll change it 👍

spec/ad-bidders/prebid/identity-hub/identity-hub.spec.ts
szymon-nawrocki: Okay I see. Thanks for the explanation

szymon-nawrocki (approved):
👌

PR #1337, opened 2022-04-05
ADEN-11776 Move porvata second playback VAST call

dmnsobczak:


src/ad-products/video/porvata/porvata-player.ts
```diff
@@ -158,7 +157,27 @@ export class PorvataPlayer {
...
+				window.google.ima.AdsManagerLoadedEvent.Type.ADS_MANAGER_LOADED,
+				runCallback,
+			);
+			this.requestAds();
+		} else {
+			this.run();
+		}
+	}
+
+	run(): void {
```
szymon-nawrocki, line 180: @dmnsobczak What do you think about giving a little more descriptive name to this function?
Something like "initializeVideoAd" for example?

dmnsobczak: Good idea, changed!

szymon-nawrocki (dismissed):

PR #1339, opened 2022-04-05
ADEN-11715 added package size report github actions

preludek:


.github/workflows/package-size-report.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
szymon-nawrocki, line 19: It would be nice to setup node with version taken from `.nvmrc` file:
https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

preludek: Done

szymon-nawrocki (approved):
👌

PR #1340, opened 2022-04-06
ADEN-11676 Disable ad slots for engagement test

jakubbanas:


platforms/shared/setup/base-context.setup.ts
```diff
@@ -185,6 +185,7 @@ export class BaseContextSetup implements DiProcess {
 
 		context.set('options.initCall', this.instantConfig.get('icLayoutInitializerSlot'));
 		context.set('pubmatic.identityHub.enabled', this.instantConfig.get('icPubmaticIdentityHub'));
+		context.set('icDisabledSlots', this.instantConfig.get('icDisabledSlots'));
```
nandy-andy, line 188: Do we still need the variable?

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -29,16 +22,6 @@ export class F2DynamicSlotsSetup implements DiProcess {
 			this.slotsDefinitionRepository.getIncontentBoxadConfig(),
 			this.slotsDefinitionRepository.getBottomLeaderboardConfig(),
 		]);
-
-		if (!topLeaderboardDefinition) {
```
nandy-andy, line 33: Isn't removing this piece of code breaking something on f2? 🤔

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,78 @@
...
+		this.disableUnit(unitName);
+	}
+
+	disableUnit(unitName: string) {
+		switch (unitName) {
+			case 'uap':
+				blockUAP();
+				return;
+			case 'interstitial_celtra':
+				this.cookieAdapter.setItem('_ae_intrsttl_imp', '1');
```
nandy-andy, line 61: Once a user gets to this group and we set the cookie she'll become lucky since next pageviews for her will be without the Celtra interstitial too, won't they? ;) However, I don't have any good idea of "fixing" it :/

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,78 @@
...
+export function blockUAP(): void {
+	const { desktop, mobile } = universalAdPackage.UAP_ADDITIONAL_SIZES.bfaSize;
+	slotsContext.removeSlotSize('top_leaderboard', desktop);
+	slotsContext.removeSlotSize('top_leaderboard', mobile);
+}
+
+export function getUnitNameToDisable(configs: NoAdsConfig[], userBeacon): string {
+	const config = configs.find((conf) => userBeacon.match(conf.beaconRegex));
+
+	return config?.unitName;
```
nandy-andy, line 38: What do you think of removing `?`, so it'll break loudly and we'll catch cases when there will be a bug in the config.

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,78 @@
...
+	constructor(
+		private instantConfig: InstantConfigService,
+		private cookieAdapter: CookieStorageAdapter,
+	) {}
+
+	execute(): void {
+		const configs = this.instantConfig.get<object>('icNoAdsExperimentConfig', []) as NoAdsConfig[];
+		const userBeacon: string = this.cookieAdapter.getItem('wikia_beacon_id');
+		const unitName = getUnitNameToDisable(configs, userBeacon);
+		this.disableUnit(unitName);
```
nandy-andy, line 52: If we make sure `getUnitNameToDisable()` always returns a string with the unit we could go with a one-liner here:
```js
this.disableUnit(getUnitNameToDisable(configs, userBeacon));
```

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
jakubbanas: Right, it was supposed to stay here! 🙈

platforms/shared/setup/base-context.setup.ts
jakubbanas: Right, it's the next thing I forgot to remove from the POC branch.

platforms/shared/setup/noads-experiment.setup.ts
jakubbanas: I believe it would not be a problem. The idea behind this "beacon targeting" is that the experiment lives as long as the beacon itself. Beacon stays the same between page views, so the only way to jump off the experiment is to clear cookies, which will also remove the  Celtra blocker.

platforms/shared/setup/noads-experiment.setup.ts
jakubbanas: The `find` method returns `undefined` when it finds nothing, which means the beacon doesn't match any experiment. This is not a bug in the config, it might happen that the beacon should not participate in any experiment.
I'll add a test for that case.

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function blockUAP(): void {
+	const { desktop, mobile } = universalAdPackage.UAP_ADDITIONAL_SIZES.bfaSize;
+	slotsContext.removeSlotSize('top_leaderboard', desktop);
+	slotsContext.removeSlotSize('top_leaderboard', mobile);
+}
+
+export function getUnitNameToDisable(configs: NoAdsConfig[], userBeacon): string {
+	const config = configs.find((conf) => userBeacon.match(conf.beaconRegex));
```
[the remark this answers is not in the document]
jakubbanas: I added a default value for the beacon, so when it's undefined it's falling back to an empty string.

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function getUnitNameToDisable(configs: NoAdsConfig[], userBeacon = ''): string {
+	const config = configs.find((conf) => userBeacon.match(conf.beaconRegex));
+
+	return config?.unitName;
+}
+
+@Injectable()
+export class NoAdsExperimentSetup implements DiProcess {
```
mfarul, line 42: I think this could go beyond experiment - if this could be a UnitDisabler with a provided function getUnitNameToDisable we could use it in an experiment environment and in a regular ICBM environment when needed. This is not needed now, just for future reference.

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -0,0 +1,80 @@
...
+import { slotsContext } from '../slots/slots-context';
+
+interface NoAdsConfig {
+	unitName: string;
+	beaconRegex: string;
+}
+
+export function skipBtfBlocker() {
+	communicationService.on(eventsRepository.AD_ENGINE_STACK_START, () => {
+		btfBlockerService.finishFirstCall();
```
dmnsobczak, line 21: This is a duplicate of this:
https://github.com/Wikia/ad-engine/blob/dev/platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts#L34
I think that logic of manually enabling the second call can be refactored, for example somewhere in the `shared/ads-mode`. You can run a code on `AD_ENGINE_STACK_START` event, check in slots context or awaiting ad stack if there is any enabled slot with `firstCall` flag in context and if not, then jumping straight to the second call 🤔

src/ad-engine/ad-engine.ts
```diff
@@ -95,6 +101,10 @@ export class AdEngine {
 
 	private pushSlot(adSlot: AdSlot): void {
 		const providersChain = context.get(`slots.${adSlot.getSlotName()}.providers`) || [];
+		if (context.get(`slots.${adSlot.getSlotName()}.disabled`)) {
```
dmnsobczak, line 104: This is risky. Some slots are disabled in the ad context and they are enabled once something happen, for example `mobile_prefooter` on mobile is enabled by UAP filling the TLB slot. If you leave this then `mobile_prefooter` won't be pushed to the ads queue and then second call start it won't be requested. Did you checked it?
Also: this looks like a duplicate, because disabled slots won't be filled there:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/btf-blocker-service.ts#L115
What's the idea behind it?

dmnsobczak (changes requested):
New `slot.disabled` logic will probably break some dynamically enabled slots. Comments added.

src/ad-engine/ad-engine.ts
nandy-andy: Blocking slots for different providers: GPT and Nativo ;)

dmnsobczak: Ack. Then it can't be done like this, because this will introduce regression in oustream, mobile_prefooter, ICB and BLB on F2 :(

platforms/shared/setup/noads-experiment.setup.ts
dmnsobczak: 👍

src/ad-engine/ad-engine.ts
dmnsobczak: Added extra check for Nativo.

platforms/shared/setup/noads-experiment.setup.ts
dmnsobczak: I'll keep it as it is since this is an experiment, so this code will be removed in the future and manually enabling second call is only used on News&Stories 🤷‍♂️

src/ad-engine/ad-engine.ts
nandy-andy: As for experiment we can go with that but it should be removed after the experiments are done and we should redesign it in near future (LIS and performance efforts 🤞)

nandy-andy (approved):
Only for the experiment 😊

nandy-andy (approved):

PR #1342, opened 2022-04-07
ADEN-11717 Prebid multi auction - second iteration

dmnsobczak:


src/ad-bidders/prebid/index.ts
```diff
@@ -197,28 +195,36 @@ export class PrebidProvider extends BidderProvider {
...
 						communicationService.emit(eventsRepository.BIDDERS_INIT_STAGE_DONE);
 						communicationService.emit(eventsRepository.BIDDERS_MAIN_STAGE_DONE);
-						communicationService.emit(eventsRepository.BIDDERS_LAZY_STAGE_DONE);
 					}
 				},
 			);
 
 			firstBidRequest = this.requestBids(
-				this.filterAdUnits(this.adUnits, 'init'),
+				this.filterAdUnits(this.adUnits, true),
```
bckatarzyna, line 209: The boolean argument here is not self-descriptive and makes the code harder to understand

bckatarzyna (approved):

src/ad-bidders/prebid/index.ts
dmnsobczak: Fixed!

PR #1345, opened 2022-04-08
ADEN-11724 Bump dependencies

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11724


package.json
```diff
@@ -34,29 +34,29 @@
 		"@types/node": "8.10.54",
 		"@types/sinon": "7.0.7",
 		"@typescript-eslint/eslint-plugin": "5.9.1",
-		"@typescript-eslint/parser": "5.9.1",
+		"@typescript-eslint/parser": "5.18.0",
```
dmnsobczak, line 37: This is just an update and it's safe to bump 👍

package.json
```diff
@@ -34,29 +34,29 @@
...
 		"awesome-typescript-loader": "5.2.1",
 		"babel-loader": "8.2.3",
 		"chai": "^4.1.2",
-		"css-loader": "6.6.0",
+		"css-loader": "6.7.1",
```
dmnsobczak, line 41: `* re-export defaultGetLocalIdent` + update of deps 👍

package.json
```diff
@@ -34,29 +34,29 @@
...
 		"express": "4.17.2",
-		"husky": "1.1.4",
+		"husky": "7.0.4",
 		"instant-mocha": "1.4.1",
 		"istanbul-instrumenter-loader": "3.0.1",
 		"jsdom": "11.11.0",
 		"jsdom-global": "^3.0.2",
 		"jwplayer-fandom": "github:wikia/jwplayer-fandom#1.7.0",
-		"lint-staged": "12.1.4",
+		"lint-staged": "12.3.7",
```
dmnsobczak, line 50: Nothing major except: `make console task titles more explicit` 👍

package.json
```diff
@@ -34,29 +34,29 @@
...
 		"mini-css-extract-plugin": "2.5.3",
 		"mocha": "9.2.1",
-		"mocha-jsdom": "^1.1.0",
+		"mocha-jsdom": "^2.0.0",
```
dmnsobczak, line 53: https://github.com/rstacruz/mocha-jsdom#upgrading-to-v200
> If you are coming from mocha-jsdom v1.x, remove jsdom if you're not using it before upgrading. jsdom is now a direct dependency of mocha-jsdom.

We are not using `jsdom`, tests are using `jsdom-global`. It should be removed after upgrade.

package.json
```diff
@@ -34,29 +34,29 @@
...
 		"mocha-junit-reporter": "2.0.2",
 		"node-sass": "7.0.1",
 		"nyc": "15.1.0",
-		"prettier": "1.19.1",
+		"prettier": "2.6.2",
```
dmnsobczak, line 57: https://prettier.io/blog/2020/03/21/2.0.0.html
There are some API changes. Did you tried to run Prettier on the whole project to see if it change anything anywhere?

package.json
```diff
@@ -34,29 +34,29 @@
...
 		"process": "0.11.10",
-		"sass-loader": "^6.0.6",
+		"sass-loader": "^12.6.0",
```
dmnsobczak, line 59: This is a big upgrade. Did you checked release logs of every major version?
I see that minified built files changed. QA should include manual inspection of our templates.

dmnsobczak: Release log:
https://github.com/webpack-contrib/sass-loader/releases/tag/v12.0.0

dmnsobczak: > minimum supported webpack version is 5

What about `"webpack-cli": "4.9.2",`?

package.json
```diff
@@ -34,29 +34,29 @@
...
+		"@typescript-eslint/parser": "5.18.0",
 		"awesome-typescript-loader": "5.2.1",
 		"babel-loader": "8.2.3",
 		"chai": "^4.1.2",
-		"css-loader": "6.6.0",
+		"css-loader": "6.7.1",
 		"eslint": "8.6.0",
 		"express": "4.17.2",
-		"husky": "1.1.4",
+		"husky": "7.0.4",
```
dmnsobczak, line 44: Upgrading Husky is a pretty complicated process including rewriting the hooks:
https://github.com/typicode/husky-4-to-7
I'm pretty sure that Prettier and Linter won't work, can you check it with some commit?

dmnsobczak (changes requested):
* Husky requires more work
* `jsdom` should be removed
* Let's test package generation after sass upgrade
* Please add better QA notes for manual testing

package.json
szymon-nawrocki: Thanks for pointing that out, I removed  `jsdom` from dependencies

package.json
szymon-nawrocki: You're right. I will create a follow up ticket for that

package.json
szymon-nawrocki: Yes, indeed that is a big upgrade. I will create a follow up ticket for that

package.json
szymon-nawrocki: Actually Prettier is run on before every commit, and I don't see in "Files changed" tab that anything was changed by Prettier, so I guess we can assume that it is safe to upgrade this dependency?

preludek (approved):

PR #1357, opened 2022-04-12
ADEN-11725 Affiliate Unit and In-Content Ad are not co-existing properly

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11725


src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -10,6 +10,7 @@ const NATIVO_LIBRARY_URL = '//s.ntv.io/serve/load.js';
 export class Nativo {
 	static INCONTENT_AD_SLOT_NAME = 'ntv_ad';
 	static FEED_AD_SLOT_NAME = 'ntv_feed_ad';
+	static NATIVO_STATUS_DISABLED = 'nativo_disabled';
```
bckatarzyna, line 13: https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/models/ad-slot.ts#L77
That's the place where we keep slot events/statuses.
And I think we don't need the special status for nativo as events [here](https://github.com/Wikia/silver-surfer/pull/973/files#diff-26662de6cb74ab7450fc75c3181f40020ab2a456fbe1e196f5d171273e2f6d01R27) ale filtered by provider: nativo three lines earlier. So as long as we have provider [here](https://github.com/Wikia/ad-engine/pull/1357/files#diff-ecb2abfcd478d08c0d066f0a71631cfe63165e48f45f86834f1b28541c795571R81) we can create a general STATUS_DISABLED

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -72,7 +72,7 @@ export class Nativo {
 		}
 	}
 
-	private sendNativoLoadStatus(status: string): void {
```
bckatarzyna, line 75: Good change 👍

src/ad-engine/providers/nativo/nativo.ts
szymon-nawrocki: Excellent suggestion. Done ✅

bckatarzyna (approved):

PR #1360, opened 2022-04-14
ADEN-11737-2 Fixing Old UAP + sticky nav third time - properly this time

dmnsobczak:


platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
```diff
@@ -52,9 +52,11 @@ export class BfaaBootstrapHandler implements TemplateStateHandler {
 	async onLeave(): Promise<void> {
 		this.adSlot.show();
 		document.body.classList.add('has-uap');
+		document.body.classList.add('uap-old');
```
nandy-andy, line 55: Consider keeping the convention and changing the calss name from `uap-old` to `has-uap-old` or `has-old-uap` 😊 It's a bit nit-picking because hopefully soon we'll get rid of the old UAP but on the other hand it's just live for selected subset of our traffic and you never know how fast we'll get to 100%, so maybe it's worth keeping the class names in the same convention.

dmnsobczak: UAP 4 started a new convention and for it it's `.has-uap .uap-light`, so I followed with `.has-uap .uap-old` :D Class `.has-uap-old` was used in UAP 3, which is now removed from the code 👍

nandy-andy: It's sad no ADR was created for discussing it 🙈

dmnsobczak: Also: it is a temporary class, whole CSS sheet requires a major cleanup after removing UAP 2.0 🔪

nandy-andy (approved):
Let's give it a try...

platforms/shared/templates/handlers/bfaa/bfaa-bootstrap-handler.ts
nandy-andy: 🤞 hopefully it'll happen sooner than later

PR #1361, opened 2022-04-14
ADEN-11743 debugging SM with Kibana

Tomasz-at-Fandom:


jakubbanas (approved):

src/ad-engine/services/slot-data-params-updater.ts
```diff
@@ -1,6 +1,21 @@
 import { AdSlot } from '../models';
 import { context } from './context-service';
 import { slotTweaker } from './slot-tweaker';
+import { KibanaLogger } from '../../../platforms/shared/sequential-messaging/kibana-logger';
```
dmnsobczak, line 4: Circular dependency - we cannot import in core objects from platforms :(

Tomasz-at-Fandom: But it doesn't raise an error 🤔 
OK so first of all I should have put the whole SM feature in core, because it's platform agnostic, but I would rather do it later in a separate PR to make it less confusing.
Second, it's just temporary to debug an issue, this will be removed later on.
Third, in what way is it a circular dependency? KlibanaLogger doesn't import any thing from `slot-data-params-updater.ts` so where is the loop?

dmnsobczak: It should :( Basically I remember the rules that:
* You cannot import from ad-products/services/tracking in ad-engine
* You cannot import from platforms in src
It also has a logic explanation 🤔

Tomasz-at-Fandom: I recall that rule, but I don't think it was because JS interpreter got in a loop every time someone tried doing it. There was some tool that kept an eye on these things and raised an error during or before bundling. It seems that at some point we lost that. I'm not sure when and why, the only thing that comes to my mind is that thing that @nandy-andy  did with static code analysis during commit, but I don't know the details of how was that done and if it actually has anything to do with what we are talking about now 🤔

Tomasz-at-Fandom: Anyway, I got rid of this import, it was only for type checking anyway. 
I'm just checking if the object has the functions with the expected name, should be good enough

dmnsobczak (approved):

PR #1368, opened 2022-04-19
ADEN-11764 Fix Nativo regression after implementing NoAdsExperimentSetup

nandy-andy:


preludek:
With this implementation, nativo AdSlots are being created twice which causes some undesired side effects like for example sending 2 success events.

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -114,6 +114,12 @@ export class Nativo {
 			communicationService.getGlobalAction(eventsRepository.AD_ENGINE_SLOT_EVENT)(payload),
 		);
 	}
+  
+  private isDisabledInNoAdsExperiment(slotName): boolean {
+		const experimentUnit = this.context.get('state.noAdsExperiment.unitName');
```
preludek, line 119: I can't find where this is set. It seems like it is missing and it breaks the current implementation.

preludek (changes requested):

preludek (approved):

spec/platforms/shared/setup/noads-experiment.setup.spec.ts
```diff
@@ -21,8 +22,9 @@ describe('NoAdsExperimentSetup', () => {
 	const configGetSpy = sandbox.spy(instantConfig, 'get');
 	const cookieGetSpy = sandbox.spy(cookieAdapter, 'getItem');
 
-	afterEach(function() {
+	afterEach(function () {
 		sandbox.restore();
+		context.remove('state.noAdsExperiment.unitName');
```
preludek, line 27: <3

platforms/shared/setup/noads-experiment.setup.ts
```diff
@@ -58,6 +58,8 @@ export class NoAdsExperimentSetup implements DiProcess {
 			return;
 		}
 
+		context.set('state.noAdsExperiment.unitName', unitName);
```
preludek, line 61: As long as beacons don't overlap it's 👍  we just have to remember to adjust it if we want to use that mechanism for more than 1 slot at a time

preludek (approved):

platforms/shared/setup/noads-experiment.setup.ts
nandy-andy: Definitely, if we want to cross-check the ad units we'll need to make it an array ;)

PR #1378, opened 2022-04-21
ADEN-11731 The easiest way to pass a key-val to DW

nandy-andy:


spec/ad-engine/tracking/slot-tracking-middleware.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+		adSlot.orderId = 3;
+		adSlot.status = 'success';
+
+		sandbox.stub(adSlot, 'getUid').returns('a2b01b9f-69df-4481-9daf-9e27bf26502b');
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+
+		context.remove('targeting.cortex-visitor');
```
szymon-nawrocki, line 25: Using global `context` shouldn't cause any issues in this case, but it would be nice to keep up with a good practise in our tests and use the mock of `context` object

spec/ad-engine/tracking/slot-tracking-middleware.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+		slotTrackingMiddleware({ data: {}, slot: adSlot }, nextSpy);
+
+		expect(nextSpy.getCall(0).args[0].data).to.deep.include({
+			key_vals: 'cortex-visitor=no',
+		});
+	});
+
+	it('does not return info about cortex-visitor if not set in targeting', () => {
+		const nextSpy = sandbox.spy();
+		context.remove('targeting.cortex-visitor');
```
szymon-nawrocki, line 53: This one is already removed after a previous test (`afterEach`)

src/ad-tracking/slot-tracking-middleware.ts
```diff
@@ -39,7 +49,7 @@ export const slotTrackingMiddleware: FuncPipelineStep<AdInfoContext> = ({ data,
 			device: utils.client.getDeviceType(),
 			document_visibility: utils.getDocumentVisibilityStatus(),
 			is_uap: isUap ? 1 : 0,
-			key_vals: '',
+			key_vals: [checkCortexVisitor()].join(';'),
```
szymon-nawrocki, line 52: Can't it be just done like that:
`key_vals: context.get('targeting.cortex-visitor') || ''`
??

szymon-nawrocki: Okay I see what you're trying to do here, so nevermind :)

spec/ad-engine/tracking/slot-tracking-middleware.spec.ts
nandy-andy: I know it's not popular in the JS world but I wanted to be as [explicit](https://peps.python.org/pep-0020/) as possible here ;)

src/ad-tracking/slot-tracking-middleware.ts
nandy-andy: It could but I assume the plural in the key name (key_vals) was meant to keep more there ;) If it works I was thinking of adding there number of Audigent segments 😈

szymon-nawrocki (approved):
👌

PR #1385, opened 2022-04-26
ADEN-11773 Adding New Takeover Configurations

dmnsobczak:


platforms/ucp-desktop/styles.scss
```diff
@@ -24,6 +24,23 @@
...
+				position: sticky;
+				z-index: 100;
+			}
+		}
+
+		&:not(.uap-sticked) {
+			.uap-companion-stick {
+				.main-page-tag-rcs,
+				#rail-boxad-wrapper {
+					top: 36px !important;
```
szymon-nawrocki, line 39: Since its SASS it would be worth considering adding "36px" to a variable `$navbar-height` so the styles are more readable

szymon-nawrocki (approved):
👌

platforms/ucp-desktop/styles.scss
dmnsobczak: Fix applied 👍

PR #1388, opened 2022-04-28
ADEN-11695 Optimera Integration trial period

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11695


spec/ad-services/optimera/optimera.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+	beforeEach(() => {
+		loadScriptStub = sandbox.stub(optimera, 'loadScript').returns(Promise.resolve({} as any));
+		contextStub = sandbox.stub(context);
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	it('when enabled - loads optimera script', async () => {
```
nandy-andy, line 20: Nit-picking: if you change the segments of the comment it'll be more like a story ;-) It'll display something like:
```
Optimera
✔️ loads optimera script when enabled 
✔️ returns rejected promise when disabled
✔️ setTargeting is called when script is loaded
```
Nothing urgent/blocking, so just a food for thoughts :)

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
+import { context, utils } from '@ad-engine/core';
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+
+const logGroup = 'Optimera';
+const CLIENT_ID = '82';
+
+const oDv = [CLIENT_ID, 'top_leaderboard', 'top_boxad', 'bottom_leaderboard'];
```
nandy-andy, line 7: Consider renaming the variable, so it's more descriptive. I wonder how to call it, though. Maybe `optimeraConfiguration` or `optimeraVariables`?

szymon-nawrocki: I've used the same names as Optimera requires in their script, so they are not misleading

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+let oVa = {
```
nandy-andy, line 8: A similar thought here but by reading the code below I have no idea how to call it 😅 It looks some kind of an ad slots dictionary. But except it's being copied in the line 42 it's hard to tell what happens with it next ;p It's like global overwritten later in the line 88 🤔

platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
```diff
@@ -88,6 +89,7 @@ export class UcpDesktopAdsMode implements DiProcess {
 		const inhibitors: Promise<any>[] = [];
 		const targeting = context.get('targeting');
 
+		inhibitors.push(optimera.call());
```
nandy-andy, line 92: We could add time tracking in order to check how much it delays the bidders' calls and the ad queue.

src/ad-services/optimera/index.ts
szymon-nawrocki: I admit that it is not elegant solution.
The `oVa` variable is first hardcoded into script -> when script is loaded, it modifies global `oVa` variable -> this modified variable is used in `setTargeting()` method

szymon-nawrocki: So since these variables are hardcoded anyway, maybe I should delete these from lines 7 and 8?

nandy-andy: Yeah, maybe move them to `getHardcodedScript()` is a good idea 🤔 Additionally, it shouldn't be a problem to have in this very function:
```js
var oDv = ${JSON.stringify(optimeraVariables)};
var oVa = ${JSON.stringify(optimeraSlots)};
```

platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
szymon-nawrocki: But on the other hand I don't think we need to call Optimera before bidders, so maybe I should just move it line below `inhibitors.push(bidders.requestBids())`, so as a result it won't slow down bidder.

szymon-nawrocki: Done ✅ 
Tested it and everything is working correctly. I hope I understand it right and that solves possible "bidders calls delay" problem.

But there is still a question: "Will it delay ad queue? If it will, how much?". 
@dmnsobczak is there a way to check it in our DW?

szymon-nawrocki: But since `optimera.call()` is wrapped with a `Promise` and everything is done asynchronously, will it really delay bidders and the ad queue?

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+			utils.logger(logGroup, 'loading failed.', e.message);
+			return Promise.reject('Optimera loading failed');
+		}
+	}
+
+	getHardcodedScript(): HTMLScriptElement {
+		const script = document.createElement('script');
+		script.type = 'text/javascript';
+		script.innerHTML = `
+			var oDv = ['${CLIENT_ID}', 'top_leaderboard', 'top_boxad', 'bottom_leaderboard'];
```
dmnsobczak, line 34: Slot names can be moved to the separate config variable.

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+			optimeraScript.async = true;
+			optimeraScript.type = 'text/javascript'; 
+			/* IF AVAILABLE: PAGE AD MAP CONFIGURATION */
+			optimeraScript.onload = function() {
+			  if(typeof oDevice == 'string' && typeof oVa.device == 'object' && typeof oVa.device[oDevice] == 'object') {
+			   oVa = oVa.device[oDevice];
+			  }
+			}
+			
+			optimeraScript.src="https://dyv1bugovvq1g.cloudfront.net/"+oDv[0]+"/"+optimeraHost+optimeraPathName+".js";
```
dmnsobczak, line 52: Even if this is a Optimera's code we can improve the formatting a bit :D

src/ad-services/optimera/index.ts
szymon-nawrocki: This is what I've done before 😄:
https://github.com/Wikia/ad-engine/pull/1388#discussion_r865817231

platforms/shared/setup/base-context.setup.ts
```diff
@@ -176,6 +176,7 @@ export class BaseContextSetup implements DiProcess {
 		context.set('services.sailthru.enabled', this.instantConfig.get('icSailthru'));
 		context.set('services.stroer.enabled', this.instantConfig.get('icStroer'));
 		context.set('services.liveConnect.enabled', this.instantConfig.get('icLiveConnect'));
+		context.set('services.optimera.enabled', this.instantConfig.get('icOptimera'));
```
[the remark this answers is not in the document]
szymon-nawrocki: But nothing is in alfabetical order here 😄

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
+import { context, utils } from '@ad-engine/core';
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+
+const logGroup = 'Optimera';
```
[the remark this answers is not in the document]
szymon-nawrocki: done ✅

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+			}
+			
+			optimeraScript.src="https://dyv1bugovvq1g.cloudfront.net/"+oDv[0]+"/"+optimeraHost+optimeraPathName+".js";
+			var node = document.getElementsByTagName('script')[0];
+			node.parentNode.insertBefore(optimeraScript, node);
+			
+			var optimeraOpsScript = document.createElement('script');
+			optimeraOpsScript.async = true;
+			optimeraOpsScript.type = 'text/javascript';
+			optimeraOpsScript.src="https://d15kdpgjg3unno.cloudfront.net/oPS.js?cid="+oDv[0];
```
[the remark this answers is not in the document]
szymon-nawrocki: At the beginning I've tried writing it using AdEngine's tools and helpers but it did not work out. I am not sure if it was because I did something wrong in the code, or it was the problem on their side (I was getting 403 code error when I tried to load "score file"). 

So I decided to not complicate things and just load hardcoded version of script. Then when we test Optimera and see that they are performing great, refactor the code to make it better.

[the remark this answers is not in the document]
szymon-nawrocki: Plus maybe when we see that they are perform great we can go with the Prebid implementation version which looks more elegant

src/ad-services/optimera/index.ts
szymon-nawrocki: But more seriously though, what is wrong with the formatting here? double quotes?

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+
+	async loadScript(): Promise<void> {
+		document.head.prepend(this.getHardcodedScript());
+		return Promise.resolve();
+	}
+
+	setTargeting(): void {
+		communicationService.on(eventsRepository.AD_ENGINE_STACK_START, () => {
+			if (!window.oVa) {
+				throw new Error(`Optimera 'oVa' variable is not defined`);
```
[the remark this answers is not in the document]
szymon-nawrocki: I threw an error here, so it can be caught and logged in try catch block of `call()` function, is this a bad practise?

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+
+		try {
+			await this.loadScript();
+			utils.logger(logGroup, 'loaded');
+
+			this.setTargeting();
+			return Promise.resolve('Optimera loaded with success');
+		} catch (e) {
+			utils.logger(logGroup, 'loading failed.', e.message);
+			return Promise.reject('Optimera loading failed');
```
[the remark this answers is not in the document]
szymon-nawrocki: I've changed it so the function returns `Promise.resolve()` only when optimera is disabled, like in `wad-runner` https://github.com/Wikia/ad-engine/blob/365270bfeb3a32e825ed5d1e9dcd1c52885b2765/platforms/shared/services/wad-runner.ts#L7

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,97 @@
...
+			/*OPTIONAL TIMEOUT SCRIPT*/
+			var s = +new Date;
+			while (+new Date < s + 200) {/*Default 200ms Timeout*/};
+			})();
+		`;
+		return script;
+	}
+
+	async loadScript(): Promise<void> {
+		document.head.prepend(this.getHardcodedScript());
```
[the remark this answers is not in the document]
szymon-nawrocki: Yes, it would be nice to use our `utils.scriptLoader` to load this script, but in this case we are loading a script, that loads another scripts and contains global variables, which makes  `utils.scriptLoader` less applicable.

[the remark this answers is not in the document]
szymon-nawrocki: So maybe I should load these scripts separately 🤔

src/ad-services/optimera/index.ts
szymon-nawrocki: But restored it because I went with your suggestion to use AdEngine tools to load Optimera scripts

src/ad-services/optimera/index.ts
szymon-nawrocki: Improved ✅

src/ad-services/optimera/index.ts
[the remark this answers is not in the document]
szymon-nawrocki: But I went with your suggestion, gave it a second try, and loaded the scripts with `utils.loadScript()` method, so consider it done ✅

src/ad-services/optimera/index.ts
szymon-nawrocki: By using our AdEngine tools 😄

src/ad-services/optimera/index.ts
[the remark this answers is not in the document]
szymon-nawrocki: I think that it is a great idea. Done ✅

platforms/shared/setup/base-context.setup.ts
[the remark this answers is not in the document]
dmnsobczak: So let's reorder everything :D Or at least `services` group :D

platforms/shared/tracking/tracking.setup.ts
```diff
@@ -352,4 +353,14 @@ export class TrackingSetup {
 			false,
 		);
 	}
+
+	private optimeraTracker(): void {
+		communicationService.on(
+			eventsRepository.OPTIMERA_LOADED,
+			() => {
+				this.pageTracker.trackProp('optimera', 'loaded');
```
dmnsobczak, line 361: Maybe instead `loaded` let's track the number which states what was the load time?
(like we do for `eventsRepository.ATS_JS_LOADED`) 🤔

spec/ad-engine/utils/wait-for.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+			obj1: 'null',
+		};
+
+		setTimeout(() => {
+			windowObj.obj1 = 'something';
+		}, 80);
+
+		const conditionToWaitFor = () => windowObj.obj1 === 'something';
+
+		const promise = await new utils.WaitFor(conditionToWaitFor, 3, 10).until();
```
dmnsobczak, line 31: It's weird to see it non-camelcase :D But whatever I guess 🤷

src/ad-engine/utils/wait-for.ts
```diff
@@ -0,0 +1,27 @@
+type Condition = () => boolean;
+
+export class WaitFor {
+	constructor(
+		private condition: Condition,
+		private noTries: number,
+		private initialTimeoutMs: number,
```
dmnsobczak, line 7: Those variables can have default values for simplification

src/communication/event-types.ts
```diff
@@ -261,4 +261,7 @@ export const eventsRepository: Dictionary<EventOptions> = {
 		name: 'Load template',
 		payload: payload<LoadTemplatePayload>(),
 	},
+	OPTIMERA_LOADED: {
```
dmnsobczak, line 264: Please move it under L172 to the `// Integrated partners events //` section

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+		try {
+			this.loadGlobalVariablesScript();
+			await this.loadScoreFileScript();
+			this.sendTrackingEvent();
+			await this.updateOvaVariable();
+			await this.loadOpsScript();
+			this.setTargeting();
+		} catch (e) {
+			utils.logger(logGroup, 'loading failed.', e.message);
```
dmnsobczak, line 34: This `.` is a misclick? :D

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	async loadScoreFileScript(): Promise<void> {
+		const optimeraHost = window.location.host;
+		const optimeraPathName = window.location.pathname;
+		const url = `https://dyv1bugovvq1g.cloudfront.net/${CLIENT_ID}/${
+			optimeraHost + optimeraPathName
+		}.js`;
+
+		await utils.scriptLoader.loadScript(url);
```
dmnsobczak, line 58: ❤️

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	sendTrackingEvent(): void {
+		communicationService.emit(eventsRepository.OPTIMERA_LOADED);
+	}
+
+	// Variable 'oVa' needs to be updated as it gets overwritten after loading score file
+	async updateOvaVariable(): Promise<void> {
+		const conditionMet = await new utils.WaitFor(this.isConfigUpdated, 3, 100).until();
+		if (!conditionMet) {
+			return Promise.reject(new Error('oVa variable not updated'));
```
dmnsobczak, line 70: Won't it break AdEngine execution?

dmnsobczak (approved):
COde is fine, few minor comments left - accept them or ignore and proceed to QA 💪

src/ad-engine/utils/wait-for.ts
```diff
@@ -0,0 +1,27 @@
...
+
+	async until(): Promise<boolean> {
+		let result = this.condition();
+		let timeout = this.initialTimeoutMs;
+		let iteration = 0;
+
+		while (!result && iteration < this.noTries) {
+			await this.delay(timeout);
+			result = this.condition();
+			timeout *= 2;
```
mfarul, line 18: this 2 could also be a parameter, by default set to 1,5 ;)

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,120 @@
...
+		`;
+
+		document.head.prepend(script);
+		utils.logger(logGroup, 'global variables loaded');
+	}
+
+	async loadScoreFileScript(): Promise<void> {
+		const optimeraHost = window.location.host;
+		const optimeraPathName = window.location.pathname;
+		const url = `https://dyv1bugovvq1g.cloudfront.net/${CLIENT_ID}/${
```
mfarul, line 54: i would extract part of this URL to constant - easier to manage when seen on top of the file ... or from config.

src/ad-services/optimera/index.ts
```diff
@@ -0,0 +1,120 @@
...
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+			return Promise.resolve();
+		}
+
+		try {
+			this.loadGlobalVariablesScript();
+			await this.loadScoreFileScript();
+			this.sendTrackingEvent();
+			await this.updateOvaVariable();
```
mfarul, line 30: here we don't react on the result - does it make sense to run the rest of the logic here if we did not wait enough for optimera?

spec/ad-engine/utils/wait-for.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { utils } from '@wikia/ad-engine';
+import { expect } from 'chai';
+
+describe('WaitFor', () => {
+	it('when condition is met withing the specified time returns true', async () => {
+		const windowObj = {
+			obj1: 'null',
+		};
+
+		setTimeout(() => {
```
mfarul, line 10: that's tricky - this could lead to flapping tests, maybe we could change it based on interactions, f.e. after 2nd call to the condition it would change it state

platforms/shared/setup/base-context.setup.ts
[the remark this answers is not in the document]
szymon-nawrocki: done ✅

src/communication/event-types.ts
szymon-nawrocki: done ✅

src/ad-services/optimera/index.ts
szymon-nawrocki: It won't because it will be catched in `catch` block of `call` method, here:
https://github.com/Wikia/ad-engine/blob/aad7fbd58df2d97a9f124eb45c4aad5545d2249b/src/ad-services/optimera/index.ts#L33

src/ad-services/optimera/index.ts
szymon-nawrocki: done ✅

src/ad-services/optimera/index.ts
szymon-nawrocki: Actually we react on the result here:
https://github.com/Wikia/ad-engine/blob/3fb678b6a2efbd9647dda5f20b7f5b6742376eec/src/ad-services/optimera/index.ts#L70

spec/ad-engine/utils/wait-for.spec.ts
szymon-nawrocki: You're right. I've changed it to make it more reliable, I hope this is something similar to your idea

src/ad-services/optimera/index.ts
szymon-nawrocki: Yes it was 😄

PR #1389, opened 2022-04-29
Evaluate Ads refresh on viewable slots

jakubbanas:


spec/ad-engine/services/slot-refresher.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+		} as AdSlot;
+
+		slotRefresher.setupSlotRefresher(basicConfig, false, loggerSpy);
+		slotRefresher.refreshSlotIfPossible(fakeSlot);
+		clock.runAll();
+
+		assert.calledOnce(refreshSpy);
+	});
+
+	it('should not refresh refresh if slot is not enabled', () => {
```
preludek, line 61: refresh refresh? 🤔

spec/ad-engine/services/slot-refresher.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+		} as AdSlot;
+
+		slotRefresher.setupSlotRefresher(basicConfig, false, loggerSpy);
+		slotRefresher.refreshSlotIfPossible(fakeSlot);
+		clock.runAll();
+
+		assert.notCalled(refreshSpy);
+	});
+
+	it('should not refresh refresh if slot is not registered in GPT', () => {
```
preludek, line 74: ditto 😄

spec/ad-engine/services/slot-refresher.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+	const sandbox = sinon.createSandbox();
+	const clock = sinon.useFakeTimers({
+		toFake: ['setTimeout'],
+	});
+
+	const loggerSpy = sandbox.spy();
+	const refreshSpy = sandbox.spy();
+
+	beforeEach(function () {
+		window.googletag = {
```
preludek, line 18: Shouldn't we clear that after tests? I think this may be leaking.

src/ad-engine/models/ad-slot.ts
```diff
@@ -137,7 +137,7 @@ export class AdSlot {
 	requested = null;
 	loaded = null;
 	rendered = null;
-	viewed = null;
+	viewed: Promise<void> = null;
```
preludek, line 140: 👍 I think we can type all of those (requested, loaded, rendered)

src/ad-engine/services/slot-refresher.ts
```diff
@@ -0,0 +1,64 @@
...
+	});
+}
+
+class SlotRefresher {
+	config: Config;
+	log(...logValues) {
+		logger(logGroup, ...logValues);
+	}
+
+	refreshSlotIfPossible(slot: AdSlot) {
```
preludek, line 32: Maybe just `refreshSlot`?

preludek (approved):

preludek (approved):

src/ad-engine/services/slot-injector.ts
```diff
@@ -5,7 +5,7 @@ import { context } from './context-service';
 import { SlotCreator, SlotCreatorConfig } from './slot-creator';
 import { slotService } from './slot-service';
 
-const logGroup = 'slot-repeater';
+const logGroup = 'slot-injector';
```
nandy-andy, line 8: 💎

src/communication/communication-service.ts
```diff
@@ -69,6 +69,19 @@ export class CommunicationService {
 			.subscribe(callback);
 	}
 
+	onLast(eventName: string | symbol, callback: (payload?: any) => void, slotName = ''): void {
```
nandy-andy, line 72: I think we can remove it as we don't use it 🤔

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -84,6 +84,37 @@ function configure(): void {
...
+		},
+	);
+
+	tag.addEventListener(
+		'slotVisibilityChanged',
+		function (event: googletag.events.SlotVisibilityChangedEvent) {
+			const adSlot = getAdSlotFromEvent(event);
+
+			if (event.inViewPercentage < 50) {
+				return adSlot.emit(AdSlot.SLOT_LEFT_VIEWPORT, event);
```
nandy-andy, line 113: With the three new listeners we're spamming the Redux console with events. If I understand correctly `adSlot` here is an observable and we could consider using `pipe()` method on it with the `debounce` operator in order to spam less maybe 🤔 https://www.learnrxjs.io/learn-rxjs/operators/filtering/debounce

jakubbanas: The event is already throttled by GPT itself, it only emits a new event every 200ms. https://developers.google.com/publisher-tag/reference#googletag.events.slotvisibilitychangedevent

I believe that increasing this debounce/throttle time might cause a refresh of the slot that is not in the viewport.
Also, this event is emitted only during scroll, when it is actually needed.

Altho, the event will be emitted anyway, we make it visible in the console so that the change would be only a developer experience improvement.

src/ad-engine/models/ad-slot.ts
```diff
@@ -294,6 +297,15 @@ export class AdSlot {
 		return Array.isArray(this.creativeSize) ? this.creativeSize.join('x') : this.creativeSize;
 	}
 
+	getCreativeSizeAsArray(): [number, number] | null {
+		const size = Array.isArray(this.creativeSize)
+			? this.creativeSize
+			: this.creativeSize.split('x').map(Number);
+
+		// Type hack to make sure it will always return two element array
+		return [size[0], size[1]];
```
preludek, line 306: this.creativeSize can be null and is null by default. This will break then.

spec/ad-engine/models/ad-slot.stub.ts
```diff
@@ -34,6 +34,7 @@ export function createAdSlotStub(sandbox: SinonSandbox): AdSlotStub {
 		getIframe: sandbox.stub(),
 		overrideIframe: sandbox.stub(),
 		getCreativeSize: sandbox.stub(),
+		getCreativeSizeAsArray: sandbox.stub(),
```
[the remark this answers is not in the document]
jakubbanas: Yeah, I guess we could remove that.

src/ad-engine/models/ad-slot.ts
jakubbanas: Good point

platforms/ucp-desktop/styles.scss
```diff
@@ -50,6 +50,12 @@ $sticky-navigation-height: 46px;
 		}
 	}
 
+	.has-sticky-tlb #top_leaderboard div {
+		// important needed because some creatives contains inline CSS overriding this properties
```
nandy-andy, line 54: Wouldn't be needed if we serve everything in a SafeFrame 😿 But then Fan Takeovers would be more complicated 🙈 Although, with LIS not necessarily 😈

nandy-andy (approved):

PR #1390, opened 2022-04-29
CATS-1302 | Add city / country_name to ensured Geo cookie

pstacherzak:
See https://github.com/Wikia/wikia-vcl/pull/3055

### DIFF
--- a/platforms/shared/ensure-geo-cookie.ts
+++ b/platforms/shared/ensure-geo-cookie.ts
@@ -4,12 +4,16 @@ interface GeoResponse {
 	continent_code: string;
 	country_code: string;
 	region: string;
+	city: string;
+	country_name: string;
 }
 
 interface GeoData {
 	continent: string;
 	country: string;
 	region: string;
+	city: string;
+	country_name: string
 }
 
 export async function ensureGeoCookie(): Promise<void> {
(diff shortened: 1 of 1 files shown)


platforms/shared/ensure-geo-cookie.ts
```diff
@@ -4,12 +4,16 @@ interface GeoResponse {
 	continent_code: string;
 	country_code: string;
 	region: string;
+	city: string;
+	country_name: string;
```
szwajkowski, line 8: just wanted to let you know that there is yet another source of Geo cookie - https://github.com/Wikia/tracking-opt-in/blob/master/src/shared/GeoManager.js#L93-L141 I don't know how often it is used but I guess @dmnsobczak will know, maybe we need to DRY it a bit.

pstacherzak: Oh, great finding, I'll apply a patch there as well. @dmnsobczak do you know why the cookie needs to be applied in tracking-opt-in?

szwajkowski: cookie in track-opt-in is used to determine whether to show GDPR/CCPA modal or not - it depends on the user's location

nandy-andy (approved):

preludek (approved):

dmnsobczak (approved):

PR #1392, opened 2022-05-05
ADEN-11182 Switching fully to UAP 4.3

dmnsobczak:


platforms/f2/templates/bfaa-template.ts
```diff
@@ -27,77 +23,82 @@ import {
...
+		],
+		resolved: [
+			SlotSizeResolvedWithPlaceholderHandler,
+			SlotHeightClippingHandler,
+			VideoSizeResolvedHandler,
+			DomCleanupHandler,
+		],
+	};
+
+	if (context.get('state.isMobile')) {
```
Tomasz-at-Fandom, line 71: This tempts me to suggest creating two separate `registerBfaaTemplate` functions, one for mobile and one for desktop, with the commonalities shared of course from one source. However if you don't expect more variations like this `if` coming to this function in the near future, than let's not bother. I leave this up to you to decide.

Tomasz-at-Fandom (approved):

platforms/f2/templates/bfaa-template.ts
dmnsobczak: Yeah, I was thinking about it: discussable "if" vs 2 files with multiple duplicated lines... and I choose the first option. In the future with any UAP improvement I would like to unify the list of handlers for mobile and desktop and across all platforms and then move this template declaration to `shared` :) I don't want to expand the scope of this ticket even more (I already merged reimplementation and cleanup together), so I would rather prefer to do it later 🙇

PR #1394, opened 2022-05-11
ADEN-11808 | Add identity DW tracking

bckatarzyna:


src/communication/event-types.ts
```diff
@@ -264,4 +269,8 @@ export const eventsRepository: Dictionary<EventOptions> = {
 		name: 'Load template',
 		payload: payload<LoadTemplatePayload>(),
 	},
+	IDENTITY_PARTNER_DATA_OBTAINED: {
```
szymon-nawrocki, line 272: I think it should be placed under `	// Integrated partners events //` (line 145)

szymon-nawrocki: When you decide to move it could you also move `OPTIMERA_FINISHED` event, because now I see I placed it in the wrong place?

src/ad-services/live-connect/index.ts
```diff
@@ -28,11 +30,23 @@ class LiveConnect {
...
 					utils.logger(logGroup, 'loaded');
+					this.track();
 				});
 
 			this.isLoaded = true;
 		}
 	}
+
+	track(): void {
+		return window.liQ.resolve((nonId) => {
```
szymon-nawrocki, line 41: I think it will be good to check if `window.liQ` object really exist with something like:
`if (!window.liQ) {
    console.warn(logGroup, 'not available or not initialised!');
    return;

}`

PR #1395, opened 2022-05-12
ADEN-11791 Integrating NTC to FV pages

dmnsobczak:


platforms/shared/setup/ad-layout-initializer.setup.ts
```diff
@@ -62,25 +71,16 @@ export class AdLayoutInitializerSetup implements DiProcess {
 
 				utils.logger(logGroup, 'Layout payload received', layoutPayload);
 
-				// TODO: move the logic below to UapAdLayout
 				if (['uap', 'jwplayer'].includes(layoutPayload.layout)) {
```
nandy-andy, line 74: I'd leave the TODO as it all still can be moved to a separate class(es) ;) A layout factory that creates concrete layouts like FanTakeoverLayout or NtcFanTakeoverLayout :P

platforms/shared/setup/ad-layout-initializer.setup.ts
```diff
@@ -17,12 +20,18 @@ interface FanTakeoverLayoutPayload extends LayoutPayload {
 	data: {
 		// TODO: possibly change to impressionMacro in order to be consistent with GAM docs WARNING! changing it here requires changes in GAM templates for existing LIS campaigns
 		impression: string;
-		lineItemId: number;
-		creativeId: number;
+		lineItemId: string;
+		creativeId: string;
```
nandy-andy, line 24: Why can't we have numbers here?

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -38,5 +42,51 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
+			this.configureStickingCompanion();
+		});
+	}
+
+	private configureStickingCompanion(): void {
+		communicationService.onSlotEvent(
+			AdSlot.STATUS_SUCCESS,
+			() => {
+				const pageElement = document.querySelector('.page');
+				const rightRailElement = document.querySelectorAll(
```
nandy-andy, line 56: We start to use the selector in [more places](https://github.com/Wikia/ad-engine/blob/478813e28adaf798fb8ba62568263e7e2a4af18d/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts#L64). Maybe it's worth extracting it to a new shared function `getRightRailElement()`? 🤔

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -38,5 +42,51 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
+					'.main-page-tag-rcs, #rail-boxad-wrapper',
+				)[0] as HTMLElement;
+
+				if (!pageElement || !rightRailElement) {
+					return;
+				}
+
+				pageElement.classList.add('uap-companion-stick');
+
+				communicationService.onSlotEvent(
```
nandy-andy, line 66: I'd extract listening to `CUSTOM_EVENT` and `SLOT_VIEW_EVENT` outside the body of `STATUS_SUCCESS` listener. This way the block of code would be more readable and we'd get rid of a side effect when we're registering more listeners on each `top_boxad` success (for example after a reload that most likely won't happen in UAP but we never know ;)). I'd call the extracted listeners along with the listener for success, though. So, the `configureStickingCompanion()` would look like:
```js
private configureStickingCompanion(): void {
  registerTopBoxadSuccessListener();
  registerTopBoxadViewedListener();
  registerTopBoxadStickedListener();
}
```
and will still be called on `AD_ENGINE_UAP_NTC_LOADED` event 🤔

platforms/shared/setup/ad-layout-initializer.setup.ts
dmnsobczak: We are passing strings from GAM (they are quoted) so I adjusted it to the real type. Also this value is then passed to `context.set()` method which requires strings so it fixed a TypeScript issue in that place.
Once we finish LIS testing and before creating a documentation for it we need to sit down and agree a data format that LIS will be passing from GAM - then we can name and type those variables properly.

platforms/shared/setup/ad-layout-initializer.setup.ts
dmnsobczak: Fine, left :D More ToDo's are harder to omit during productizing this feature!

nandy-andy: You can also remove the TODO and implement the class :)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -67,6 +68,24 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
...
 		]);
 
+		if (context.get('custom.hasFeaturedVideo')) {
+			communicationService.on(eventsRepository.AD_ENGINE_UAP_NTC_LOADED, () => {
+				new utils.WaitFor(
+					() =>
+						!document.querySelector('body').classList.contains('featured-video-on-scroll-enabled'),
+					2500,
+					0,
+					500,
```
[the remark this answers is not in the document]
dmnsobczak: Those numbers are labeled if you use IDE to browse this code 🤔 
```	constructor(
		private condition: Condition,
		private noTries: number = 100,
		private growingTimeout: number = 0,
		private fixedTimeout: number = 0,
	) {}```

File: platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts (hunk 4 above)
Comment (dmnsobczak) [reply platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts:66]: Disagree :/ `communicationService.onSlotEvent()` is a function call that passes all data required data to Event Engine to register it. After that we will call a method that only calls a method - like a matryoshka doll 🙈  Easier way to explain this code is to add some comments. For me to extract something to a method we should have more that one line or a chunk of code that is reused somewhere else.
Also: `top_boxad` success event listener won't be fired twice, because Duration Media reload is not triggering it and (for now) we don't have any other mechanisms like that 🤔

Comment (dmnsobczak) [reply platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts:66]: Done

File: platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (dmnsobczak) [reply platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts]: Done

**Review by nandy-andy — APPROVED** [review]
Thanks for the changes. Since it's an experiment I feel OK with testing it and running the experiment. It would be good, though, if we'd have a ticket to clean the code up once we learnt if the feature stays or not.

File: platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts [hunk 6]
```diff
@@ -67,6 +68,27 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
 			}),
 		]);
 
+		if (context.get('custom.hasFeaturedVideo')) {
+			communicationService.on(eventsRepository.AD_ENGINE_UAP_NTC_LOADED, () => {
+				const noTries = 2500;
+				const retryTimeout = 500;
+
+				new utils.WaitFor(
```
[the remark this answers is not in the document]
dmnsobczak: Yes, that's exactly what that means. And unlock the Promise waiting for this `wait` :)
I could pass an event from UCP repository but testing and releasing it would take more time and I wanted to make it ready to review for Nef as quickly as possible.

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -38,5 +42,64 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
+
+				this.registerStickingCompanionViewedListener();
+			},
+			'top_boxad',
+			true,
+		);
+	}
+
+	private registerStickingCompanionStickedListener(): boolean {
+		const rightRailElement = document.querySelectorAll(
```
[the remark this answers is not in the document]
dmnsobczak: Actually I didn't want to polish this code too much, because (hopefully) it will be removed in the next sprint or later, because we integrated UAP:JWP with LIS in the different ticket and there you can pass `ntc: true` from the LIS creative to switch to different AdLayout and this layout should be responsible for reconfiguring the companions. This is the creative:
https://admanager.google.com/5441#creatives/creative/detail/line_item_id=5979612610&creative_id=138387699623
So it's a waste of time to improve the readability in the chunk of code marked for removal :D

PR #1419, opened 2022-05-17
ADEN-11353 Click event not send after clicking on video

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11353


Tomasz-at-Fandom (approved):

platforms/shared/styles/uap-porvata/_bottom-panel.scss
```diff
@@ -37,6 +37,7 @@
 
 	.toggle-ui-overlay {
 		cursor: pointer;
+		pointer-events: none;
```
szymon-nawrocki, line 40: The problem with `clicked` event not being sent was caused by `<div class='toggle-ui-overlay/>`  (div with gradient) which overlays the video ad on UAP.

The `pointer-events:none` css property makes the click "fall-through" this object.

Tomasz-at-Fandom: Can we put it in a comment next to this property or class?

szymon-nawrocki: done ✅

src/ad-products/templates/interface/video/toggle-ui.ts
```diff
@@ -59,10 +59,6 @@ export class ToggleUI {
 			container.addEventListener('mouseleave', () => {
 				fadeOut();
 			});
-
-			overlay.addEventListener('click', () => {
```
szymon-nawrocki, line 63: Since the click events "fall-through" `overlay` (`toggle-ui-overlay`) element, this 'click' listener is redundant

nandy-andy (approved):

src/ad-tracking/click-tracker.ts
```diff
@@ -48,6 +48,22 @@ class AdClickTracker {
 			},
 			false,
 		);
+
+		communicationService.on(
```
nandy-andy, line 52: We keep growing the `register()` method the code with similar lines of code. Consider moving all the `communicationService.on` events with assigned to them callbacks to a private dictionary and just loop through them here. It will close the `register()` method for similar changes but we'll be still able to extend 🤔

Something of this sort:
```ts
private eventsToRegister = {
  eventsRepository.AD_ENGINE_VIDEO_LEARN_MORE_DISPLAYED: this.handleLearnMoreDisplayed,
  eventsRepository.AD_ENGINE_VIDEO_OVERLAY_CLICKED: this.handleVideoOverlayClicked,
  eventsRepository.AD_ENGINE_VIDEO_TOGGLE_UI_OVERLAY_CLICKED: this.handleVideoToggleUiOverlayClicked,
};
```

szymon-nawrocki: Great idea, done ✅

nandy-andy (approved):

PR #1420, opened 2022-05-19
ADEN-11791-2 Expanding NTC for non-LIS Roadblocks

dmnsobczak:


platforms/ucp-desktop/styles.scss
```diff
@@ -2,31 +2,14 @@
 @import 'platforms/shared/styles/index';
 @import 'src/ad-products/styles/porvata';
 
+$sticky-navigation-width: 46px;
```
szymon-nawrocki, line 5: I think it should be named `sticky-navigation-height`

dmnsobczak: Right! Renamed!

szymon-nawrocki (approved):
👌

PR #1432, opened 2022-06-03
ADEN-11799 | Integrating prebid native ads from AppNexus

bckatarzyna:


src/ad-bidders/prebid/adapters/appnexus.ts
```diff
@@ -16,31 +17,88 @@ export class Appnexus extends PrebidAdapter {
...
+						required: true,
+					},
+					icon: {
+						required: true,
+						aspect_ratios: [
+							{
+								min_width: 100,
+								min_height: 100,
+								ratio_width: 1,
+								ratio_height: 1,
```
nandy-andy, line 78: Are the `1`s and `100`s needed only for the whole adapter to work? Maybe it's worth setting some real minimal sizes like mobile's minimum 300x50? 🤔 On the other hand it's a native ad, so it doesn't really need to match regular display ads sizes. The "magic numbers" just caught my attention 😅

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	private getAssetValue(assetName: string, data: PrebidNativeData): string {
+		if (assetName == 'icon' || assetName == 'image') {
+			return data[assetName]['url'];
+		}
+		return data[assetName];
+	}
+
+	getPrebidNativeTemplate(): string {
```
nandy-andy, line 60: If we can't use here any templating system like [mustache](https://github.com/janl/mustache.js/) or just load an HTML file maybe it'll be worth changing the concatenated string to a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in order to make it more readable.

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -125,6 +126,11 @@ export class Nativo {
 		window.ntv.Events?.PubSub?.subscribe('noad', (e: NativoNoAdEvent) => {
 			const slotName = Nativo.AD_SLOT_MAP[e.data[0].id];
 			this.handleNtvNativeEvent(e, slotName, AdSlot.STATUS_COLLAPSE); // init or collapse
+			if (slotName == Nativo.INCONTENT_AD_SLOT_NAME) {
```
nandy-andy, line 129: Let's get rid off the comment above it seems it already got lost 🙈

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+				this.renderPrebidNativeAd(payload.slotName, data.native);
+			}
+			communicationService.emit(eventsRepository.NO_NATIVE_PREBID_AD, {
+				slotName: payload.slotName,
+			});
+		});
+	}
+
+	private renderPrebidNativeAd(adSlotName: string, data: PrebidNativeData): void {
+		const ntvAdSLot = document.getElementById(adSlotName);
```
nandy-andy, line 30: We don't need capital `L` here.

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+import { context, utils } from '@ad-engine/core';
+import { PrebidNativeData } from './native-models';
+
+const logGroup = 'prebid-native-provider';
+
+export class PrebidNativeProvider {
```
nandy-andy, line 7: I wonder if we should use here "Provider" word as it suggests the Provider interface is used where it is not 😊 Maybe nothing holds us from implementing the interface and making the class doing the provider's job: initialising itself and filling/rendering slots. The rest of the class responsibilities (getting/preparing the template) could be moved to `PrebidAdRenderer` and `PrebidAdTemplate` classes that would be a dependency for `PrebidNativeProvider` and would be used in the `fillIn()` method of the provider. But it's easier to write as a comment than code. We could discuss it in a wider group and maybe work on it as follow-up that we'd add to the "ads providers brainstorming" 😉

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -125,6 +126,11 @@ export class Nativo {
...
+				communicationService.emit(eventsRepository.NO_NATIVO_AD, {
```
nandy-andy, line 130: Basically we could emit such event for all nativo slots and let the provider decide if it can fill the slot or not 🤔

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+			if (data.native) {
+				this.renderPrebidNativeAd(payload.slotName, data.native);
+			}
+			communicationService.emit(eventsRepository.NO_NATIVE_PREBID_AD, {
+				slotName: payload.slotName,
+			});
+		});
+	}
+
+	private renderPrebidNativeAd(adSlotName: string, data: PrebidNativeData): void {
```
nandy-andy, line 29: I wonder why we want to keep it `private`. Making it `protected` could help us unit test the most crucial for business part -- checking if it's been called when `data.native` has been received from a bidder (through `window.pbjs.getHighestUnusedBidResponseForAdUnitCode()`). Making it public could help us unit test the input-output scenario in which we mock DOM element and check if after calling `renderPrebidNativeAd()` with proper data we got the HTML we expect.

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+			utils.logger(logGroup, 'disabled');
+			return;
+		}
+
+		communicationService.on(eventsRepository.NO_NATIVO_AD, (payload) => {
+			const data = window.pbjs.getHighestUnusedBidResponseForAdUnitCode(payload.slotName);
+			if (data.native) {
+				this.renderPrebidNativeAd(payload.slotName, data.native);
+			}
+			communicationService.emit(eventsRepository.NO_NATIVE_PREBID_AD, {
```
nandy-andy, line 23: Shouldn't we emit this only when `data.native` is undefined/in the `else` part of the condition?

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+		const assetsMap = {
+			title: 'hb_native_title',
+			body: 'hb_native_body',
+			image: 'hb_native_image',
+			icon: 'hb_native_icon',
+			clickUrl: 'hb_native_linkurl',
+		};
+
+		for (const assetName in data) {
+			if (assetsMap[assetName]) {
```
nandy-andy, line 45: In order to make the code more readable and with less indentation you could extract the if and its body to `replaceAssetPlaceholderWithValueIfExist()` -- java-like name but tells you exactly what happens in the if without scanning the code.

nandy-andy (changes requested):
Let's replace string concatenation with something more readable and try add some unit tests. In terms of the provider's future/brainstorming maybe we should fill a follow-up.

src/ad-bidders/prebid/adapters/appnexus.ts
bckatarzyna: I set 1:1 ratio to have the ad similar to Nativo one. In case of minimum size - I have no idea what would be the best in terms of our ads and prebid inventory availability. I event don't know what would be better: logo or image. Image seems to be more detailed and attractive but I'm not sure if they are available in small sizes and 1:1 ratios? This is why I created this ticket: https://fandom.atlassian.net/browse/ADEN-11849. Hope all the questions like this will find the explanation over there.

src/ad-bidders/prebid/native/prebid-native-provider.ts
bckatarzyna: Yes... we should 🙈

src/ad-engine/providers/nativo/nativo.ts
bckatarzyna: But using `.on(event)` don't we take care only of the first of the events?

src/ad-bidders/prebid/native/prebid-native-provider.ts
bckatarzyna: Follow up ticket created https://fandom.atlassian.net/browse/ADEN-11913

src/ad-engine/providers/nativo/nativo.ts
nandy-andy: By default we do but there is a flag ;) Anyway, I think this could be a future discussion included in https://fandom.atlassian.net/browse/ADEN-11913 👌

src/ad-engine/providers/nativo/nativo.ts
```diff
@@ -124,7 +125,12 @@ export class Nativo {
 	private watchNtvEvents(): void {
 		window.ntv.Events?.PubSub?.subscribe('noad', (e: NativoNoAdEvent) => {
 			const slotName = Nativo.AD_SLOT_MAP[e.data[0].id];
-			this.handleNtvNativeEvent(e, slotName, AdSlot.STATUS_COLLAPSE); // init or collapse
+			this.handleNtvNativeEvent(e, slotName, AdSlot.STATUS_COLLAPSE);
```
nandy-andy, line 128: OK, before we release it globally we'll need to add one more functionality to the Nativo module: handling collapses from PrebidNativeProvider. It may be a `watchPrebidNativeEvnets()` in which we just listen to `NO_NATIVE_PREBID_AD` and if it's emitted we execute `replaceWithAffiliateUnit()`. So, when Nativo does not return anything, PrebidNative does not fill the slot we have still chances to render the affiliate unit 🤔

nandy-andy (approved):
If we don't plan to release it globally, we can move on and share with AppNexus special links to check if it works fine on their side 👍 We should take care of the follow-ups before we release it globally, though.

PR #1441, opened 2022-06-17
ADEN-11954 Audigent changes

nandy-andy:


src/ad-services/audigent/index.ts
```diff
@@ -32,12 +32,24 @@ class Audigent {
...
+					utils.logger(logGroup, 'audience tag script loaded');
+				});
 
 			utils.scriptLoader
 				.loadScript(segmentsScriptUrl, 'text/javascript', true, 'first')
 				.then(() => {
-					this.setup();
+					utils.logger(logGroup, 'segment tag script loaded');
+
+					new utils.WaitFor(() => typeof window['au_seg'] !== 'undefined', 100, 100)
```
szymon-nawrocki, line 46: Maybe it will be better to just check if Condition is truthy or falsy, because technically value `-1` is different than `undefined` and it passes the condition which I think we don't want, so I'd change it to something like: 

`new utils.WaitFor(() => return window['au_seg'], 100, 100)`

szymon-nawrocki: Ahh sory, now I see that we don't set value of  `window['au_seg']` to -1 (I mismatched it with targeting), so never mind

jakubbanas (approved):

PR #1446, opened 2022-06-23
ADEN-11849 | Improve prebid native media types

bckatarzyna:


src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -61,11 +62,43 @@ export class PrebidNativeProvider {
 		return data[assetName];
 	}
 
+	getPrebidNativeMediaTypes(position: string): PrebidNativeMediaType {
```
nandy-andy, line 65: What do you think of a suggestion of changing name of the argument from `position` to `deviceType`? The current names if I read correctly comes from Prebid but in the AdEngine world could too easily be connected with an ad slot's name/position 🤔

src/ad-bidders/prebid/native/prebid-native-provider.ts
```diff
@@ -74,12 +107,31 @@ export class PrebidNativeProvider {
...
 							<a href="##hb_native_linkurl##">
-								<button class="ntv-ad-button">Check!</button>
+								<button class="ntv-ad-button">##hb_native_displayUrl##</button>
 							</a>
 						</div>
 					</div>
 				</div>`;
 	}
+
+	getMinImageSize(position: string): number {
```
nandy-andy, line 117: Consider moving the getters to a new class (maybe `PrebidNativeConfig`) as `PrebidNativeProvider` does a lot of things already, it:
* initializes itself adding a listener on Nativo event (or not),
* gets the highest native ad bids from Prebid,
* renders the native ad,
* parses the template,
* configures the request to Prebid.

nandy-andy (approved):

PR #1448, opened 2022-06-24
ADEN-11823 - Page Level Targeting Part I

Tomasz-at-Fandom:


dmnsobczak (changes requested):
`.DS_Store` and `spec/.DS_Store` shouldn't be committed - please remove them and add it to the `.gitignore` file :)
Apart from that: all good 👍

Tomasz-at-Fandom:
Good catch, thanks!  Totally missed that 😨

dmnsobczak (approved):

dmnsobczak (approved):

PR #1451, opened 2022-06-27
ADEN-11849 | Prebid native: change image back to icon

bckatarzyna:


@types/pbjs/index.d.ts
```diff
@@ -42,7 +42,7 @@ interface PrebidNativeMediaType {
 	displayUrl: {
 		required: boolean;
 	};
-	image: PrebidNativeImageType;
+	icon: PrebidNativeImageType;
```
szymon-nawrocki, line 45: To be consistent it would be nice to change the type name to `PrebidNativeIconType`

src/ad-bidders/prebid/native/native-models.ts
```diff
@@ -3,7 +3,7 @@ export interface PrebidNativeData {
 	body: string;
 	clickUrl: string;
 	displayUrl: string;
-	image: PrebidNativeImage;
+	icon: PrebidNativeImage;
```
szymon-nawrocki, line 6: And this interface to `PrebidNativeIcon`, @bckatarzyna what do you think?

szymon-nawrocki (approved):
👌

PR #1453, opened 2022-06-28
ADEN-11979 Introduce new page-level key-val from IAS

nandy-andy:


src/ad-services/ias-publisher-optimization/index.ts
```diff
@@ -116,32 +125,48 @@ class IasPublisherOptimization {
...
+		IasPublisherOptimization.setBrandSafetyKeyValuesInTargeting(iasTargetingData.brandSafety);
+		IasPublisherOptimization.setCustomKeyValuesInTargeting(iasTargetingData.custom);
+
+		for (const [slotName, slotTargeting] of Object.entries(iasTargetingData.slots)) {
+			context.set(`slots.${slotName}.targeting.vw`, slotTargeting.vw || slotTargeting.vw_vv);
+		}
+	}
+
+	private static setBrandSafetyKeyValuesInTargeting(brandSafetyData): void {
+		if (brandSafetyData) {
```
Tomasz-at-Fandom, line 143: How about we reverse the condition and do an early return, less indentation and we will know how this ends if brandSafetyData is not truthy on line 3 instead of the function bottom ;)

Tomasz-at-Fandom (approved):

PR #1456, opened 2022-07-04
ADEN-12005 Keep checking global var from Audigent up to 4s

nandy-andy:


spec/ad-services/audigent/audigent.spec.ts
```diff
@@ -35,6 +35,8 @@ describe('Audigent', () => {
 
 		context.set('services.audigent.enabled', undefined);
 		context.set('services.audigent.tracking.sampling', undefined);
+		context.set('services.audigent.newIntegrationEnabled', undefined);
```
dmnsobczak, line 38: Instead of setting `undefined` you can call `context.remove('services.audigent.newIntegrationEnabled')`

dmnsobczak (approved):

spec/ad-services/audigent/audigent.spec.ts
nandy-andy: Right :) I'll keep it unified but maybe next time I will remember about `context.remove()` 🤞

PR #1457, opened 2022-07-05
ADEN-11972 NTC: Make right rail sticky

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11972


platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -102,4 +108,28 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
 
 		pageElement.classList.add('companion-stick');
 	}
+
+	private expandRightRailWrapperHeight(rightRailElement: HTMLElement) {
+		rightRailElement.style.height = '100%';
```
nandy-andy, line 113: Consider creating a CSS class that adds 100% height to an element and in JS just add the class to the element. It's a good practice and seems cleaner to just use CSS classes instead of modifying DOM element's styles in JS when you don't need to calculate anything.

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -49,25 +49,25 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
 			true,
 		);
 	}
 
-	private registerStickingCompanionStickedListener(): boolean {
-		const rightRailElement = document.querySelectorAll(
-			'.main-page-tag-rcs #top_boxad, #rail-boxad-wrapper',
-		)[0] as HTMLElement;
-
+	private registerStickingCompanionStickedListener(rightRailElement: HTMLElement): boolean {
```
nandy-andy, line 70: This method has two responsibilities, it:
1. checks if there is a rightRailElement on a page,
2. registers a handler for `SLOT_STICKED_STATE` events

We should consider splitting the responsibilities into two methods.

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -86,14 +89,17 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
-	private registerStickingCompanionViewedListener(): void {
+	private registerStickingCompanionViewedListener(rightRailElement: HTMLElement): void {
 		const pageElement = document.querySelector('.page');
 
 		communicationService.onSlotEvent(
 			AdSlot.SLOT_VIEWED_EVENT,
 			() => {
 				setTimeout(() => {
 					pageElement.classList.remove('companion-stick');
+					if (rightRailElement.className.includes('right-rail-wrapper')) {
```
nandy-andy, line 100: This condition is repeated in two methods: `registerStickingCompanionStickedListener()` and `registerStickingCompanionViewedListener()`. Consider extracting it to a method for which you'd pass a callback that either reduces or expands the right rail wrapper height.

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -102,4 +108,28 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
+
+	private reduceRightRailWrapperHeight(rightRailElement: HTMLElement) {
+		const reducedHeight = this.getRightRailElementsTotalHeight();
+		if (reducedHeight > 0) {
+			rightRailElement.style.height = `${reducedHeight}px`;
+		}
+	}
+
+	private getRightRailElementsTotalHeight() {
+		const divsInsideRightRail: NodeListOf<HTMLElement> =
```
nandy-andy, line 124: Couldn't we use here already created `rightRailElement` and just gets its child instead of `document.querySelectorAll()`? Something like `rightRailElement.querySelectorAll(':scope > div')`

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
szymon-nawrocki: Great suggestion. Done ✅

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
szymon-nawrocki: Done ✅

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
szymon-nawrocki: Applying changes from your previous suggestions helped improved code, so this is not the case anymore 💪

platforms/ucp-desktop/styles.scss
```diff
@@ -94,22 +94,27 @@ $sticky-navigation-height: 46px;
...
+	.right-rail-wrapper {
 		position: sticky;
-		z-index: 100;
+		top: calc(#{$sticky-navigation-height} - 10px);
+	}
+}
+
+.companion-done {
+	.right-rail-wrapper {
+		height: 100% !important;
```
nandy-andy, line 110: !important is not a good practice and we should avoid it if possible. Why is it needed here? Without it the height is set rarely to 100% or never? 🙈

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -78,6 +78,9 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
 				if (payload.status === universalAdPackage.SLOT_STICKED_STATE) {
 					const tlbHeight = document.getElementById('top_leaderboard')?.offsetHeight || 36;
 					rightRailElement.style.top = `${tlbHeight}px`;
+					if (rightRailElement.className.includes('right-rail-wrapper')) {
```
nandy-andy, line 81: In order to make indentation smaller we could move the condition to `reduceRightRailWrapperHeight()` and maybe rename it to `reduceRightRailWrapperHeightWhenApplies()` 🤔

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -64,9 +64,9 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
 	}
 
 	private registerStickingCompanionStickedListener(): boolean {
-		const rightRailElement = document.querySelectorAll(
-			'.main-page-tag-rcs #top_boxad, #rail-boxad-wrapper',
-		)[0] as HTMLElement;
+		const rightRailElement: HTMLElement = document.querySelector(
```
nandy-andy, line 67: Unlike `registerStickingCompanionViewedListener()` the `registerStickingCompanionStickedListener()` keeps doing two things: checking if the right rail element is present and only if it is registering the event handler. Let's split it.

platforms/ucp-desktop/styles.scss
szymon-nawrocki: It is important, because it has to override `style.height` set by Javascript, without `!important` it never overrides this tyle

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -102,4 +106,26 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
...
+	private reduceRightRailWrapperHeight(rightRailElement: HTMLElement) {
+		const reducedHeight = this.getRightRailElementsTotalHeight(rightRailElement);
+		if (reducedHeight > 0) {
+			rightRailElement.style.height = `${reducedHeight}px`;
+		}
+	}
+
+	private getRightRailElementsTotalHeight(rightRailElement: HTMLElement) {
+		const rightRailCssSelector = `.${rightRailElement.className.replace(' ', '.')}`;
+		const divsInsideRightRail: NodeListOf<HTMLElement> = document.querySelectorAll(
```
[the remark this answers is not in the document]
szymon-nawrocki: Sorry, I did not know  that something like `:scope` pseudo-class exist and I thought you wrote is as a placeholder 🙈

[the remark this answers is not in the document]
szymon-nawrocki: Done ✅

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
szymon-nawrocki: Done ✅

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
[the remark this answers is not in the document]
nandy-andy: Yeah, it's a funny pseudo-class. I was surprised it existed ;) It's not supported by IE but we'll worry about it once a client of NTC asks for IE support 🤷‍♂️  Fandom [does not support IE](https://community.fandom.com/wiki/Help:Supported_browsers).

nandy-andy (approved):

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -102,4 +99,30 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
 
 		pageElement.classList.add('companion-stick');
 	}
+
+	private reduceRightRailWrapperHeightWhenApplies(rightRailElement: HTMLElement) {
+		if (
```
nandy-andy, line 104: I assume the condition here is needed in order not to make the code throw an error on main pages where the rightRailElement is actually the `#top_boxad` 🤔 How about you switch the condition from what it is currently to:
```js
if( this.isNotTopBoxad(rightRailElement) )
```
where:
```js
// ...
private isNotTopBoxad(rightRailElement: HTMLElemenet) {
  return rightRailElement.id != 'top_boxad';
}
```

nandy-andy (approved):

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
szymon-nawrocki: Actually this condition is here, because at the beginning the concept was to stick right rail on the main pages, which have class `.main-page-tag-rcs`. And this right rail did not require to reduce the height, so this is why this condition was made, to exclude right rails that did not require reducing the height from executing logic from this function.

But the concept changed during the development, and in this case indeed it would be better to just check if the right-rail element is top_boxad.
So I will take this idea into consideration when I will be sure that this is the final version of the implementation, and I don't have to stick right rail in more places :) Thanks

szymon-nawrocki:
Closing this PR because I came up with another solution:
https://github.com/Wikia/ad-engine/pull/1461

PR #1458, opened 2022-07-07
ADEN-11907 NTC on F2

preludek:


platforms/f2/setup/dynamic-slots/f2-slots-definition-repository.ts
```diff
@@ -40,6 +51,61 @@ export class F2SlotsDefinitionRepository {
...
+											universalAdPackage.SLOT_VIDEO_DONE,
+										].includes(payload.status)
+									) {
+										activateFloorAdhesion();
+									}
+								},
+								'top_leaderboard',
+							);
+						} else {
+							activateFloorAdhesion();
```
dmnsobczak, line 101: Floor adhesion should be available only for UAP, so this `else` can be removed. Then `activateFloorAdhesion()` can be refactored and include `if` with conditions.

platforms/f2/setup/dynamic-slots/f2-dynamic-slots.setup.ts
```diff
@@ -26,6 +26,7 @@ export class F2DynamicSlotsSetup implements DiProcess {
 		insertSlots([
 			topLeaderboardDefinition,
 			this.slotsDefinitionRepository.getTopBoxadConfig(),
+			this.slotsDefinitionRepository.getFloorAdhesionConfig(),
```
dmnsobczak, line 29: If FloorAdhesion on F2 is enabled only for NTC, then you can connect initialization steps with NTC-loaded event, like here:
https://github.com/Wikia/ad-engine/blob/dev/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts#L73

platforms/f2/styles.scss
```diff
@@ -51,6 +51,20 @@
...
+	body {
+		.companion-stick {
+			.article-layout__top-box-ad,
+			.feed-layout__right-rail {
+				position: sticky;
+				z-index: 100;
+			}
+
+			#top_boxad {
+				padding-top: 24px;
```
dmnsobczak, line 63: Why 24px? Can you introduce any variable to explain it and/or reuse in other places? 🤔

platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
```diff
@@ -14,6 +15,16 @@ export class BfaaF2ConfigHandler implements TemplateStateHandler {
 
 	async onEnter(): Promise<void> {
 		const enabledSlots: string[] = ['top_boxad', 'incontent_boxad', 'bottom_leaderboard'];
+		const isMobile = context.get('state.isMobile');
+
+		if (this.params.newTakeoverConfig) {
+			communicationService.emit(eventsRepository.AD_ENGINE_UAP_NTC_LOADED);
```
dmnsobczak, line 21: Did you tested this new feature with Roadblock, not only BFAA?

dmnsobczak: Should work, but just checkin' xD

platforms/f2/templates/f2-templates.setup.ts
```diff
@@ -24,12 +30,76 @@ export class F2TemplatesSetup implements DiProcess {
 		const bfab$ = registerBfabTemplate(this.registry);
 		const stickyTlb$ = registerStickyTlbTemplate(this.registry);
 		const roadblock$ = registerRoadblockTemplate(this.registry);
+		const floorAdhesion$ = registerFloorAdhesionTemplate(this.registry);
```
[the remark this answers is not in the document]
preludek: Couldn't make that work 🤷

platforms/f2/templates/handlers/bfaa/bfaa-f2-config-handler.ts
preludek: Yes, it behaves like on communities on production.

platforms/f2/styles.scss
preludek: This is not re-usable since it's the space between companions components (TLB/ TB). I've set a variable so it's descriptive now.

platforms/f2/templates/f2-templates.setup.ts
```diff
@@ -24,12 +30,76 @@ export class F2TemplatesSetup implements DiProcess {
...
 		});
+
+		communicationService.on(eventsRepository.AD_ENGINE_UAP_NTC_LOADED, () => {
+			if (!context.get('state.isMobile')) {
+				this.configureStickingCompanion();
+			}
+		});
+	}
+
+	private configureStickingCompanion(): void {
```
szymon-nawrocki, line 49: This is almost exact same code to https://github.com/Wikia/ad-engine/blob/5d2858d1c5eb9fd9119b8af011fdb3974ebe1d3d/platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts#L51
Can't it be shared between platforms, by putting it to `shared` folder?

platforms/f2/templates/handlers/floor-adhesion/floor-adhesion-bootstrap-handler.ts
```diff
@@ -0,0 +1,30 @@
...
+	AdSlot,
+	slotTweaker,
+	TEMPLATE,
+	TemplateStateHandler,
+	TemplateTransition,
+} from '@wikia/ad-engine';
+import { Inject, Injectable } from '@wikia/dependency-injection';
+
+@Injectable({ autobind: false })
+export class FloorAdhesionBootstrapHandler implements TemplateStateHandler {
```
szymon-nawrocki, line 11: This is the exact same file to https://github.com/Wikia/ad-engine/blob/2945f8e21b65566f6df92fb70ece686fcbfed4bb/platforms/ucp-desktop/templates/handlers/floor-adhesion/floor-adhesion-bootstrap-handler.ts#L11

Can it be placed somehow in `shared/templates` and shared between the platforms?

preludek: In general, like in the previous comment, it probably can. I'm not sure if it should. For now, we would have 2 similar implementations of floor adhesion (f2 and ucp-desktop), and a different one with ucp-mobile one would be shared, one wouldn't. 

Btw this it's interesting to check what would happen to bundle sizes if something is moved to shared 🤔  wonder if tree shaking works correctly there.

szymon-nawrocki (approved):
👌

PR #1462, opened 2022-07-12
ADEN-11966 - Send PPID to GAM experiment

larsire:


src/ad-engine/providers/gpt-provider.ts
```diff
@@ -210,6 +211,30 @@ export class GptProvider implements Provider {
 		});
 	}
 
+	checkSilverSurferAvailability() {
+		return (
+			window.SilverSurferSDK &&
```
preludek, line 216: This can be shortened with optional chaining

preludek (approved):
👍

PR #1465, opened 2022-07-15
ADEN-11995 Removing unused DW trackers

dmnsobczak:


szymon-nawrocki (changes requested):
There is one more place where adengine status is tracked:
https://github.com/Wikia/ad-engine/blob/d09b586c2608cb4de2bf93c8ea42253f179ae87b/platforms/shared/modes/no-ads/no-ads.mode.ts#L14

preludek:
> 

Fixed

szymon-nawrocki (approved):
👌

PR #1476, opened 2022-07-22
ADEN-11996 | Track additional load times

bckatarzyna:


src/ad-engine/services/load-times-service.ts
```diff
@@ -1,7 +1,16 @@
 import { communicationService, eventsRepository } from '@ad-engine/communication';
 import { DataWarehouseTracker } from '../../../platforms/shared';
```
dmnsobczak, line 2: This could be replaced by from `'@platforms/shared';`, but we shouldn't import from platforms to AdEngine core. This method should in AdEngine core or context should be used.

src/ad-engine/services/load-times-service.ts
```diff
@@ -42,22 +50,40 @@ export class LoadTimesService {
...
+		});
+
+		Object.keys(eventsToTrack).forEach((eventName) => {
+			communicationService.on(eventsToTrack[eventName], () => {
+				this.trackLoadTime(eventName, Date.now());
+			});
+		});
+
+		communicationService.on(eventsRepository.AD_ENGINE_SLOT_LOADED, (payload) => {
+			if (payload.name == 'top_leaderboard') {
```
dmnsobczak, line 70: FYI: be aware of `cdm-zone-01` on Sports

dmnsobczak (approved):

PR #1481, opened 2022-07-26
ADEN-12036 - Improved setting PPID process

larsire:
**Ticket**

https://fandom.atlassian.net/browse/ADEN-12036


src/ad-engine/providers/gpt-provider.ts
```diff
@@ -35,6 +38,22 @@ export function postponeExecutionUntilGptLoads(method: () => void): any {
 	};
 }
 
+function postponeExecutionUntilSilverSurferLoads(method: () => void): any {
+	return function (...args: any): void {
+		let silverSurferAvailabilityCheckCount = 0;
+		const interval = setInterval(() => {
```
jakubbanas, line 44: We have handy util for such cases:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/utils/wait-for.ts

larsire: I've refactored it to use WaitFor class as you suggested :) Thanks for pointing that!

jakubbanas (approved):

PR #1483, opened 2022-07-29
ADEN-12055 Adding Audigent assets preloading

dmnsobczak:


bckatarzyna (approved):

platforms/shared/utils/instant-config-setup.ts
```diff
@@ -34,5 +35,30 @@ export class InstantConfigSetup implements DiProcess {
...
+			const newIntegrationEnabled = instantConfig.get('icAudigentNewIntegrationEnabled');
+
+			context.set('services.audigent.gamDirectTestEnabled', gamDirectTestEnabled);
+			context.set('services.audigent.newIntegrationEnabled', newIntegrationEnabled);
+
+			if (gamDirectTestEnabled) {
+				window['au_gam_direct_test'] = true;
+			}
+
+			if (newIntegrationEnabled) {
```
bckatarzyna, line 59:
```suggestion
			if (gamDirectTestEnabled || newIntegrationEnabled) {
```

src/ad-services/audigent/index.ts
```diff
@@ -32,21 +52,17 @@ class Audigent {
...
+		if (!this.audienceTagScriptLoader) {
+			this.preloadLibraries();
+		}
 
-			utils.scriptLoader
-				.loadScript(audienceTagScriptUrl, 'text/javascript', true, 'first')
-				.then(() => {
-					utils.logger(logGroup, 'audience tag script loaded');
-				});
+		if (gamDirectTestEnabled) {
```
bckatarzyna, line 62:
```suggestion
		if (!this.isLoaded && gamDirectTestEnabled) {
```

src/ad-services/audigent/index.ts
```diff
@@ -32,21 +52,17 @@ class Audigent {
...
+			this.audienceTagScriptLoader.then(() => {
+				utils.logger(logGroup, 'audience tag script loaded');
```
bckatarzyna, line 64:
```suggestion
				utils.logger(logGroup, 'audience tag script loaded');
				this.isLoaded = true;
```

bckatarzyna (approved):

platforms/shared/utils/instant-config-setup.ts
```diff
@@ -34,5 +35,30 @@ export class InstantConfigSetup implements DiProcess {
 			context.set('bidders.prebid.libraryUrl', instantConfig.get('icPrebidVersion'));
 			pbjsFactory.init();
 		}
+
+		if (instantConfig.get('icAudigent')) {
```
mfarul, line 39: Not a very good place to load audigent - how about extracting this from instant-config-setup ?

mfarul (changes requested):

platforms/shared/utils/instant-config-setup.ts
dmnsobczak: I think that calling Audigent preload from here is fine, because it's `Preload libraries based on ICBM configuration`. I moved the most of the code though into Audigent class - you right, easier to maintain and refactor from there :)

PR #1487, opened 2022-08-02
ADEN-10791 Refactor lABlador/BAsset code

jakubbanas:


src/ad-engine/services/instant-config-cache-storage-serializer.ts
```diff
@@ -0,0 +1,38 @@
+import { CacheDictionary } from './instant-config-cache-storage';
+
+const KEY_SEPARATOR = '|';
+const VALUE_SEPARATOR = ':';
+const CACHE_LENGTH_LIMIT = 5;
+
+function mapCacheKeys(cache) {
+	return (key) => {
```
nandy-andy, line 8: What do you think of forcing `CacheData` type on the `key`?

nandy-andy (approved):

src/ad-engine/services/instant-config-cache-storage-serializer.ts
jakubbanas: Actually, the `key` is just a string, `CacheData` is what we are getting out of `cache` by this `key`.

I just typed it a little bit more explicitly.

nandy-andy: Types help 👍 I got lost writing the question 😊

PR #1491, opened 2022-08-03
ADEN-11899 allow icp on sm UAP

preludek:


platforms/ucp-desktop/templates/handlers/bfaa/bfaa-ucp-desktop-config-handler.ts
```diff
@@ -21,6 +21,10 @@ export class BfaaUcpDesktopConfigHandler implements TemplateStateHandler {
 			communicationService.emit(eventsRepository.AD_ENGINE_UAP_NTC_LOADED);
 		}
 
+		if (this.params.sequentialUapConfig) {
```
nandy-andy, line 24: The name of the parameter is misleading IMHO. The UAP usually blocks `incontent_player` slot and here we're making an exception. How about naming the parameter in the way it describes what it does: `enableIncontentPlayerSlot`? 🤔

platforms/ucp-desktop/ucp-desktop-legacy-setup.ts
```diff
@@ -23,9 +23,9 @@ export class UcpDesktopLegacySetup {
 			UcpDesktopA9ConfigSetup,
 			UcpDesktopDynamicSlotsSetup,
 			UcpDesktopSlotsStateSetup,
-			BiddersStateSetup,
 			UcpDesktopTemplatesSetup,
 			SequentialMessagingSetup, // SequentialMessagingSetup needs to be after *TemplatesSetup or UAP SM might break
+			BiddersStateSetup,
```
nandy-andy, line 28: Why is it necessary?

platforms/ucp-desktop/templates/handlers/bfaa/bfaa-ucp-desktop-config-handler.ts
preludek: I disagree with misleading. It's a custom configuration for sequential UAPs 🤷  Firstly I wanted to go a different route and set `type` to smUap or something like that but I was afraid of potential regression. Because of that I did something similar to newTakeoverConfig 🤷  Usage of this flag will potentially be extended because BLB doesn't work as desired as well.

platforms/ucp-desktop/ucp-desktop-legacy-setup.ts
preludek: I wanted to push it even higher because of a race condition with listeners in SMSetup. I think this may help a bit, but I wasn't sure about putting it higher since some of the setups may depend on each other.

nandy-andy (approved):
If it helps and does not break anything let's move forward. However, I'm not super happy with the code change but I don't have anything better to offer ATM :(

PR #1493, opened 2022-08-03
ADEN-12041 Combine site and page tags

nandy-andy:


platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+	combineSiteAndPageTags(siteTags: PageLevelTags, pageTags: PageLevelTags): PageLevelTags {
+		for (const [key, value] of Object.entries(siteTags)) {
+			if (Array.isArray(value) && value.length > 0) {
+				siteTags[key] = siteTags[key].concat(pageTags[key]);
+			} else if (Array.isArray(value) && value.length === 0) {
+				siteTags[key] = pageTags[key];
+			}
+		}
+
+		return siteTags;
```
Tomasz-at-Fandom, line 73: These aren't the `sideTags` you are looking for 🤭 We are not returning just site tags here, so maybe instead of mutating `siteTags` let's create a new variable, something like "combinedTags".

Tomasz-at-Fandom (approved):

PR #1494, opened 2022-08-05
ADEN-12075 Refactor targeting strategies

nandy-andy:


platforms/shared/context/targeting/targeting-strategies/configurators/targeting-strategies-configurator.ts
```diff
@@ -1,13 +1,41 @@
...
-		[TargetingStrategiesNames.PAGE_CONTEXT]: new PageContextStrategyBuilder().build(skin),
-		[TargetingStrategiesNames.COMBINED]: new CombinedStrategyBuilder().build(skin),
+		[TargetingStrategiesNames.DEFAULT]: new LegacyStrategy(skin),
+		[TargetingStrategiesNames.SITE_CONTEXT]: new LegacyStrategy(skin),
+		[TargetingStrategiesNames.PAGE_CONTEXT]: new PageContextStrategy(skin, context),
+		[TargetingStrategiesNames.COMBINED]: new CombinedStrategy(skin, context),
 	};
 }
+
+function validateWindowContext(windowContext: WindowContextDto): Context {
```
nandy-andy, line 21: It's actually not really validating anything, maybe we should rename it to `createTargetingContext()`? 🤔

platforms/shared/context/targeting/targeting-strategies/configurators/targeting-strategies-configurator.ts
```diff
@@ -1,13 +1,41 @@
...
+	// @ts-ignore because it does not recognize context correctly
+	const windowContext: WindowContextDto = window.fandomContext;
+	const context = validateWindowContext(windowContext);
+
 	return {
-		[TargetingStrategiesNames.DEFAULT]: new LegacySiteContextStrategyBuilder().build(skin),
-		[TargetingStrategiesNames.SITE_CONTEXT]: new LegacySiteContextStrategyBuilder().build(skin),
-		[TargetingStrategiesNames.PAGE_CONTEXT]: new PageContextStrategyBuilder().build(skin),
-		[TargetingStrategiesNames.COMBINED]: new CombinedStrategyBuilder().build(skin),
+		[TargetingStrategiesNames.DEFAULT]: new LegacyStrategy(skin),
```
nandy-andy, line 14: Do we need `targetingStrategiesConfigurator()`? Once we know what strategy is going to be used we don't need instances to all of them. Therefore, maybe the configurator isn't needed and we'll just use a factory in the `TargetingStrategyExecutor` that we'll create just one, selected instance of a strategy.

platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.ts
```diff
@@ -29,7 +29,7 @@ export class CombinedStrategy extends CommonStrategy implements TargetingStrateg
 			wpage: this.context.page.pageName && this.context.page.pageName.toLowerCase(),
 		};
 
-		const pageLevelTags: PageLevelTags = {
+		const pageLevelTags: SharedTaxonomyTags = {
```
szymon-nawrocki, line 32: The objects `targeting`, `pageLevelTags` and `siteLevelTags` are shared between strategies.
Maybe it will be worth extracting it to `common-strategy.ts`, and use particular elements for building different strategies?

platforms/shared/context/targeting/targeting-strategies/configurators/targeting-strategies-configurator.ts
szymon-nawrocki: Sounds like a great idea 👍

platforms/shared/context/targeting/targeting-strategies/interfaces/shared-taxonomy-tags.ts
```diff
@@ -1,4 +1,4 @@
-export interface PageLevelTags {
+export interface SharedTaxonomyTags {
```
szymon-nawrocki, line 1: How about naming the interface just `TaxonomyTags`? The fact that they are "shared" is clearly visible in the code

szymon-nawrocki: This caught my eye by looking at parts of the code like this:

`	combineSiteAndPageTags(
		siteTags: SharedTaxonomyTags,
		pageTags: SharedTaxonomyTags,
	): SharedTaxonomyTags {}`

platforms/shared/context/targeting/targeting-strategies/factories/create-fandom-context.ts
```diff
@@ -0,0 +1,30 @@
+import { WindowContextDto } from '../interfaces/window-context-dto';
+import { FandomContext, Page, Site } from '../models/fandom-context';
+
+export function createFandomContext() {
+	// @ts-ignore because it does not recognize context correctly
+	const windowContext: WindowContextDto = window.fandomContext;
```
szymon-nawrocki, line 6: I recommend moving fandomContext to @types->Window, so we don't have to `@ts-ignore` it:

https://github.com/Wikia/ad-engine/blob/e7c42dc016601337467146389f8fee452a4d0357/@types/window/index.d.ts#L1

platforms/shared/context/targeting/targeting-strategies/configurators/targeting-strategies-configurator.ts
szymon-nawrocki: Yes, I think the code should be moved to `createFandomContext()` function

platforms/shared/context/targeting/targeting-strategies/factories/create-selected-strategy.ts
```diff
@@ -0,0 +1,29 @@
+import { utils } from '@wikia/ad-engine';
+
+import { createFandomContext } from './create-fandom-context';
+import { TargetingStrategy } from '../interfaces/targeting-strategy';
+import { LegacyStrategy } from '../strategies/legacy-strategy';
+import { PageContextStrategy } from '../strategies/page-context-strategy';
+import { CombinedStrategy } from '../strategies/combined-strategy';
+
+export enum TargetingStrategiesNames {
```
szymon-nawrocki, line 9: I recommend using singular name for enums.

Also it would be neat to move it to the file _targeting-strategy.ts_

szymon-nawrocki: Also it would be neat to move this enum to the file _targeting-strategy.ts_ so Typescript types are organised

platforms/shared/context/targeting/targeting-strategies/factories/create-selected-strategy.ts
```diff
@@ -0,0 +1,29 @@
...
+	DEFAULT = 'default',
+	SITE_CONTEXT = 'siteContext',
+	PAGE_CONTEXT = 'pageContext',
+	COMBINED = 'combined',
+}
+
+export function createSelectedStrategy(selectedStrategy: string, skin: string): TargetingStrategy {
```
szymon-nawrocki, line 16: ❤️

spec/platforms/shared/context/targeting/targeting-strategies/strategies/page-context-strategy.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+	it('Returns empty tags when site tags are not empty and page tags are empty', function () {
+		const mockedSiteTags = {
+			gnre: ['test1', 'drama', 'comedy', 'horror'],
+			theme: ['test2', 'superheroes'],
+		};
+		const mockedContext: FandomContext = new FandomContext(
+			new Site([], true, 'ec', 'test', false, mockedSiteTags, 'lifestyle'),
+			new Page(666, 'pl', 666, 'test', 'article-test', {}),
+		);
+		const expectedTargeting = { ...defaultExpectedTargeting };
```
szymon-nawrocki, line 75: This line is redundant

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
```diff
@@ -1,7 +1,8 @@
-import { context, Targeting, utils } from '@wikia/ad-engine';
+import { context as adContext, Targeting, utils } from '@wikia/ad-engine';
```
szymon-nawrocki, line 1: @nandy-andy Could you explain why the type assertion from `context` to `adContext` was done?

szymon-nawrocki:
I think current page context strategy is missing `sex` key value in its `targeting` object, could you add it?

https://github.com/Wikia/ad-engine/blob/1c7c485ff6f06704c5c04ca9386616565bb01c14/platforms/shared/context/targeting/targeting-strategies/strategies/page-context-strategy.ts#L19

szymon-nawrocki:
By the way, the targeting in _page-context-strategy.ts_
https://github.com/Wikia/ad-engine/blob/1c7c485ff6f06704c5c04ca9386616565bb01c14/platforms/shared/context/targeting/targeting-strategies/strategies/page-context-strategy.ts#L19
and targeting in `combined-strategy.ts` 
https://github.com/Wikia/ad-engine/blob/1c7c485ff6f06704c5c04ca9386616565bb01c14/platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.ts#L19
will be exactly the same object after adding `sex` key-value.

Maybe it will be worth to extract it to _factories/create-common-targeting.ts_ ?

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
nandy-andy: I was thinking about readability. To make it easier to the reader that one `context` (from `wikia/ad-engine`) is not the same as another `Context` (from `ad-engine/platforms/shared/context/targeting/targeting-strategies/models/context`). I thought about using `legacyContext` for the `context` from `@wikia/ad-engine` but it's not really legacy, isn't it? ;) But now since I renamed `Context` to `FandomContext` we can start using `context` from `@wikia/ad-engine` and remove `as adContext` part 👌

szymon-nawrocki: Okay now I see, thanks for the explanation.
I think that renaming it back to `context` will be a good idea so it won't be misleading :)

nandy-andy:
> I think current page context strategy is missing `sex` key value in its `targeting` object, could you add it?
> 
> https://github.com/Wikia/ad-engine/blob/1c7c485ff6f06704c5c04ca9386616565bb01c14/platforms/shared/context/targeting/targeting-strategies/strategies/page-context-strategy.ts#L19

Nope, it's here: https://github.com/Wikia/ad-engine/blob/1c7c485ff6f06704c5c04ca9386616565bb01c14/platforms/shared/context/targeting/targeting-strategies/strategies/page-context-strategy.ts#L25

platforms/shared/context/targeting/targeting-strategies/interfaces/targeting-strategy.ts
```diff
@@ -1,5 +1,12 @@
 import { Targeting } from '@wikia/ad-engine';
 
-export interface TargetingStrategy {
+export enum TargetingStrategy {
```
rafalgiemza, line 3: what do you think about this approach:
https://frontlive.pl/blog/typescript-enum
?

rafalgiemza (approved):
looks good to me 👍

szymon-nawrocki (approved):
👌

PR #1495, opened 2022-08-09
ADEN-12073 Finish off new integration with Audigent

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12073


src/ad-services/audigent/index.ts
```diff
@@ -28,46 +28,26 @@ class Audigent {
 	}
 
 	init(instantConfig: InstantConfigService): void {
-		context.set(
```
szymon-nawrocki, line 31: Since we don't use this variable anymore I removed this dependency, @nandy-andy is that ok?

nandy-andy: Uhm, sure. We weren't asked to change this asset. And removing it it'll be one variable less for Audigent in the ICBM 🤔

nandy-andy (approved):

PR #1496, opened 2022-08-09
ADEN-12023 Implementing sticking TopBoxad experiment

dmnsobczak:


platforms/shared/dynamic-slots/sticked-boxad-helper.ts
```diff
@@ -0,0 +1,91 @@
...
+	AdSlot,
+	communicationService,
+	context,
+	eventsRepository,
+	universalAdPackage,
+} from '@wikia/ad-engine';
+
+interface StickedBoxadHelperConfiguration {
+	slotName: string;
+	pusherSlotName: string;
```
preludek, line 12: Why is this necessary?

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
```diff
@@ -0,0 +1,91 @@
...
+import {
+	AdSlot,
+	communicationService,
+	context,
+	eventsRepository,
+	universalAdPackage,
+} from '@wikia/ad-engine';
+
+interface StickedBoxadHelperConfiguration {
+	slotName: string;
```
preludek, line 11: That's like always top_boxad right? :D

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
```diff
@@ -43,64 +39,11 @@ export class UcpDesktopTemplatesSetup implements DiProcess {
 		templateService.register(PorvataTemplate, getOutstreamConfig());
 		templateService.register(SafeFanTakeoverElement);
 
-		communicationService.on(eventsRepository.AD_ENGINE_UAP_NTC_LOADED, () => {
-			this.configureStickingCompanion();
+		this.stickedBoxadHelper.initialize({
+			slotName: 'top_boxad',
+			pusherSlotName: 'top_leaderboard',
+			pageSelector: '.page',
+			railSelector: '.main-page-tag-rcs #top_boxad, #rail-boxad-wrapper',
```
preludek, line 46: Maybe it's time (or maybe not :P) to think about setting this and similar selectors as constants somewhere in platforms?

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	private registerViewedListener(): void {
+		const pageElement = document.querySelector(this.configuration.pageSelector);
+
+		communicationService.onSlotEvent(
+			AdSlot.SLOT_VIEWED_EVENT,
+			() => {
+				setTimeout(() => {
+					pageElement.classList.add('companion-viewed');
```
preludek, line 81: This may break AdEngine if something goes wrong in line 75.

preludek (changes requested):

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
dmnsobczak: I don't like hardcoded slot names in the class located in `shared` directory. Also I want this class to be expanded on mobile while implementing sticky incontents there - configuration will handle different logic there 🤔

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
dmnsobczak: Yes, but same reason - instead of hardcoding `top_boxad` 2 or 3 times I prefer to store it in the variable and use when needed :D

platforms/ucp-desktop/templates/ucp-desktop-templates.setup.ts
dmnsobczak: Layout constants are used in the Templates 2.0 (for example in UAP) and I think that it is a bit overkill to create a template handler for this sticking functionality. It will also require some changes allowing two templates to work at the same time in case of UAP. So maybe in the future :D

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
dmnsobczak: Additional check added! 🙇

platforms/shared/dynamic-slots/sticked-boxad-helper.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const ADDITIONAL_STICK_TIME = 500;
+const PADDING_TOP = 36;
+
+@Injectable()
+export class StickedBoxadHelper {
+	private configuration: StickedBoxadHelperConfiguration;
+	private railElement: HTMLElement;
+
+	initialize(params: StickedBoxadHelperConfiguration): void {
```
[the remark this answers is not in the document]
dmnsobczak: Splitted!

preludek (approved):

PR #1502, opened 2022-08-22
ADEN-11884 Setting COPPA flag for GAM and Prebid

dmnsobczak:


platforms/shared/setup/base-context.setup.ts
```diff
@@ -121,6 +121,8 @@ export class BaseContextSetup implements DiProcess {
 			this.instantConfig.get('icComscoreJwpTracking'),
 		);
 		context.set('options.adsInitializeV2', this.instantConfig.get('icAdsInitializeV2'));
+		context.set('options.coppaGam', this.instantConfig.get('icCoppaGam'));
+		context.set('options.coppaPrebid', this.instantConfig.get('icCoppaPrebid'));
```
nandy-andy, line 125: What do you think of renaming the variables, so they describe what they stand for more? Something like: `icEnableCoppaGamFlag` and `icEnableCoppaPrebidFlag`?

src/ad-bidders/prebid/index.ts
```diff
@@ -95,6 +95,10 @@ export class PrebidProvider extends BidderProvider {
 			},
 		};
 
+		if (context.get('options.coppaPrebid') && context.get('targeting.kid_wiki') === '1') {
```
nandy-andy, line 98: We usually check if the kid_wiki targeting is set 🤔 Maybe it's worth extracting this part of the condition to some reusable method and then use it in all the places:
* https://github.com/Wikia/ad-engine/blob/6a0e10631bb16f41f98476ee3ce485e27ef4ea3a/src/ad-bidders/prebid/live-ramp/live-ramp.ts#L42
* https://github.com/Wikia/ad-engine/blob/434a02e430a0d2a29451c9f134e26159cf90cdd9/src/ad-services/facebook-pixel/index.ts#L13
* https://github.com/Wikia/ad-engine/blob/6a0e10631bb16f41f98476ee3ce485e27ef4ea3a/src/ad-bidders/prebid/ats/ats.ts#L42
* https://github.com/Wikia/ad-engine/blob/434a02e430a0d2a29451c9f134e26159cf90cdd9/src/ad-bidders/prebid/identity-hub/identity-hub.ts#L40
* https://github.com/Wikia/ad-engine/blob/434a02e430a0d2a29451c9f134e26159cf90cdd9/src/ad-services/live-connect/index.ts#L17
* https://github.com/Wikia/ad-engine/blob/434a02e430a0d2a29451c9f134e26159cf90cdd9/src/ad-services/ias-publisher-optimization/index.ts#L51
* https://github.com/Wikia/ad-engine/blob/71950bca33bacd1d008bf95731a65fd948e62310/src/ad-services/audigent/index.ts#L27

src/ad-engine/providers/gpt-provider.ts
```diff
@@ -205,10 +205,15 @@ export class GptProvider implements Provider {
...
 	setupRestrictDataProcessing(): void {
 		const tag = window.googletag.pubads();
-
-		tag.setPrivacySettings({
+		const settings: Dictionary = {
 			restrictDataProcessing: trackingOptIn.isOptOutSale(),
-		});
+		};
+
+		if (context.get('options.coppaGam') && context.get('targeting.kid_wiki') === '1') {
```
nandy-andy, line 212: `context.get('targeting.kid_wiki') === '1'` is used here for the second time within this PR - enough to make it a reusable method 🤔

src/ad-engine/utils/tagless-request-url-builder.ts
```diff
@@ -91,6 +91,11 @@ export function buildVastUrl(
 	if (ppid) {
 		params.push(`ppid=${ppid}`);
 	}
+
+	if (context.get('options.coppaGam') && context.get('targeting.kid_wiki') === '1') {
```
nandy-andy, line 95: The whole condition is repeated - worth making a reusable method :)

src/ad-engine/utils/tagless-request-url-builder.ts
```diff
@@ -126,12 +131,18 @@ export function buildVastUrl(
...
 	const params: string[] = [`c=${correlator}`, 'tile=1', 'd_imp=1'];
+
 	params.push(`iu=${options.adUnit}`);
 	params.push(`sz=${options.size}`);
 
 	if (ppid) {
 		params.push(`ppid=${ppid}`);
 	}
+
+	if (context.get('options.coppaGam') && context.get('targeting.kid_wiki') === '1') {
```
nandy-andy, line 142: Ditto

nandy-andy (dismissed):
We can remove some repeated code by extracting it to reusable method. Also, consider unit testing it since we've already have unit tests where we mock our god-object: context ;)

platforms/shared/setup/base-context.setup.ts
dmnsobczak: These are rollout variables on AdEng Config Editor access level. Oct 3rd is a deadline for full rollout after which they're going to be removed. I don't see a reason to make them more descriptive and not worth removing old ones to create new in ICBM (since I already did that) 🤔

src/ad-bidders/prebid/index.ts
dmnsobczak: Refactored

dmnsobczak: But all those occurrences I would do in the follow up ticket for Targeting refactor to not expand the scope of this ticket even more. Every occurence require changes in unit tests and additional QA. Let's do it in the next step.

src/ad-engine/providers/gpt-provider.ts
dmnsobczak: Done!

src/ad-engine/utils/tagless-request-url-builder.ts
dmnsobczak: Done

src/ad-engine/utils/tagless-request-url-builder.ts
dmnsobczak: Done

platforms/shared/setup/base-context.setup.ts
nandy-andy: I respectfully disagree. We need our code to be clean and understandable. What if there will be no Damian and nAndy after Oct 3rd? Or no AdEng? 🙀

src/ad-engine/services/config-service.ts
```diff
@@ -0,0 +1,18 @@
+import { context } from './context-service';
+
+/*
+ *  ToDo: Development improvement refactor
+ *  This class is about to be expanded in ADEN-10310
+ */
+class Config {
```
nandy-andy, line 7: Does it have to include "service" in the name? We can discuss it while refining ADEN-10310 👌

nandy-andy (approved):
The conditions look much more cleaner. A pity we didn't improve test coverage but the config variables will be removed, so maybe it'll be fine 🤞

src/ad-engine/services/config-service.ts
dmnsobczak: The same rule applies to existing `Context` class and since those 3 services (context, targeting and config) will work together I kept that rule, but we should definitely discuss it :D

platforms/shared/setup/base-context.setup.ts
dmnsobczak: But there will be Jira ticket and comments 😈 We should prioritize the ticket adding ICBM descriptions in ICBM Dashboard to allow descriptions more detailed than one additional work in icVar name 🤔 
https://fandom.atlassian.net/browse/ADEN-10883

dmnsobczak:
Yeah, this code will be removed soon and also `Config` and `Targeting` classes are proposals / RFC, so we should discuss it with the team before investing more in this module 🤔

PR #1508, opened 2022-08-26
ADEN-12121 add other LiveConnect id types

jakubbanas:


spec/ad-services/live-connect/live-connect.spec.ts
```diff
@@ -58,4 +59,42 @@ describe('LiveConnect', () => {
 
 		expect(loadScriptStub.called).to.equal(false);
 	});
+
+	describe('LiveConnect ids resolution', () => {
+		let emitSpy: SinonSpy;
+
+		beforeEach(() => {
+			emitSpy = sandbox.spy(communicationService, 'emit');
+		});
```
szymon-nawrocki, line 68: I think it is a good practise too add
`afterEach(function () {
        sandbox.restore();
    });`
To avoid any side effects in the future

jakubbanas: Good point, done!

src/ad-services/live-connect/index.ts
```diff
@@ -1,11 +1,27 @@
...
+	params?: LiQParams;
+}
+
 const partnerName = 'liveConnect';
 const partnerNameUnifiedId = 'liveConnect-unifiedId';
 const logGroup = partnerName;
 const liveConnectScriptUrl = 'https://b-code.liadm.com/a-07ev.min.js';
 
+const idConfigMapping: IdConfig[] = [
+	{ id: 'unifiedId', name: partnerNameUnifiedId },
```
szymon-nawrocki, line 16: To keep the pattern between the objects we can remove `partnerNameUnifiedId` and make a name value like the other ones, so: `name: `${partnerName}-unifiedId`

spec/ad-services/live-connect/live-connect.spec.ts
```diff
@@ -58,4 +59,44 @@ describe('LiveConnect', () => {
...
+
+		afterEach(() => sandbox.restore());
+
+		it('resolve valid id', async () => {
```
szymon-nawrocki, line 72: To make tests more readable I'd recommend modifying test case names a little, so it will look something like:

LiveConnect ids resolution
    md5 is resolved
    unifiedId is resolved
    wrong or no id is not resolved

@jakubbanas  what do you think?

jakubbanas: Right, it would be more consistent 👍

szymon-nawrocki (approved):
👌

PR #1509, opened 2022-08-29
ADEN-12097 Refactored BaseServiceSetup into Decorator

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12097)


platforms/shared/setup/ad-engine.setup.ts
```diff
@@ -0,0 +1,15 @@
+import { AdEngine, ServiceStage, Service } from '@wikia/ad-engine';
+
+let adEngineInstance: AdEngine;
```
szymon-nawrocki, line 3: Why this variable is not encapsulated inside a class?

szymon-nawrocki (approved):

platforms/shared/setup/ad-engine.setup.ts
domjanik: Good point, thanks for noticing it!

dmnsobczak (changes requested):
Everything is fine, but please reset `package-lock.json`, because it was completely overwritten. You should use `npm ci` while setting up local AdEngine instance and only update the lock with required packages - though I don't see anything added in this PR 🤔

dmnsobczak (approved):

dmnsobczak (approved):

PR #1510, opened 2022-08-30
ADEN-11910 [WIP] Adding Blockthrough side units placements

dmnsobczak:


preludek (approved):
Could use some sort of an explanation what some of those numbers mean.

dmnsobczak:
> Could use some sort of an explanation what some of those numbers mean.

Sure: I added more variables to better explain the math used there.

PR #1515, opened 2022-09-01
ADEN-12035 ICBM fallback cleanup + backend bundles

dmnsobczak:


preludek (approved):

@types/ads/index.d.ts
```diff
@@ -56,6 +56,7 @@ interface MediaWikiFeaturedVideoInfo {
 
 interface AdTagManagerTags {
 	age?: string[];
+	bundles?: string[];
```
nandy-andy, line 59: `bundles` are not coming from AdTagManager 🤔

src/ad-engine/utils/targeting.ts
```diff
@@ -62,6 +62,13 @@ class Targeting {
 			bundles.push('interstitial_disabled');
 		}
 
+		if (
+			context.get('targeting.skin').includes('ucp_') &&
```
nandy-andy, line 66: What if we add new skins or change the targeting? Consider moving this logic outside of the core and place it within platforms code 🤔

src/ad-engine/utils/targeting.ts
```diff
@@ -32,7 +32,7 @@ class Targeting {
 	}
 
 	getTargetingBundles(bundles: Dictionary<Dictionary<string[]>>): string[] {
-		let targetingBundles = [];
+		const targetingBundles = context.get('targeting.bundles') || [];
```
nandy-andy, line 35: It looks weird that the getter needs a parameter. And yet, I have a controversial suggestion pass the bundles from the context as a parameter in order to make the method more testable and less error-prone.

Now, what do you think of changing the getter to a setup function and call it within a platform setup?

PR #1518, opened 2022-09-01
ADEN-12123 Clean-up page-level key-vals we send to GAM

nandy-andy:
In order to make space for new key-vals (from automated page-level tagging) we want to minimise the risk of too many key-vals in an ad call and remove some unnecessary ones like the `cortex-visitor=yes|no`, `galactus_status=on_time|too_late` and bunch of `g_*` ones.

The code changes below does exactly what's described above by removing not needed files and their references.


szymon-nawrocki (approved):
👌

PR #1523, opened 2022-09-05
ADEN-12108 Removing AppNexusGroupM, Beachfront and MediaGrid bidders

dmnsobczak:


src/ad-bidders/prebid/index.ts
```diff
@@ -158,33 +156,6 @@ export class PrebidProvider extends BidderProvider {
 		return {};
 	}
 
-	private configureJWPlayerDataProvider(): object {
```
nandy-andy, line 161: Interesting 🤔 We're removing it because only MediaGrid was using it but maybe it's worth keeping and sharing with other bidding partners. Worth to follow-up on it with our business partners :)

nandy-andy (approved):

src/ad-bidders/prebid/index.ts
dmnsobczak: Yep. This might be explored here, because JWP Data Provider was a part of ORTB FPD :)
https://fandom.atlassian.net/browse/ADEN-12067

PR #1530, opened 2022-09-07
ADEN-12002 | Send real ad sizes to DW

bckatarzyna:


package.json
```diff
@@ -53,6 +53,7 @@
 		"nyc": "15.1.0",
 		"prettier": "2.6.2",
 		"process": "0.11.10",
+		"resize-observer": "1.0.4",
```
nandy-andy, line 56: I wonder if we need it 🤔 I assume it provides polyfills for browsers that don't support [ResizeObserver](https://caniuse.com/?search=ResizeObserver). However, these are [browsers we don't support](https://community.fandom.com/wiki/Help:Supported_browsers) and adding dependency generates maintenance cost and makes our bundle bigger. How about starting without it and checking how much ads sizes we're missing because of pageviews on nut supported browsers? :)

src/ad-engine/models/ad-slot.ts
```diff
@@ -646,4 +648,28 @@ export class AdSlot {
...
+		const adFrame = adSlotElement
+			.querySelectorAll('[id^=google_ads_iframe_]')[0]
+			.querySelectorAll('[id^=google_ads_iframe_]')[0];
+
+		if (adType.includes('success')) {
+			const resizeObserver = new ResizeObserver((entries) => {
+				for (const entry of entries) {
+					const width = Math.floor(entry.target.clientWidth);
+					const height = Math.floor(entry.target.clientHeight);
+					if (width && height) {
```
nandy-andy, line 663: Do we need it? I think all elements passed in the observable entries have the `clientWidth` and `clientHeight` 🤔

nandy-andy: Right, you've mentioned at our meetings about it. Hopefully switching to built-in in browsers `ResizeObserver` will help 🤞 If not, I'd like to suggest making it more obvious that we want to skip 0x0 size with:
```
if (width > 0 && height > 0)
```
This will make it clear as current, shorter version could be read in a way that we expect some elements not to have width and height defined 🤔

nandy-andy: Yeah, it seems that built-in `ResizeObserver` understands `100%` very well :D

package.json
nandy-andy: It works for me ;)

src/ad-engine/models/ad-slot.ts
```diff
@@ -646,4 +647,28 @@ export class AdSlot {
 	getSlotsToInjectAfterRendered(): string[] {
 		return context.get(`events.pushAfterRendered.${this.getSlotName()}`) || [];
 	}
+
+	setupSizesTracking(event: googletag.events.SlotRenderEndedEvent, adType: string): void {
+		const adSlotElement = document.getElementById(event.slot.getSlotElementId());
+		const adFrame = adSlotElement
+			.querySelectorAll('[id^=google_ads_iframe_]')[0]
```
mfarul, line 654: is it always safe to get first element from the table?

nandy-andy: Good point! I thought it's safe since GAM always returns the iframe but it may not work with native and the collapse messages - worth checking 👌

bckatarzyna: This function is fired by rendered event from GPT. So
- native slots (or widgets): don't go through GPT so nothing is fired
- collapses: don't render so nothing is fired
- collapsinator (`?cid=adeng-collapse`): renders empty slots so they have the same iframe as regular ones (+ nothing is sent to DW as these slots don't change their size - so it's also expected behaviour)
Nothing else comes to my mind but maybe @kwfandom will come up with some more ideas while testing.

The other thing is that this querying could fail due to e.g. changes introduced by Google. But I also talked with @kwfandom and he is going to write some automated tests so we will know when it happens.

Is that ok @mfarul?

mfarul: let's add a helper method that will check the nullability size, a constant for the selector would also be nice

mfarul (approved):

PR #1532, opened 2022-09-12
ADEN-12194 Refactor targeting strategies

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12194


@types/fandom/index.d.ts
```diff
@@ -6,6 +6,8 @@ interface WindowFandomContext {
 		siteName: string;
 		top1000: boolean | null;
 		tags: any;
+		legacyVertical: string;
```
mfarul, line 9: let's move this to fields into a one called taxonomy, that will contain string[] table of a full taxonomy path,f.e. ['life']['lifestyle']['tv']['zombies']['the walking dead']. Then we could just use s0v=taxonomy[0], s0=taxonomy[1] if they are available.
Thus we will be open for all sites that will have richer / poorer taxonomy.

platforms/shared/context/targeting/targeting-strategies/interfaces/taxonomy-tags.ts
```diff
@@ -1,8 +1,14 @@
-export interface TaxonomyTags {
+export interface PrefixableTaxonomyTags {
```
mfarul, line 1: Why call them prefixable? This is a site effect of some strategies not the type per se. They are rather strategy resolved depended tags, instead of the 3 ones down that are always get from the same place.

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
```diff
@@ -53,16 +74,49 @@ export abstract class CommonStrategy {
...
+			media: contextTags?.media || [],
+			pform: contextTags?.pform || [],
+			pub: contextTags?.pub || [],
+			sex: contextTags?.sex || [],
+			theme: contextTags?.theme || [],
+			tv: contextTags?.tv || [],
+		};
+	}
+
+	protected addPagePrefixToValues(tags: TaxonomyTags): PrefixableTaxonomyTags {
```
mfarul, line 91: why this is a part of common? I would extract this into separate class that will add prefixes and combinations

platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy-site-tags-based.ts
```diff
@@ -0,0 +1,32 @@
+import { context, Targeting, utils } from '@wikia/ad-engine';
+import { CommonStrategy } from './common-strategy';
+import { TaxonomyTags } from '../interfaces/taxonomy-tags';
+import { TargetingStrategyInterface } from '../interfaces/targeting-strategy';
+import { FandomContext } from '../models/fandom-context';
+
+export class CombinedStrategySiteTagsBased
```
mfarul, line 7: is it possible to build this based on SiteContextStrategy class? Or even better - could we use delegation strategy to build combined strategy from any other one?

mfarul (changes requested):

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
```diff
@@ -1,31 +1,52 @@
...
+			uap_c: 'none',
+			is_mobile: utils.client.isMobileSkin(skin) ? '1' : '0',
+		};
+
+		const commonContextParams: CommonStrategyParams = {
+			artid: fandomContext.page.articleId ? fandomContext.page.articleId.toString() : '',
+			esrb: fandomContext.site.esrbRating || '',
+			kid_wiki: fandomContext.site.directedAtChildren ? '1' : '0',
+			lang: fandomContext.page.lang || 'unknown',
+			s0: fandomContext.site.legacyVertical || fandomContext.site.vertical,
```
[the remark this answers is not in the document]
szymon-nawrocki: * `legacyVertical` is needed to set targeting `s0` value
* `wikiVertical` is needed to set `s0v` value

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
```diff
@@ -1,31 +1,52 @@
...
+			s0c: fandomContext.site.categories,
+			s0v: fandomContext.site.wikiVertical || wiki.targeting.wikiVertical,
```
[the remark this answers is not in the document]
szymon-nawrocki: I've added the information about it in the release notes, but good point that it should be also added here 👍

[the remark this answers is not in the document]
szymon-nawrocki: done ✅

platforms/shared/context/targeting/targeting-strategies/interfaces/taxonomy-tags.ts
szymon-nawrocki: I've divided the interface into two, to make it more visible, to which tags we are adding `p_` prefix.
But on the other hand it is visible in place where we are adding prefixes here:
https://github.com/Wikia/ad-engine/blob/933ad5c8ea057e9b01fe25657b7ba608fb6e51b9/platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts#L11

So now I see that indeed it is not needed here, and probably makes things less readable.

platforms/shared/context/targeting/targeting-strategies/strategies/common-strategy.ts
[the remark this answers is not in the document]
szymon-nawrocki: I moved it to `taxonomy` variable ✅

@types/fandom/index.d.ts
szymon-nawrocki: Done in:
https://github.com/Wikia/unified-platform/pull/11246

platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy-site-tags-based.ts
szymon-nawrocki: Probably it is possible and it should be done in this way to make the code more clean.
However I decided not doing this in this PR. but revisit it after the Sprint work will be done

platforms/shared/context/targeting/targeting-strategies/utils/tag-combinator.ts
```diff
@@ -0,0 +1,31 @@
...
+		return tags;
+	}
+
+	combinePrefixableTags(siteTags: TaxonomyTags, pageTags: TaxonomyTags): TaxonomyTags {
+		const combinedTags: TaxonomyTags = {};
+
+		for (const [key, value] of Object.entries(siteTags)) {
+			if (this.prefixableKeys.includes(key) && Array.isArray(value)) {
+				if (value.length > 0) {
+					combinedTags[key] = siteTags[key].concat(pageTags[key]);
```
mfarul, line 22: does this make any sense? e.g. if we are going through prefixable tags and sending them with p_ why we are combining the values?

domjanik (approved):

src/platforms/shared/context/targeting/targeting-strategies/strategies/common-tags.ts
```diff
@@ -0,0 +1,123 @@
...
+			original_host: wiki.opts?.isGamepedia ? 'gamepedia' : 'fandom',
+			skin: this.skin,
+			uap: 'none',
+			uap_c: 'none',
+			is_mobile: utils.client.isMobileSkin(this.skin) ? '1' : '0',
+		};
+
+		const commonContextParams: CommonStrategyParams = {
+			artid: this.fandomContext.page.articleId ? this.fandomContext.page.articleId.toString() : '',
+			esrb: this.fandomContext.site.esrbRating || '',
```
domjanik, line 34: I think that esrbRating should be moved to the optionalKeyVals as well - currently it has been moved to that section and it was done after last changes to ESRB Rating

domjanik (changes requested):

src/platforms/shared/context/targeting/targeting-strategies/strategies/common-tags.ts
szymon-nawrocki: Great suggestion! Now instead of empty string it will be 1 key val less and shorter URL

szymon-nawrocki: Done ✅

domjanik (approved):

src/platforms/shared/context/targeting/targeting-strategies/decorators/summary-decorator.ts
```diff
@@ -0,0 +1,14 @@
+import { TargetingStrategyInterface } from '../interfaces/targeting-strategy';
+import { Targeting } from '@wikia/ad-engine';
+import { CommonTags } from '../strategies/common-tags';
+
+export class SummaryDecorator implements TargetingStrategyInterface {
+	constructor(private commonTags: CommonTags, private strategyToSum: TargetingStrategyInterface) {}
```
mfarul, line 6: This could sum any of the TargetingStrategyInterface and not only a one with CommonTags?
If this is possible it should be:
`constructor(private tagsToSum: TargetingStrategyInterface[]) {}`

src/platforms/shared/context/targeting/targeting-strategies/decorators/combine-tags-decorator.ts
```diff
@@ -0,0 +1,35 @@
+import { TargetingStrategyInterface } from '../interfaces/targeting-strategy';
+import { Targeting } from '@wikia/ad-engine';
+
+export class CombineTagsDecorator implements TargetingStrategyInterface {
```
mfarul, line 4: looking at this now, you were right, it is not the Decorator rather Composer, let's rename this one into TagsByKeyComposer and the package to structures

src/platforms/shared/context/targeting/targeting-strategies/decorators/summary-decorator.ts
```diff
@@ -0,0 +1,14 @@
+import { TargetingStrategyInterface } from '../interfaces/targeting-strategy';
+import { Targeting } from '@wikia/ad-engine';
+import { CommonTags } from '../strategies/common-tags';
+
+export class SummaryDecorator implements TargetingStrategyInterface {
```
mfarul, line 5: let's call it TagsPlainSumBuilder

mfarul (changes requested):

src/platforms/shared/context/targeting/targeting-strategies/decorators/summary-decorator.ts
szymon-nawrocki: You're right, so it will enable easy adding of new set of tags in the future

szymon-nawrocki: Done ✅

src/platforms/shared/context/targeting/targeting-strategies/decorators/combine-tags-decorator.ts
szymon-nawrocki: Great suggestion! Done ✅

src/platforms/shared/context/targeting/targeting-strategies/decorators/summary-decorator.ts
szymon-nawrocki: Done ✅

spec/platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.spec.ts
```diff
@@ -20,81 +25,127 @@ describe('CombinedStrategy execution', () => {
...
+			gnre: [],
+			media: [],
+			pform: [],
+			pub: [],
+			sex: [],
+			theme: [],
+			tv: [],
+		};
+
+		const combinedStrategy = new TagsPlainSumBuilder([
```
mfarul, line 68: move the creation code to one method or just use Factory from create-selected-strategy.st

spec/platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.spec.ts
```diff
@@ -20,81 +25,127 @@ describe('CombinedStrategy execution', () => {
...
 		hostpre: '',
 		is_mobile: '0',
 		kid_wiki: '1',
 		lang: 'pl',
-		media: [],
 		original_host: 'fandom',
-		pform: [],
-		pub: [],
-		s0: 'lifestyle',
+		s0: 'life',
```
mfarul, line 39: move this to constant and use here and in #52

spec/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+			opts: {},
+			targeting: {},
+		});
+	});
+
+	afterEach(() => {
+		context.set('geo.country', undefined);
+	});
+
+	it('commonTags are returned correctly', function () {
```
mfarul, line 23: here the taxonomy[] -> s0, s0v is hidden it could be an explicit new test

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		const commonContextParams: CommonTargetingParams = {
+			artid: this.fandomContext.page.articleId ? this.fandomContext.page.articleId.toString() : '',
+			kid_wiki: this.fandomContext.site.directedAtChildren ? '1' : '0',
+			lang: this.fandomContext.page.lang || 'unknown',
+			// 'this.fandomContext.site.vertical' should be removed after UCP release from ADEN-12194
+			s0: this.fandomContext.site.taxonomy?.[0] || this.fandomContext.site.vertical,
+			s0c: this.fandomContext.site.categories,
+			// 'wiki.targeting.wikiVertical' should be removed after UCP release from ADEN-12194
+			s0v: this.fandomContext.site.taxonomy?.[1] || wiki.targeting.wikiVertical,
```
mfarul, line 40: [1] will not throw if we have 1 element array?

spec/platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.spec.ts
szymon-nawrocki: done ✅

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
szymon-nawrocki: no, it will return `undefined`

spec/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.spec.ts
szymon-nawrocki: I extracted taxonomy values into `const` and now testing this functionality is included in the main test

szymon-nawrocki: But I added the tests for the case: `'s0 and s0v get assigned empty string when taxonomy value is empty array'`

mfarul (approved):

domjanik (approved):

PR #1533, opened 2022-09-13
ADEN-12179 Refactored UserIdentity, Added UserIdentity to Sports platform

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12179)


src/ad-engine/utils/storage.ts
```diff
@@ -0,0 +1,18 @@
+class Storage {
```
mfarul, line 1: what Storage ? Take a look also on StorageProvider and UniversalStorage class ;)

src/ad-engine/utils/uuid.ts
```diff
@@ -0,0 +1,9 @@
+import { v4 } from 'uuid';
+
+class Uuid {
```
mfarul, line 3: do we need this encapsulation ? can we just use uuid() function?

src/ad-services/user-identity/adms-strategy/adms-client.ts
```diff
@@ -0,0 +1,62 @@
+import { ActionType, Action, ActiveData } from './adms-actions';
+import { utils } from '@ad-engine/core';
+import { UserIdentity } from '../index';
+
+export function getServicesBaseURL() {
+	return window.location.hostname.includes('fandom.com')
```
mfarul, line 6: will it work on futhead.com  or muthead.com ?

src/ad-services/user-identity/adms-strategy/adms-client.ts
```diff
@@ -0,0 +1,62 @@
...
+		: 'https://services.fandom-dev.' +
+				(location.hostname.match(/(?!\.)(pl|us)$/) || ['us'])[0] +
+				'/';
+}
+
+class AdmsClient {
+	private ADMS = getServicesBaseURL() + 'active-data-management-service/';
+
+	fetchData(): Promise<ActiveData> {
+		return fetch(this.ADMS + `profiles?cb=${Date.now()}`, {
```
mfarul, line 17: what are the cookies required for this call on futhead.com? beacon's ?

src/ad-services/user-identity/adms-strategy/adms-service.ts
```diff
@@ -0,0 +1,26 @@
+import { admsClient } from './adms-client';
+import { ActionType, ActiveData } from './adms-actions';
+
+class AdmsService {
```
mfarul, line 4: for future: this could a generic functionality that will combine local and backend storages

src/ad-services/user-identity/adms-strategy/index.ts
```diff
@@ -0,0 +1,40 @@
+import { admsService } from './adms-service';
+import { ActionType, IdentityAction } from './adms-actions';
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+import { StorageStrategyInterface } from '../storage-strategies';
+import { utils } from '@ad-engine/core';
+import { UserIdentity } from '../';
+
+class AdmsStrategy implements StorageStrategyInterface {
```
mfarul, line 8: AdmsIdentityStorage?

src/ad-services/user-identity/index.ts
```diff
@@ -1,18 +1,36 @@
...
+			utils.logger(UserIdentity.logGroup, 'Setting up PPID has failed!', e);
 		}
 	}
 
 	async call(): Promise<void> {
-		if (context.get('services.ppid.enabled')) {
-			await this.setupPPID();
+		const ppid = context.get('services.ppid');
+		if (ppid?.enabled) {
+			await this.setupPPID(ppid.strategy);
```
mfarul, line 31: where do we set ppid.strategy?

mfarul (changes requested):

src/ad-engine/utils/uuid.ts
domjanik: It's necessary for sake of our tests - There are pretty serious problems with unit tests when you use that library and "cleanest" solution is to use additional layer, that can be mocked in tests

src/ad-services/user-identity/adms-strategy/adms-client.ts
domjanik: We need FandomGlobalID, however if we won't provide it - it'll work anyway, create proper object and save it in LocalStorage but won't sync it with backend.

src/ad-services/user-identity/index.ts
domjanik: It has been moved to separate feature flag. I was thinking about changing existing one

src/ad-services/user-identity/adms-strategy/adms-service.ts
domjanik: If we will need it somewhere else I think it won't be a problem to move it to some other place. Right now I think we can leave it there :)

spec/ad-services/user-identity/user-identity.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+	before(() => {
+		v4Stub = stub(uuid, 'v4');
+		v4Stub.returns(mockId);
+	});
+
+	beforeEach(() => {
+		sandbox = createSandbox();
+	});
+	afterEach(() => {
+		sandbox.restore();
```
nandy-andy, line 25: I think it's a good practice of ours to also reset our god-object config here with something like:
```js
context.set('services.ppid.enabled', undefined);
```
or:
```js
context.remove('services.ppid.enabled');
```
It helps not polluting the god object in all tests run.

src/ad-services/user-identity/adms-identity-repository/adms-client.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	return fandomDomains.find((domain) => window.location.hostname.includes(domain))
+		? 'https://services.fandom.com/'
+		: 'https://services.fandom-dev.' +
+				(location.hostname.match(/(?!\.)(pl|us)$/) || ['us'])[0] +
+				'/';
+}
+
+class AdmsClient {
+	storage = new UniversalStorage();
```
nandy-andy, line 17: I wonder if it's worth passing it as an argument in the constructor just in case we'd like to replace `UniversalStorage` with something else 🤔

src/ad-services/user-identity/index.ts
```diff
@@ -1,18 +1,36 @@
...
+	}
 
-class UserIdentity extends BaseServiceSetup {
 	async setupPPID(): Promise<void> {
-		if (window.SilverSurferSDK?.requestUserPPID) {
-			const ppid = await window.SilverSurferSDK.requestUserPPID(
-				context.get('services.ppidAdmsStorage.enabled'),
-			);
-			context.set('targeting.ppid', ppid);
+		try {
```
nandy-andy, line 21: Should we make `setupPPID()` also private? I'm guessing we'd like clients to use only `call()` 🤔

src/ad-services/user-identity/adms-identity-repository/adms-client.ts
```diff
@@ -0,0 +1,66 @@
...
+			},
+			mode: 'cors',
+			credentials: 'include',
+			method: 'POST',
+			body: JSON.stringify(action),
+		});
+	}
+
+	getLocalData(): ActiveData {
+		return this.storage.getItem('silver-surfer-active-data-v2') as ActiveData;
```
nandy-andy, line 58: Consider keeping the item's name in a class field, so then you can reuse it in these two lines: 58 and 62 ;)

src/ad-services/user-identity/adms-identity-repository/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IdentityRepositoryInterface } from '../identity-repositories';
+import { utils } from '@ad-engine/core';
+import { UserIdentity } from '../';
+
+class AdmsIdentityRepository implements IdentityRepositoryInterface {
+	async get(): Promise<string> {
+		const userData = await admsService.get();
+		utils.logger(UserIdentity.logGroup, userData);
+		let identity: IdentityAction = userData?.IDENTITY?.find(
+			(id) => id.payload.identityType === 'ppid',
```
nandy-andy, line 13: Shall we extract it, so it's easily extensible in the future without modifying the code here? I'm thinking here that in the future we'd like also to retrieve HEMs for example from the ADMS but maybe I'm confusing different topics 😊 But if I am not, then adding more types will force us to modify this piece of code and it'll get ugly after a while 🙈 I'm afraid of something like:
```js
let identity: IdentityAction = userData?.IDENTITY?.find(
			(id) => id.payload.identityType === 'ppid' || id.payload.identityType === 'ppid' || id.payload.identityType === 'nandy-is-over-engineering'
);
```

src/ad-services/user-identity/index.ts
domjanik: Unfortunately - I can't make it private if I want to spy on it in tests :(

nandy-andy: Pity, thanks for checking!

nandy-andy: Hm, alternatively we could create a `didSetup` public flag to check in the tests and then make `setupPPID()` private 🤔 But maybe it's not worth it... I mean we (the AdEng team) are the only clients of the service ;p

nandy-andy (approved):

src/ad-services/user-identity/adms-identity-repository/adms-client.ts
```diff
@@ -0,0 +1,66 @@
...
+import { UserIdentity } from '../index';
+import { UniversalStorage } from '../../../ad-engine/services/universal-storage';
+
+export function getServicesBaseURL() {
+	const fandomDomains = ['fandom.com', 'muthead.com', 'futhead.com'];
+
+	return fandomDomains.find((domain) => window.location.hostname.includes(domain))
+		? 'https://services.fandom.com/'
+		: 'https://services.fandom-dev.' +
+				(location.hostname.match(/(?!\.)(pl|us)$/) || ['us'])[0] +
```
mfarul, line 12: this could be used as index solely

mfarul (approved):

PR #1545, opened 2022-09-19
ADEN-12189 Integrate LiveConnect into f2

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12189

### DIFF
--- a/src/ad-services/live-connect/index.ts
+++ b/src/ad-services/live-connect/index.ts
@@ -14,7 +14,6 @@ const liveConnectScriptUrl = 'https://b-code.liadm.com/a-07ev.min.js';
 const idConfigMapping: IdConfig[] = [
 	{ id: 'unifiedId', name: `${partnerName}-unifiedId` },
 	{ id: 'sha2', name: partnerName, params: { qf: '0.3', resolve: 'sha2' } },
-	{ id: 'sha256', name: `${partnerName}-sha256`, params: { qf: '0.3', resolve: 'sha256' } },
 	{ id: 'md5', name: `${partnerName}-md5`, params: { qf: '0.3', resolve: 'md5' } },
 	{ id: 'sha1', name: `${partnerName}-sha1`, params: { qf: '0.3', resolve: 'sha1' } },
 ];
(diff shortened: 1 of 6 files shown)


domjanik (approved):

domjanik:
Looks good to me :)

src/ad-services/live-connect/index.ts
```diff
@@ -39,6 +43,8 @@ class LiveConnect extends BaseServiceSetup {
 			return;
 		}
 
+		this.isLoaded = this.isAvailableInLocalStorage(`${partnerName}-unifiedId`);
```
mfarul, line 46: Here we should check for unifiedId and sha2 presence as the first one will be almost always available but sha2 could be resolved later. 
Then I would also add checking optional once in resolveId.

mfarul (changes requested):

mfarul (approved):

src/ad-services/live-connect/index.ts
mfarul: On the other hand we use session storage here, so the chance of resolving sha2 during session is almost 0%. Let's leave it as it is.

domjanik (approved):

src/ad-services/live-connect/index.ts
```diff
@@ -49,23 +56,27 @@ class LiveConnect extends BaseServiceSetup {
...
 					this.track();
 				});
-
-			this.isLoaded = true;
+		} else {
+			utils.logger(logGroup, `already loaded and available in ${this.storageConfig.type}Storage`);
 		}
 	}
 
 	resolveId(idName, partnerName) {
```
mfarul, line 64: shouldn't we skip resolving if we have ID in cache?

src/ad-services/live-connect/index.ts
```diff
@@ -92,6 +103,67 @@ class LiveConnect extends BaseServiceSetup {
...
+			return this.storage.get(key);
+		} else {
+			return this.storage.getItem(key);
+		}
+	}
+
+	saveToStorage(key: string, value: string) {
+		const idName = this.getConfigId(key);
+
+		if (!this.storageConfig.mandatoryParams.includes(idName)) {
```
mfarul, line 149: why saving only mandatory? those are only set up to skip or fire resolving

PR #1548, opened 2022-09-21
ADEN-12235 Improved handling JWP player in pipeline-approach initialization

domjanik:


platforms/f2/modes/f2-ads-mode.service.ts
```diff
@@ -29,7 +30,11 @@ export class F2AdsMode implements DiProcess {
 					timeout: context.get('options.jwpMaxDelayTimeout'),
 				}),
 				gptSetup.setOptions({
-					dependencies: [userIdentity.initialized],
+					dependencies: [
+						userIdentity.initialized,
+						playerSetup.initialized,
+						jwPlayerInhibitor.initialized,
```
nandy-andy, line 36: Hm, it suggests we always delay `gptSetup` with `playerSetup` and `jwPlayerInhibitor` but it's only partially true. ATM I can't think of a better way of signalling it, though.

nandy-andy (approved):
I think the idea and direction is clear: we don't want `PlayerSetup` to always delay the setup (only on pages with the article video [legacy name: featured video]) 👍 I just left a style note and a thought we can keep in mind.

Let's now test it in all the strange scenarios:
1. the article video scenario on FandomDesktop/FandomMobile ([1](https://project43.fandom.com/wiki/SyntheticTests/Premium/FeaturedVideo))
2. the UAP:JWP scenario on FandomDesktop/FandomMobile ([1](https://harrypotter.fandom.com/wiki/List_of_spells?icbm__icUAPJWPlayer=true&icbm__icUAPJWPlayerDelay=10000&cid=adeng-uap-jwp&adengine_version=ADEN-11170))
3. the JWP pages on f2 ([1](https://www.fandom.com/articles/hgt-cult-of-the-lamb), [2](https://www.fandom.com/articles/the-best-movies-of-2017-so-far))
4. the video page on f2 ([1](https://www.fandom.com/video))

PR #1552, opened 2022-09-23
ADEN-12209 Change the video ads logic

nandy-andy:
The code changes introduce more requests for video ads for 3rd, 4th and so on video plays. We block requesting ads for the 2nd video because it may be a sponsored one.

### DIFF
--- a/src/ad-engine/listeners/listeners.ts
+++ b/src/ad-engine/listeners/listeners.ts
@@ -13,6 +13,7 @@ export interface VideoData {
 	position: string;
 	user_block_autoplay?: -1 | 0 | 1;
 	video_id?: string;
+	video_depth?: number | string;
 }
 
 export interface VideoEventData extends VideoData {
--- a/src/ad-engine/listeners/listeners.ts
+++ b/src/ad-engine/listeners/listeners.ts
@@ -13,7 +13,7 @@ export interface VideoData {
 	position: string;
 	user_block_autoplay?: -1 | 0 | 1;
 	video_id?: string;
-	video_depth?: number | string;
+	video_number?: number | string;
 }
 
 export interface VideoEventData extends VideoData {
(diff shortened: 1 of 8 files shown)


src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -114,11 +114,9 @@ export class JWPlayerHelper {
 	}
 
 	private shouldPlayAdOnNextVideo(depth: number): boolean {
-		const capping = context.get('options.video.adsOnNextVideoFrequency');
+		const SPONSORED_VIDEO_INDEX = 2;
 
-		return (
-			context.get('options.video.playAdsOnNextVideo') && capping > 0 && (depth - 1) % capping === 0
-		);
+		return context.get('options.video.playAdsOnNextVideo') && depth !== SPONSORED_VIDEO_INDEX;
```
mfarul, line 119: how  about making the `options.video.playAdsOnNextVideo` a skip index variable? E.g. 
`context.get('options.video.skipAdsOnVideoIndex') && context.get('options.video.skipAdsOnVideoIndex') != depth`

src/ad-engine/listeners/listeners.ts
```diff
@@ -13,6 +13,7 @@ export interface VideoData {
 	position: string;
 	user_block_autoplay?: -1 | 0 | 1;
 	video_id?: string;
+	video_depth?: number | string;
```
mfarul, line 16: is it depth the index or length of the video list?

nandy-andy: it's the video number on the list weirdly counted from 1 ;)

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
nandy-andy: Doable. With a bit more effort we can go with an array of video numbers to skip requesting an ad ;) I'm just not sure if we need since we want to introduce a more advanced logic in the future that will know which video is a sponsored one 🤔

dmnsobczak (approved):

PR #1563, opened 2022-09-29
ADEN-12189 Add LI to N&S

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12189

### DIFF
--- a/src/ad-services/live-connect/index.ts
+++ b/src/ad-services/live-connect/index.ts
@@ -75,7 +75,7 @@ class LiveConnect extends BaseServiceSetup {
 				return;
 			}
 
-			this.saveToStorage(partnerName, partnerIdentityId);
+			this.storage.setItem(partnerName, partnerIdentityId, this.storageConfig.ttl);
 
 			communicationService.emit(eventsRepository.IDENTITY_PARTNER_DATA_OBTAINED, {
 				partnerName,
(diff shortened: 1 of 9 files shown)


src/ad-services/live-connect/index.ts
```diff
@@ -89,9 +100,48 @@ class LiveConnect extends BaseServiceSetup {
 			return;
 		}
 		idConfigMapping.forEach(({ id, name, params }) => {
-			this.resolveAndReportId(id, name, params);
+			const storageKey = this.getConfigName(id);
```
domjanik, line 103: getConfigName returns the same name as you get 1 line above in forEach. Is it really necessary to use this method here ?

domjanik (approved):

src/ad-services/live-connect/index.ts
```diff
@@ -89,9 +100,48 @@ class LiveConnect extends BaseServiceSetup {
...
+		if (!this.storageConfig) {
+			return true;
+		}
+
+		let shouldLoadScript = false;
+
+		this.storageConfig.mandatoryParams.forEach((param) => {
+			const storageKey = this.getConfigName(param);
+			if (!this.isAvailableInStorage(storageKey)) {
+				shouldLoadScript = true;
```
mfarul, line 131: and break ;)

src/ad-services/live-connect/index.ts
szymon-nawrocki: Good catch! It is not necessary. Fixed ✅

src/ad-services/live-connect/index.ts
szymon-nawrocki: done ✅

PR #1566, opened 2022-10-03
ADEN-12278 Add feature flag for collapse messages AKA MessageBox

nandy-andy:


platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
```diff
@@ -238,7 +238,7 @@ export class UcpMobileDynamicSlotsSetup implements DiProcess {
 	}
 
 	private registerAdPlaceholderService(): void {
-		const messageBoxService = new MessageBoxService();
+		const messageBoxService = new MessageBoxService(context.get('services.messageBox.enabled'));
```
dmnsobczak, line 241: Usually we don't pass context keys as values in constructor, because context is a global object and class can read directly from it. In test we are also setting it and removing at the end of the spec. Why you did it like that? New trend? :D

dmnsobczak (approved):
Basically code is technically ok, but two code-style-type comments left. It's fine for me to move the ticket to QA after resolving them :)

platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts
nandy-andy: This way we remove hard and hidden dependency on `Context` in the `MessageBoxService`. It's also easier to unit test `MessageBoxService` when you just pass the flag instead of reaching out directly to the god object.

mfarul: I'll prefer nAndy approach as it gives an option to test the class correctly without mocking some global objects. Well designed class should get's all what is needed on construction time. Either this is a config object or just a flag it does not matter, but it should NOT use any global configuration.

szymon-nawrocki (approved):

PR #1575, opened 2022-10-10
ADEN-12303 Extend Load Times Tracking to Fut, Mut and F2

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12303


jakubbanas (approved):

src/platforms/f2/f2-platform.ts
```diff
@@ -42,9 +43,10 @@ export class F2Platform {
 			F2IocSetup,
 			TrackingParametersSetup,
 			() => context.set('state.isMobile', f2env.isPageMobile),
+			F2TargetingSetup,
```
nandy-andy, line 46: This should be fine as the RxJS action executes quite soon: https://github.com/Wikia/f2/blob/65c908dbca5166bb27d9d0aa0a25ca966da725f9/frontend/bundles/_common.js#L53

nandy-andy (approved):

PR #1577, opened 2022-10-11
ADEN-12171 | case insensitive VIDEO_TIER_3 bundle recognition

mfarul:
We will upload bundles as TAGS into tag manager and they are set up in lower case there, whereas in ICBM we use upper case ones. For GAM it does not matter

### DIFF
--- a/spec/core/utils/targeting.spec.ts
+++ b/spec/core/utils/targeting.spec.ts
@@ -11,6 +11,11 @@ describe('targeting', () => {
 	});
 
 	afterEach(() => {
+		context.remove('targeting.bundles');
+		context.remove('targeting.esrb');
+		context.remove('targeting.s1');
+		context.remove('targeting.skin');
+
 		sandbox.restore();
 	});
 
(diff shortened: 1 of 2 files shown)


dmnsobczak (approved):

spec/core/utils/targeting.spec.ts
```diff
@@ -73,6 +73,35 @@ describe('targeting', () => {
 		).to.deep.equal([]);
 	});
 
+	it('getTargetingBundles will not add VIDEO_TIER_3_BUNDLE for tier 1 and 2', () => {
+		context.set('targeting.s1', '_project34');
+		context.set('targeting.skin', 'ucp_desktop');
```
nandy-andy, line 78: It could be good for future if we reset the context settings for a test case in `afterEach()` hook. So for this file, I'd add in the line #14:
```ts
context.remove('targeting.bundles');
context.remove('targeting.s1');
context.remove('targeting.skin');
```

This way we won't pollute `context` for other unit/integration tests.

nandy-andy (approved):

PR #1578, opened 2022-10-11
ADEN-12284 - Added additional trackings to LiveRamp

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12284)


szymon-nawrocki (approved):
The code looks ok for temporary analysis, but we should think about a better place/way to  track these partner events than `adengloadtimes` DW table

PR #1582, opened 2022-10-13
ADEN-12161 Adding new bidder Ogury

dmnsobczak:


@types/window/index.d.ts
```diff
@@ -8,6 +8,7 @@ interface Window {
 	__tcfapi: WindowTCF;
 	__uspapi?: WindowUSP;
 	AdEngine_adType?: ManualAdType;
+	AdEngine_adDemand?: string;
```
nandy-andy, line 11: Do we need to add it? It could be a "no_template" adType 🤔

nandy-andy (dismissed):
I wonder if we really need to make `AdSlot` class bigger and changing existing code 🙈 I'd like to see it implemented with just adding new functionality and no touching existing code or touching it in a minimal way ;)

While thinking about it one idea came to my mind: `postMessage()` from Ogury creative or `postQuecast` like in the [SequentialMessaging](https://fandom.atlassian.net/wiki/spaces/ADEN/pages/2338488376/Sequential+Messaging+v0.3+-+replaced+with+v0.4+v1.0+Draft) implementation. What do you think? We can brainstorm it on a call.

The basic idea: the template gets in with `forced_success` and sends the event, AdEngine receives the event and either suppresses rendering the `floor_adhesion` template or hides it. Do you think it could work? It should require less changes in existing code.

@types/window/index.d.ts
dmnsobczak: Discussed on Slack - will be changed

nandy-andy (approved):

PR #1583, opened 2022-10-14
ADEN-12273 - Added reporting failed pixel loading in Eyeota

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12273)


szymon-nawrocki (changes requested):
EYEOTA_FAILED event should be added to https://github.com/Wikia/ad-engine/blob/7b8428958cefafe020b41d16e83aaf5a36b733d1/src/platforms/shared/tracking/load-times-tracker.ts#L5

szymon-nawrocki (approved):

PR #1594, opened 2022-10-31
ADEN-12231 - Added frontend integration with IdentityStorage

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12231)


src/ad-services/user-identity/utils/service-base-url.ts
```diff
@@ -0,0 +1,9 @@
+export function servicesBaseURL() {
```
szymon-nawrocki, line 1: I would stick with the name we had before `getServicesBaseURL` as it better describes the method

spec/ad-services/user-identity/adms-strategy.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+		v4Stub.returns(mockId);
+	});
+	afterEach(() => {
+		sandbox.restore();
+		context.remove('services.ppid.enabled');
+		context.remove('services.ppidRepository');
+	});
+
+	it('use ADMS strategy and gets PPID from API', async () => {
+		context.set('services.ppid.enabled', true);
```
szymon-nawrocki, line 27: These 2 lines are repeated in every test case, we can call it once in `before` block.
Also we should reset the context variables `after` the tests are done, so it will save us from unexpected behaviour in different tests (because those context variables will be set in another tests)

spec/ad-services/user-identity/identity-storage-strategy.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+		context.remove('services.ppid.enabled');
+		context.remove('services.ppidRepository');
+	});
+
+	it('use Identity Storage strategy and gets synced PPID from API', async () => {
+		context.set('services.ppid.enabled', true);
```
szymon-nawrocki, line 25: Same here, these 2 variables are called in every test case (4 times) which is unnecessary

spec/ad-services/user-identity/local-storage-strategy.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+		context.remove('services.ppid.enabled');
+		context.remove('services.ppidRepository');
+	});
+
+	it("use LocalStore strategy and don't have PPID stored", async () => {
+		context.set('services.ppid.enabled', true);
```
szymon-nawrocki, line 27: Same here

src/ad-services/user-identity/identity-storage-repository/identity-storage-client.ts
```diff
@@ -0,0 +1,48 @@
...
+	async postData(data: IdentityStorageDto): Promise<IdentityStorageDto> {
+		return fetch(this.ISUrl + 'identity', {
+			credentials: 'include',
+			mode: 'cors',
+			headers: {
+				'Content-Type': 'application/json',
+			},
+			method: 'PUT',
+			body: JSON.stringify({ ppid: data.ppid }),
+		}).then((response) => response.json());
```
szymon-nawrocki, line 36: In case of error, won't the error crash the app? Maybe it will be worth to add `catch` statement with the log like "Updating Identity Storage data failed,  error"?

szymon-nawrocki (approved):

PR #1598, opened 2022-11-04
ADEN-12061 Preparing StickyICB for the experiment

dmnsobczak:


src/core/services/slot-impact-watcher.ts
```diff
@@ -0,0 +1,43 @@
+interface ImpactDetails {
+	id: string;
```
rafalgiemza, line 2: why not a number? :)

rafalgiemza (approved):

PR #1599, opened 2022-11-07
ADEN-12241 Add tracking for Anyclip player

nandy-andy:


src/ad-services/anyclip/index.ts
```diff
@@ -35,8 +41,8 @@ class Anyclip {
 			.then(() => {
 				incontentPlayerContainer.classList.remove('hide');
 				utils.logger(logGroup, 'ready');
+
+				this.tracker.register();
```
rafalgiemza, line 45: Are we sure, tracker isn't null here? 
Maybe good idea is to add class interface like here: https://www.typescriptlang.org/docs/handbook/interfaces.html#class-types
?

rafalgiemza (approved):
code looks good for me, only 1 suggestion from me :) 
good job @nandy-andy

rafalgiemza (approved):

PR #1601, opened 2022-11-08
ADEN-12354 | combine bundles from page tags

mfarul:


src/platforms/shared/context/targeting/targeting-strategies/strategies/combined-strategy.ts
```diff
@@ -24,10 +24,19 @@ export class CombinedStrategy extends CommonStrategy implements TargetingStrateg
 
 		const wiki: MediaWikiAdsContext = context.get('wiki');
 
+		const bundles = this.context.site.tags?.bundles || [];
```
szymon-nawrocki, line 27: To make `execute` method a little cleaner I'd move the line 27-35 logic to function `getCombinedBundles()` which returns combined bundles

szymon-nawrocki (approved):

PR #1607, opened 2022-11-15
ADEN-12438 Create a bundle for GameFAQs and Gamespot

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12438

### DIFF
--- a/src/platforms/gamefaqs/ad-context.ts
+++ b/src/platforms/gamefaqs/ad-context.ts
@@ -1,7 +1,9 @@
 export const basicContext = {
-	adUnitId: '/{custom.dfpId}/aw-gamefaqs',
+	adUnitId: '/{custom.dfpId}/{custom.region}-{custom.property}',
 	custom: {
 		dfpId: '5441',
+		region: 'aw',
+		property: 'gamefaqs',
 	},
 	slots: {},
 	src: ['test'],
--- a/src/platforms/gamefaqs/ad-context.ts
+++ b/src/platforms/gamefaqs/ad-context.ts
@@ -1,0 +1,8 @@
+export const basicContext = {
+	adUnitId: '/{custom.dfpId}/aw-gamefaqs',
+	custom: {
+		dfpId: '5441',
+	},
+	slots: {},
+	src: ['test'],
+};
--- a/src/platforms/gamefaqs/gamefaqs-platform.ts
+++ b/src/platforms/gamefaqs/gamefaqs-platform.ts
@@ -1,0 +1,25 @@
+import { Injectable } from '@wikia/dependency-injection';
+import { communicationService, eventsRepository, ProcessPipeline } from '@wikia/ad-engine';
+import { gptSetup } from '@platforms/shared';
+
+import { GamefaqsSlotsContextSetup } from './setup/context/slots/gamefaqs-slots-context.setup';
+import { GamefaqsDynamicSlotsSetup } from './setup/dynamic-slots/gamefaqs-dynamic-slots.setup';
+
+@Injectable()
+export class GamefaqsPlatform {
+	constructor(private pipeline: ProcessPipeline) {}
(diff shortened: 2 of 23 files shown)


src/platforms/gamefaqs/ad-context.ts
```diff
@@ -0,0 +1,8 @@
+export const basicContext = {
+	adUnitId: '/{custom.dfpId}/aw-gamefaqs',
```
nandy-andy, line 2: Something to be addressed in the future probably. We could move `'aw'` to `custom` under `region` field so it follows the current config (`dfpPath.region`). But I've never seen anything else than `'aw'`, so probably it's not used and therefore we could address it in the future ;)

src/platforms/gamefaqs/setup/context/slots/gamefaqs-slots-context.setup.ts
```diff
@@ -0,0 +1,22 @@
...
+			leader_plus_top: {
+				defaultSizes: [
+					[728, 90],
+					[970, 66],
+					[970, 250],
+				],
+			},
+			mpu_top: {
+				defaultSizes: [[300, 250]],
+			},
```
nandy-andy, line 17: I was thinking if we could add here all the slot names and sizes from the current config (look for `"units"` node in the JSON).

src/platforms/gamefaqs/setup/dynamic-slots/gamefaqs-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,20 @@
+import { context, DiProcess } from '@wikia/ad-engine';
+
+export class GamefaqsDynamicSlotsSetup implements DiProcess {
+	execute(): void {
+		this.wrapActiveAdSlots();
+	}
+
+	private wrapActiveAdSlots(): void {
```
nandy-andy, line 8: It's not really wrapping but injecting, so I'd rename the method :) In more advanced code of ours (ucp_mobile or ucp_desktop) you can actually find `injectSlots()` function being used in the dynamic-slots but I wonder if we could go with something simpler first 😊 If we get to an obstacle easily solved by adding similar logic to UCP bundles we'd go there but maybe before we could rethink the whole slots/dynamic slots setup again ;)

src/platforms/gamefaqs/gamefaqs-platform.ts
```diff
@@ -0,0 +1,25 @@
...
+export class GamefaqsPlatform {
+	constructor(private pipeline: ProcessPipeline) {}
+
+	execute(): void {
+		this.pipeline.add(
+			// TODO: we need a CMP step here, so we won't call for ads unless we have a clear idea of the privacy policy of a visitor
+			// TODO: to decide if we want to call instant-config service for the first releases?
+			GamefaqsSlotsContextSetup,
+			GamefaqsDynamicSlotsSetup,
+			// TODO: add targeting setup once we have idea of page-level and slot-level targeting
```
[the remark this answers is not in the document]
szymon-nawrocki: I think I have done it by adding `targeting.pos` to file: src/platforms/gamefaqs/setup/context/slots/gamefaqs-slots-context.setup.ts

src/platforms/gamefaqs/setup/context/slots/gamefaqs-slots-context.setup.ts
szymon-nawrocki: done ✅

src/platforms/gamefaqs/gamefaqs-platform.ts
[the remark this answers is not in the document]
nandy-andy: Replied offline ;)

nandy-andy (approved):

PR #1608, opened 2022-11-16
SPORTS-355 add pth to sports targeting

preludek:


src/platforms/sports/setup/context/targeting/sports-targeting.setup.ts
```diff
@@ -38,14 +39,12 @@ export class SportsTargetingSetup implements DiProcess {
...
 
-	private getSportsPageType(): string {
-		const pathName = window.location.pathname;
-		const hostname = window.location.hostname.toLowerCase();
-		const pieces = hostname.split('.').filter((piece) => piece !== 'www');
+	private getPathName(): string {
+		return window.location.pathname;
+	}
 
-		if (pieces[0] === 'old') {
```
preludek, line 46: That's supposed to be redudant.

src/platforms/sports/setup/context/targeting/sports-targeting.setup.ts
```diff
@@ -20,6 +20,7 @@ export class SportsTargetingSetup implements DiProcess {
 			s0: 'gaming',
 			s1: context.get('application'),
 			s2: this.isSquadPage() ? 'squad' : this.getSportsPageType(),
```
preludek, line 22: We thought of replacing that with what we needed, but after consultation with Seba, we decided to just add a new key.

dmnsobczak (approved):

PR #1611, opened 2022-11-17
ADEN-12355 short/long page

lnawrotek:


src/core/utils/targeting.ts
```diff
@@ -80,6 +81,13 @@ class Targeting {
 			bundles.push('VIDEO_TIER_3_BUNDLE');
 		}
 
+		const wordCount = context.get('targeting.word_count') || -1;
+
+		if (wordCount <= shortPageWordsLimit) {
```
nandy-andy, line 86: I wonder if we shouldn't push `short_page` bundle only if we do have the word count and it's smaller or equals the limit 🤔 So, the condition would be `wordCount > -1 && wordCount <= shortPageWordsLimit`. This way we won't count some special case pages when the MediaWiki's Output won't return the HTML -- we'll focus mostly on the articles, I guess 🤔 But this is something worth putting in the QA notes - to check different page types.

lnawrotek: It's a good question. We have to decide which type is the default type. I think `wordCount > -1 && wordCount <= shortPageWordsLimit` is OK, we can focus on articles.

spec/core/utils/targeting.spec.ts
```diff
@@ -50,7 +50,7 @@ describe('targeting', () => {
 					skin: ['fandom_desktop'],
 				},
 			}),
-		).to.deep.equal(['bundle2', 'bundle3']);
+		).to.deep.equal(['bundle2', 'bundle3', 'short_page']);
```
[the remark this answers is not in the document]
lnawrotek: I can remove the `short_page` bundle if I change condition https://github.com/Wikia/ad-engine/pull/1611#discussion_r1026263177

src/core/utils/targeting.ts
nandy-andy: Let's do it this way then 👌

lnawrotek: I've added a unit test to check `word_count = -1`

nandy-andy (approved):

PR #1612, opened 2022-11-17
ADEN-12455 Fix only blue ad bug on fandom.com/video

nandy-andy:


nandy-andy:
Not needed anymore. We got it fixed in a better place and it's not quick and dirty anymore ;) Closing.

PR #1618, opened 2022-11-22
ADEN-12412 sponsored videos

lnawrotek:
https://fandom.atlassian.net/browse/ADEN-12412


src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -113,6 +113,18 @@ export class JWPlayerHelper {
 		);
 	}
 
+	requestForSponsoredVideos() {
```
mfarul, line 116: do we need this here or should this be only available in the child class JwplayerHelperSkippingSponsoredVideo ?

nandy-andy: I'd move it to the helper class and send the request only in case of fallback: https://github.com/Wikia/ad-engine/blob/dev/src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts#L58-L65

So, when `window.sponsoredVideo` is not set by JWPlayer let's call only then the API. It's one less request on pageview for most of the visits. @lnawrotek, do you see any potential issues with this approach that we miss? 🤔

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -113,6 +113,18 @@ export class JWPlayerHelper {
...
+		if (!Array.isArray(window.sponsoredVideos)) {
+			const url = utils.getServicesBaseURL() + 'article-video/jw-platform-api/get-sponsored-videos';
```
nandy-andy, line 118: We could add additional logging to the browser console here informing us it's fallbacking, for example:
```
utils.logger(
  JWPlayerHelper.LOG_GROUP_NAME,
  'Incorrect window.sponsoredVideos, using fallback to Pandora!',
  window.sponsoredVideos,
);
```

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -113,6 +113,18 @@ export class JWPlayerHelper {
...
+
+			utils.scriptLoader.loadAsset(url).then((response) => {
+				if (Array.isArray(response)) {
+					window.sponsoredVideos = response;
+				}
```
nandy-andy, line 123: In order to make debugging and testing easier we could add `else` with a log to the browser's console:
```js
utils.logger(
  JWPlayerHelper.LOG_GROUP_NAME,
  'Incorrect sponsored videos list from Pandora!',
  response,
);
```

src/ad-products/video/jwplayer/handlers/jwplayer-handler.ts
```diff
@@ -96,6 +96,9 @@ export class JWPlayerHandler {
 	private adsManager(): Observable<unknown> {
 		return this.stream$.pipe(
 			ofJwpEvent('adsManager'),
+			tap(async () => {
+				await this.helper.ensureAdditionalSettings();
```
dmnsobczak, line 100: `{}` are nod needed for one-line function bodies :D

dmnsobczak (approved):

PR #1623, opened 2022-11-24
ADEN-12346 Adding FanCentral bundle

dmnsobczak:


src/platforms/f2/ad-context.ts
```diff
@@ -51,9 +51,6 @@ export const basicContext = {
 			appName: 'f2',
 			fallback: 'https://script.wikia.nocookie.net/fandom-ae-assets/icbm/prod/icbm_state_f2.json',
 		},
-		moatYi: {
-			partnerCode: 'wikiaprebidheader490634422386',
```
nandy-andy, line 55: How come we're removing something in News&stories bundle where the task is about creating a new bundle? :D Is it a left over that we didn't clean up properly?

src/platforms/fan-central/fan-central-platform.ts
```diff
@@ -0,0 +1,65 @@
...
+
+@Injectable()
+export class FanCentralPlatform {
+	constructor(private pipeline: ProcessPipeline, private noAdsDetector: NoAdsDetector) {}
+
+	execute(): void {
+		// Config
+		this.pipeline.add(
+			() => context.extend(basicContext),
+			PlatformContextSetup,
```
nandy-andy, line 39: Do we need it now? Is there `mw.config` on Fan Central? Is there `fandomContext` available?

src/platforms/fan-central/fan-central-platform.ts
```diff
@@ -0,0 +1,65 @@
...
+			parallel(InstantConfigSetup, () => bootstrapAndGetConsent()),
+			TrackingParametersSetup,
+			() => context.set('state.isMobile', getDeviceMode() === 'mobile'),
+			FanCentralTargetingSetup,
+			LoadTimesSetup,
+			BaseContextSetup,
```
nandy-andy, line 45: Do we really need it at the beginning?

src/platforms/fan-central/setup/context/slots/fan-central-slots-context.setup.ts
```diff
@@ -0,0 +1,124 @@
...
+							[320, 50],
+							[320, 100],
+							[728, 90],
+						],
+					},
+				],
+			},
+		};
+
+		slotsContext.setupSlotVideoContext();
```
nandy-andy, line 120: I don't think we need it now. It may be needed when we start working on JWPlayer ads integration on Fan Central.

src/platforms/fan-central/setup/context/targeting/fan-central-targeting.setup.ts
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	private getPageLevelTargeting(): Partial<Targeting> {
+		const targeting: Partial<Targeting> = {
+			host: window.location.hostname,
+			lang: 'en',
+			post_id: '-1',
+			skin: `fc_${context.get('state.isMobile') ? 'mobile' : 'desktop'}`,
+			s0: 'fan-central',
+			s1: '_fc',
```
nandy-andy, line 32: Let's share the targeting values with AdOps, so there are no surprises.

src/platforms/fan-central/setup/dynamic-slots/fan-central-dynamic-slots.setup.ts
```diff
@@ -0,0 +1,73 @@
...
+		if (!slotDefinition) {
+			return;
+		}
+
+		if (document.querySelector(slotDefinition.slotCreatorConfig.anchorSelector)) {
+			insertSlots([slotDefinition]);
+			return;
+		}
+
+		new utils.WaitFor(
```
nandy-andy, line 48: Why can't we just add empty placeholders in the app and fill them when GAM responds with an ad? Adding a microtask is an overkill and a way for introducing unexpected behaviour because of race conditions :/

src/platforms/f2/ad-context.ts
dmnsobczak: Yes, this is a garbage - old, forgotten key, so I decided to remove it :)

src/platforms/fan-central/fan-central-platform.ts
dmnsobczak: You're right - this step is not needed here 👍

src/platforms/fan-central/fan-central-platform.ts
dmnsobczak: Of course: this sets all required ICBM variables (that can be enabled for this new platform), partners flags and no-ads settings.

src/platforms/fan-central/setup/context/targeting/fan-central-targeting.setup.ts
dmnsobczak: Sure. CC @sseyda

src/platforms/fan-central/setup/dynamic-slots/fan-central-dynamic-slots.setup.ts
dmnsobczak: Simple reasons:
1. This is a experiment so we might need to move slots to the other positions - declaring placeholders in FanCentral requires more work (and releases) to change it and experiment
2. FanCentral won't have a real ad slots and these are used only for UAP - to be honest, TLB could be injected next to the `head` element if we are sure that UAP would render properly. So if you speaking about placeholders that reserve space in order to minimize the CLS impact, then they are not allowed on FanCentral.
3. FanCentral page is rendered in multiple steps and not every selector is available right away - this will require us to define a set of event emitted by FanCentral up that trigger slot injections and this sound like a bigged task that need solid definitions that I would like to extract to a separate ticket :)

src/platforms/fan-central/fan-central-platform.ts
nandy-andy: Do we need all of the variables like the steam one, prebidium one, bidders ones, sticky TLB one etc.? 🤔  I'd go with minimal stuff required for the AB tests first. It gives us two benefits (maybe more if we think more about it): smaller bundle and thanks to this exercise we can rethink what's supposed to be in the "basic" context setup as I tend to think in this way of "BaseContextSetup" 😊

src/platforms/fan-central/setup/context/slots/fan-central-slots-context.setup.ts
nandy-andy: Sure, there is Porvata player but I don't think we use `audio` or `videoDepth` key-vals in UAP Porvata, so maybe we can remove it for now 🤔

src/platforms/fan-central/setup/dynamic-slots/fan-central-dynamic-slots.setup.ts
nandy-andy: So, are you saying we're taking shortcuts now to do it properly after we figure out the end state of ads in Fan Central? I can live with it as long as there is a summary/clean-up ticket 😅

src/platforms/fan-central/setup/context/targeting/fan-central-targeting.setup.ts
nandy-andy: Or if we demo Fan Central again we can mention the hardcoded values in targeting ;)

src/platforms/fan-central/setup/context/slots/fan-central-slots-context.setup.ts
nandy-andy: But if `videoDepth` is then being translated to `rv` it may be useful. However, I recall a comment in some old code before this line saying "TODO: find a better place for it" and I thought: we don't need it now, let's remove it from here and if it's needed we'll think of a better place 😄

dmnsobczak: RIght, I see it now. Removed.

PR #1624, opened 2022-11-24
Aden 12430

lnawrotek:
https://fandom.atlassian.net/browse/ADEN-12430


nandy-andy (approved):
It should be fine for now but we'd need a follow-up to either add unit tests or refactor the JWPlayer logic 😊

PR #1629, opened 2022-11-28
ADEN-12452 Improve 1st video preroll behaviour 

nandy-andy:
Moving `sponsoredVideos` from global to class property and ensure its set earlier doesn't help minimise the errors logged to Kibana. This is a 2nd quick-fix that forces the fallback API call to be synchronous.

### DIFF
--- a/src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts
+++ b/src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts
@@ -9,7 +9,7 @@ export class JwplayerHelperSkippingSponsoredVideo extends JWPlayerHelper {
 		protected adSlot: AdSlot,
 		protected jwplayer: JWPlayer,
 		protected readonly targeting: VideoTargeting,
-		private sponsoredVideos = undefined,
+		private sponsoredVideos = window.sponsoredVideos,
 	) {
 		super(adSlot, jwplayer, targeting);
 	}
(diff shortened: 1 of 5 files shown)


dmnsobczak (approved):

src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts
```diff
@@ -53,6 +53,16 @@ export class JwplayerHelperSkippingSponsoredVideo extends JWPlayerHelper {
 		}
 
 		if (!Array.isArray(this.sponsoredVideos)) {
+			this.log(
```
mfarul, line 56: this is called before each play, e.g. for fallback that should not be a problem but with sync method here we should be able to do it once

src/core/utils/script-loader.ts
```diff
@@ -95,6 +95,26 @@ class ScriptLoader implements ScriptLoaderInterface {
 			request.send();
 		});
 	}
+
+	loadSync(url: string): string | boolean {
```
mfarul, line 99: will this work ? 
async loadSync (url) {
     await loadScript(...) 
}

mfarul (changes requested):

src/core/utils/script-loader.ts
nandy-andy: I don't think so. It'll make `loadSync` async and create race-condition with the rest of asyncs in AdEngine and JWPlayer.

src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts
nandy-andy: It only logs to the browser's console if `adengine_debug` is set, so I don't think it's a problem.

mfarul (approved):

src/ad-products/video/jwplayer/helpers/jwplayer-helper-skipping-sponsored-video.ts
nandy-andy: Discussed offline.

PR #1630, opened 2022-11-28
ADEN-12394 Extend Eyeota pixel payload with fandom context tags data

jakubbanas:


spec/ad-services/eyeota/eyeota.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('Eyeota', () => {
 
 	beforeEach(() => {
 		window.__tcfapi = window.__tcfapi as WindowTCF;
+		window.fandomContext = { site: {}, page: {} } as any;
```
szymon-nawrocki, line 12: Instead of `any` you should use `FandomContext` type (https://github.com/Wikia/ad-engine/blob/d7361dcf77671447407039a11e14ff23daa14823/src/platforms/shared/context/targeting/targeting-strategies/models/fandom-context.ts#L3-L2)

szymon-nawrocki: But to be honest I'd delete this line of code

spec/ad-services/eyeota/eyeota.spec.ts
```diff
@@ -79,6 +81,17 @@ describe('Eyeota', () => {
 		context.remove('targeting.s0v');
 	});
 
+	it('constructs proper url with context', async () => {
+		sandbox
+			.stub(window.fandomContext, 'site')
+			.value({ tags: { gnre: [1, 2, 3], pub: ['test'], pform: ['xbox'] } });
```
szymon-nawrocki, line 87: And create a proper FandomContext object instead, like here:
```
		const mockedTags: TaxonomyTags = { gnre: ['1', '2', '3'], pub: ['test'], pform: ['xbox'] }
		const mockedContext: FandomContext = new FandomContext(
			new Site([], true, 'ec', 'test', false, mockedTags, null, 'general'),
			null,
		);

		sandbox
			.stub(window.fandomContext, 'site')
			.value(mockedContext.site);
```

spec/ad-services/eyeota/eyeota.spec.ts
```diff
@@ -25,6 +26,7 @@ describe('Eyeota', () => {
 		sandbox.restore();
 		context.remove('services.eyeota.enabled');
 		delete window.__tcfapi;
+		delete window.fandomContext;
```
szymon-nawrocki, line 29: And delete this one

spec/ad-services/eyeota/eyeota.spec.ts
jakubbanas: I did it on purpose, TS enforces by default FandomContext type, and I don't need all of it, empty object is fine here because later in the test I stub it with the proper value.

Also, I could not delete it, because in order to stub/mock it, there has to be something, even an empty object.

src/ad-services/eyeota/index.ts
```diff
@@ -38,13 +44,24 @@ class Eyeota extends BaseServiceSetup {
...
-		let url = `https://ps.eyeota.net/pixel?pid=${pid}&sid=${siteName}&t=ajs&s0v=${s0v}`;
+
+		const url = new URL('https://ps.eyeota.net/pixel');
+		url.searchParams.append('pid', pid);
+		url.searchParams.append('sid', siteName);
+		url.searchParams.append('t', 'ajs');
+		url.searchParams.append('s0v', s0v);
+
+		if (window.fandomContext?.site?.tags) {
+			const { gnre, pform, pub, tv } = window.fandomContext.site.tags;
```
szymon-nawrocki, line 55: You can use TaxonomyTags interface here:
`const { gnre, pform, pub, tv }: TaxonomyTags = window.fandomContext.site.tags;`

szymon-nawrocki (approved):
👌

PR #1636, opened 2022-11-30
ADEN-12479 | send rv=2+ for 2,3,4,... video

mfarul:


src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -12,11 +12,15 @@ const EMPTY_VAST_CODE = 21009;
...
 
+	private readonly trackEachSeqInRv;
+
 	constructor(
 		protected adSlot: AdSlot,
 		protected jwplayer: JWPlayer,
 		protected readonly targeting: VideoTargeting,
-	) {}
+	) {
+		this.trackEachSeqInRv = context.get('options.video.trackEachSequence') == true;
```
nandy-andy, line 22: One more `=` wouldn't hurt in JS ;)

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -170,10 +174,18 @@ export class JWPlayerHelper {
...
 				v1: state.playlistItem.mediaid || '',
 				...this.targeting,
 			},
 		});
 	}
+
+	private getRvKeyVal(rv: number): string | string[] {
+		if (rv === 1) return '1';
+		if (this.trackEachSeqInRv) {
+			return [rv.toString(), '2+'];
```
nandy-andy, line 187: 👍

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
```diff
@@ -170,10 +174,18 @@ export class JWPlayerHelper {
...
 			targeting: {
-				rv: state.rv,
+				rv: this.getRvKeyVal(state.rv),
 				v1: state.playlistItem.mediaid || '',
 				...this.targeting,
 			},
 		});
 	}
+
+	private getRvKeyVal(rv: number): string | string[] {
```
nandy-andy, line 184: I'd suggest simplifying it and reversing the effect of `this.trackEachSeqInRv`:
```ts
private getRvKeyVal(rv: number): string | string[] {
  if (this.trackEachSeqInRv) {
    return [rv.toString(), '2+'];
  }

  return rv.toString();
}
```
This way we'll keep the "old" functionality (tracking each `rv`) untouched and with the new variable we'll introduce new functionality of passing to GAM `'2+'` to GAM as well. Although, then the new variable would need a new name something like `icMakeRvArray` or something 😊

src/platforms/shared/setup/base-context.setup.ts
```diff
@@ -110,6 +110,7 @@ export class BaseContextSetup implements DiProcess {
 		context.set('options.maxDelayTimeout', this.instantConfig.get('icAdEngineDelay', 2000));
 		context.set('options.jwpMaxDelayTimeout', this.instantConfig.get('icUAPJWPlayerDelay', 0));
 		context.set('options.video.iasTracking.enabled', this.instantConfig.get('icIASVideoTracking'));
+		context.set('options.video.trackEachSequence', this.instantConfig.get('icTrackVideoSeq'));
```
nandy-andy, line 113: I'm not sure if we need it. Maybe we could move the new logic behind the long variable that is still not released to 100% 🤔

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
dmnsobczak:
```suggestion
		this.trackEachSeqInRv = !!context.get('options.video.trackEachSequence');
```
If you really want to be sure that ICBM flag is casted to a simple boolean, but:
* if it's defined as `true/false` in ICBM, then it is
* we don't usually do that :D

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
dmnsobczak: From GAM perspective there is no difference in passing string or array of strings 🤔 Targeting don't need to know about that, unless you read this keyval in some creative template, and as far as I know - we don't do this.

dmnsobczak (approved):

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
dmnsobczak: DIscussed on Slack: why do we even need ICBM variable for that? Let's just add `2+` to the `rv` keyval and done - it's not regressive, two chars in the ad call payload won't do any difference and we won't have more rollout variables to clean up in the future :D

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
nandy-andy: Right, GAM doesn't care ;) However, the name was just an example, not a perfect one 😊 I'm worried because of the regression we're going to introduce when the flag is not set (only `rv=1` or `rv=2+` sent). I'm also not sure the flag is needed without it we'll have it "the old way" `rv=1` or `rv=2` or `rv=3`... and `rv=2+` :)

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
mfarul: we now send all numbers, so i feel that if we will collapse them into 2+ someone will ask where the hell are other :). I can remove the flag after a week

src/ad-products/video/jwplayer/helpers/jwplayer-helper.ts
mfarul: it cannot be simplified we are sending 1 or 2+ when flag is disabled, and 1 or 2+ and 2,3,4,5,6 respectively if flag is enabled.

src/platforms/shared/setup/base-context.setup.ts
mfarul: this one i actually working on 100% - this code is not hidden behind the long one ...

nandy-andy (approved):

PR #1639, opened 2022-12-07
ADEN-12473 - Added integration with Captify

domjanik:


@types/window/index.d.ts
```diff
@@ -16,6 +16,7 @@ interface Window {
 	au_seg?: any;
 	beaconId?: string;
 	beacon_id?: string;
+	captify_kw_query_12974: string;
```
mfarul, line 19: will be always available? eg ?:

src/ad-services/captify/index.ts
```diff
@@ -0,0 +1,35 @@
+import { BaseServiceSetup, context, utils } from '@ad-engine/core';
+
+const logGroup = 'captify';
+
+class Captify extends BaseServiceSetup {
+	PIXEL_URL = 'https://p.cpx.to/p/12974/px.js';
+
+	isEnabled(): boolean {
```
mfarul, line 8: i see this so many times .... maybe a base class method with service variable param ?

src/ad-services/captify/index.ts
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	async call(): Promise<void> {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+
+			return Promise.resolve();
+		}
+		window.captify_kw_query_12974 = '';
+		const elem = document.createElement('script');
```
mfarul, line 24: can we use ScriptLoader.createScript ?

mfarul (approved):

src/ad-services/captify/index.ts
```diff
@@ -0,0 +1,35 @@
...
+		);
+	}
+
+	async call(): Promise<void> {
+		if (!this.isEnabled()) {
+			utils.logger(logGroup, 'disabled');
+
+			return Promise.resolve();
+		}
+		window.captify_kw_query_12974 = '';
```
mfarul, line 23: do we need to set this? Captify script does not handle this on their own ?

domjanik: According to Captify's documentation we got we should set this by ourself :(

PR #1641, opened 2022-12-08
ADEN-12504 Add proper src values

nandy-andy:


src/platforms/gamefaqs/ad-context.ts
```diff
@@ -6,4 +6,5 @@ export const basicContext = {
 		property: 'gamefaqs',
 	},
 	slots: {},
+	src: ['gamefaq'],
```
szymon-nawrocki, line 9: Suggestion according to:

https://fandom.atlassian.net/browse/ADEN-12504?focusedCommentId=595936

szymon-nawrocki, line 9:
```suggestion
	src: ['gamefaq'],
```
```suggestion
	src: ['gamefaqs'],
```

szymon-nawrocki, line 9:
```suggestion
	src: ['gamefaqs'],
```

src/platforms/comicvine/comicvine-platform.ts
```diff
@@ -13,7 +19,9 @@ export class ComicvinePlatform {
 	execute(): void {
 		this.pipeline.add(
 			() => context.extend(basicContext),
-			// TODO: we need a CMP step here, so we won't call for ads unless we have a clear idea of the privacy policy of a visitor
+			() => context.set('custom.dfpId', this.shouldSwitchGamToRV() ? 22309610186 : 5441),
```
szymon-nawrocki, line 22: What do you think about moving lines 21-24 to external class like `ComicvineBaseContextSetup`, so the pipeline of platform setup is nice and clean (like for other platforms)?

szymon-nawrocki: Plus we can go even further and move this code to some sort of `/shared-news-and-ratings` directory,  so the same code is not repeated in 7 places?

szymon-nawrocki: You can add it to same shared directory as I created for TargetingSetup here:

https://github.com/Wikia/ad-engine/blob/cb6c8a47e8efa8803544e3db2e6b4b260329ed2f/src/platforms/shared-news-and-ratings/context/targeting/news-and-ratings-targeting.setup.ts

And then after refinement meeting where we'll talk about sharing the logic, we can move/refactor this logic.

What do you think?

szymon-nawrocki: Or we can just move it later, but I think it will be worth at least to separate setting the context to separate class. And it will be worth to add a TODO comment to move it to some sort of shared directory, or add it to AC of the _ADEN-12446_ ticket

nandy-andy: Hm, I don't know. I thought we'd look at the code holistically and then clean up such parts while working on the ticket you've mentioned 🤔

nandy-andy: Yeah, since I don't have `platforms/shared-news-and-ratings` in my branch yet, I'll leave it for ADEN-12446 after the targeting is merged and we've discussed the topic of shared functionalities.

nandy-andy: I updated the ticket description, so we won't miss it ;)

PR #1649, opened 2022-12-12
ADEN-12516 Add instant-config support to NnR bundles

nandy-andy:


src/platforms/comicvine/comicvine-platform.ts
```diff
@@ -22,7 +18,7 @@ export class ComicvinePlatform {
 			() => context.set('custom.dfpId', this.shouldSwitchGamToRV() ? 22309610186 : 5441),
 			() => context.set('src', this.shouldSwitchSrcToTest() ? ['test'] : context.get('src')),
 			() => bootstrapAndGetConsent(),
-			// TODO: to decide if we want to call instant-config service for the first releases?
+			InstantConfigSetup,
```
szymon-nawrocki, line 21: In every other (non RV) platform this is done by calling it in parralel with `bootstrapAndGetConsent()` like `parallel(InstantConfigSetup, () => bootstrapAndGetConsent())`.
Probably it was done for a reason, and it would be nice to do it in the same way in every platform setup to keep things unified.

example:
https://github.com/Wikia/ad-engine/blob/3a171ba8fca8c1ac847e12bcf6a4aae47a2b5180/src/platforms/sports/sports-platform.ts#L48

src/platforms/gamefaqs/gamefaqs-platform.ts
```diff
@@ -24,12 +30,12 @@ export class GamefaqsPlatform {
...
 			() => context.set('custom.dfpId', this.shouldSwitchGamToRV() ? 22309610186 : 5441),
 			() => context.set('src', this.shouldSwitchSrcToTest() ? ['test'] : context.get('src')),
-			// TODO: to decide if we want to call instant-config service for the first releases?
 			() => bootstrapAndGetConsent(),
+			InstantConfigSetup,
 			GamefaqsTargetingSetup,
 			GamefaqsSlotsContextSetup,
 			GamefaqsDynamicSlotsSetup,
-			// TODO: add targeting setup once we have idea of page-level and slot-level targeting
+			BiddersStateSetup,
```
szymon-nawrocki, line 38: Should we call this class if N&R platforms have different bidders / bidders ids?
Shouldn't we make separate `BidderStateSetup` for N&R platforms?

src/platforms/comicvine/comicvine-platform.ts
nandy-andy: Correct! But this time we don't want it to be in parallel because we don't have geo cookie set from our varnishes and then the tracking-opt-in wrapper in `bootstrapAndGetConsent()` will set it. However, it requires synchronous execution 🤷‍♂️ Once we have cookies set on varnishes we can unify it. (//cc @domjanik)

src/platforms/gamefaqs/gamefaqs-platform.ts
nandy-andy: Good idea! 👍 I'll put it in [the follow-up](https://fandom.atlassian.net/browse/ADEN-12510) once we decide on the ultimate bidders setup. I assume at the end we'll have the same bidders setup across all of our sites. 

Right now, reusing this step is useful for testing prebid integration with our test bidder "Wikia Adapter".

src/platforms/comicvine/comicvine-platform.ts
szymon-nawrocki: Okay thanks for the explanation, we can add some TODO comment in the code to not forget about it

szymon-nawrocki (approved):

PR #1654, opened 2022-12-15
ADEN-12535 Quick-fix for controls issue on mobile

nandy-andy:


nandy-andy:
Not needed because of https://github.com/Wikia/jwplayer-fandom/pull/138

PR #1656, opened 2022-12-19
ADEN-12531 - Added Identity, LiveRamp and LiveIntent to N&R platforms

domjanik:


jakubbanas (approved):

src/platforms/gamefaqs/gamefaqs-platform.ts
```diff
@@ -23,6 +28,7 @@ export class GamefaqsPlatform {
 			// once we have Geo cookie set on varnishes we can parallel bootstrapAndGetConsent and InstantConfigSetup
 			() => bootstrapAndGetConsent(),
 			InstantConfigSetup,
+			BaseContextSetup,
```
nandy-andy, line 31: Could we fill a follow-up ticket to revert it and replace with a smaller steps? `BaseContextSetup` is a big bag of unnecessary settings and operations for NnR for this moment.

We're not obliged to remove ads on Steam platforms for NnR it was an old deal for Fandom Communities: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/setup/base-context.setup.ts#L38

We don't have `top_leaderboard` yet on NnR, so we don't want to run travers DOM and look for them: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/setup/base-context.setup.ts#L41-L45

We most likely won't be getting the information if user is logged-in based on the MediaWiki global variable: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/setup/base-context.setup.ts#L56

We don't plan to add the rotator https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/setup/base-context.setup.ts#L108

WAD has been already set up in a separate setup step for GameFAQs: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/setup/base-context.setup.ts#L137-L147 (see https://fandom.atlassian.net/browse/ADEN-12557)

Let's just create separate steps for setting up identity, LiveRamp and LiveIntent options. At the end we'll have a smaller, cleaner setup steps that we could reuse in our main applications and remove the "bag of everything and nothing". I guess there is a ticket in developer improvements epic to clean up the base context setup.

PR #1662, opened 2022-12-22
ADEN-AdEngine-2023 New Year's major upgrade pack

dmnsobczak:


spec/ad-products/templates/interface/listeners/player-overlay-click-listener.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { expect } from 'chai';
 import { createSandbox } from 'sinon';
 
-import { onVideoOverlayClick } from '../../../../../src/ad-products/templates/interface/listeners/player-overlay-click-listener';
+import { onVideoOverlayClick } from '@wikia//ad-products/templates/interface/listeners/player-overlay-click-listener';
```
preludek, line 4: import { onVideoOverlayClick } from '@wikia/ad-products/templates/interface/listeners/player-overlay-click-listener';
```suggestion
import { onVideoOverlayClick } from '@wikia//ad-products/templates/interface/listeners/player-overlay-click-listener';
```

spec/ad-products/templates/uap/ui/ui-template.spec.ts
```diff
@@ -2,8 +2,8 @@ import { selectTemplate } from '@wikia/ad-products';
 import { UapVideoSettings } from '@wikia/ad-products/templates/uap/uap-video-settings';
 import { expect } from 'chai';
 import { createSandbox } from 'sinon';
-import { CloseButton } from '../../../../../src/ad-products/templates/interface/video/close-button';
-import { PlayerOverlay } from '../../../../../src/ad-products/templates/interface/video/player-overlay';
+import { CloseButton } from '@wikia//ad-products/templates/interface/video/close-button';
```
preludek, line 5:
```suggestion
import { CloseButton } from '@wikia/ad-products/templates/interface/video/close-button';
```

.nycrc
```diff
@@ -1,26 +1,28 @@
 {
   "check-coverage": true,
-  "lines": 50,
-  "functions": 50,
-  "branches": 50,
-  "statements": 50,
+  "lines": 55,
```
nandy-andy, line 3: Why not the old 50? Here and in the next 3 lines below :D

package.json
```diff
@@ -3,82 +3,63 @@
...
-	"sideEffects": [
-		"*.scss"
-	],
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/Wikia/AdEngine"
 	},
 	"dependencies": {
-		"@babel/runtime-corejs2": "7.18.9",
+		"@babel/runtime-corejs2": "7.20.7",
```
nandy-andy, line 11: Do we still need it? I thought you mentioned removing babel and above there is a `babel-app.config.js` being removed 🤔

package.json
```diff
@@ -90,9 +71,9 @@
 		"prepare": "husky install",
 		"prebuild": "rm -rf dist",
 		"preversion": "./maintenance/preversion.sh",
-		"test": "instant-mocha --full-trace './spec/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.common.js",
+		"test": "instant-mocha './spec/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.common.js",
 		"test:watch": "npm run test -- -w",
-		"test-coverage": "nyc instant-mocha './spec/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.common.coverage.js"
+		"test-coverage": "nyc instant-mocha './src/!(platforms)*/**/*.ts' './src/platforms/shared/**/*.ts' './spec/**/*.spec.ts' -r jsdom-global/register --webpack-config webpack.coverage.js"
```
nandy-andy, line 76: What are the paths we're passing here? They seem similar except the first one seem like we want to exclude any `*.ts` files inside `src/platforms/` and the other one seem like we want to include any `*.ts` files within `src/platforms/shared` and then include all `spec/**/*.spec.ts` files with unit tests 🤔

spec/spec-utils/html-element.stub.ts
```diff
@@ -1,10 +0,0 @@
-import { SinonSandbox, SinonStubbedInstance } from 'sinon';
-
-// @TODO: remove if not needed
```
nandy-andy, line 3: I've used it in one [recently added test](https://github.com/Wikia/ad-engine/pull/1674/files#diff-5340cb2ea465ad9c489f2f4153318cf2faf8253639584140aee8b770e6356db8R7) 😊 But if we don't use it anywhere else, we can 🔪 it, I guess 🤔

src/ad-products/video/jwplayer/streams/jwplayer-stream-stateless.ts
```diff
@@ -108,9 +107,7 @@ function onlyOncePerVideo<T>(jwplayer: JWPlayer): RxJsOperator<T, T> {
 				event,
 				playlistItem: jwplayer.getPlaylistItem() || ({} as JWPlayerListItem),
 			})),
-			distinctUntilChanged(
-				(a: JwpState, b: JwpState) => a.playlistItem.mediaid === b.playlistItem.mediaid,
-			),
+			distinctUntilChanged((a, b) => a.playlistItem.mediaid === b.playlistItem.mediaid),
```
nandy-andy, line 110: 🙀 you removed typing! Why?! ;)

src/core/services/templates-registry/template-registry.ts
```diff
@@ -45,6 +45,7 @@ export class TemplateRegistry {
 			StateHandlerTypesDict,
 			initialStateKey,
 			emitter$,
+			// @ts-ignore "this context fix"
```
nandy-andy, line 48: Could you elaborate more why we need the `@ts-ignore` here? Is it waiting for changes with the god object (`context`)?

nandy-andy (changes requested):
Looks good. Let's add follow-up tickets for all the TODOs.

.nycrc
dmnsobczak: I included all files in code coverage report, not only those imported in existing test files. This resulted in a visible drop in real code coverage :< So I set those limit below current readings and in the future we will use them to enforce unit testing 🧐

package.json
dmnsobczak: I want to import everything inside `/src/`, but not platform directories, but I cannot list `excludes` here :/ So I reversed it :D

dmnsobczak: Also: I added it to enforce `instabul-loader` to load all `.ts` files.

spec/spec-utils/html-element.stub.ts
dmnsobczak: Is it (your test) on master? 🤔

src/ad-products/video/jwplayer/streams/jwplayer-stream-stateless.ts
dmnsobczak: Because `a` and `b` are not actually `JwpState` but `Partial< JwpState>` and fixing this mapping would require more, potentially regressive changes. Also compiler understand what `a` and `b` should look like, so I decided that it's not worth it :D

spec/spec-utils/html-element.stub.ts
nandy-andy: Not yet on master/dev but soon I hope ;)

src/core/services/templates-registry/template-registry.ts
dmnsobczak: I think there is a bug in TypeScript and it don't properly understand this type creation:
`export type TemplateDependency<T = any> = Parameters<Container['bind']>[0];`
This results in a TS error that I don't know how to solve :< I'll try to get rid of this `@ts-ignore` later 🤞

spec/spec-utils/html-element.stub.ts
dmnsobczak: Ok! Unit tests won't pass in that case so I will restore it during this branch merge to master :) In this universe this file is redundant 🤔

src/core/services/templates-registry/template-registry.ts
dmnsobczak: I don't know how :< I'll leave it for the future 🤞

nandy-andy (approved):

lnawrotek (approved):

PR #1663, opened 2022-12-23
ADEN-12557 Add AB features to GameFAQs, Gamespot and TVGuide bundles

nandy-andy:


dmnsobczak (approved):

src/core/services/bab-detection.ts
```diff
@@ -24,10 +24,15 @@ class BabDetection {
...
 
 	async run(): Promise<boolean> {
-		const isBabDetected: boolean = await this.checkBlocking();
+		let isBabDetected: boolean = await this.checkBlocking();
 
 		utils.logger(logGroup, 'BAB detection, AB detected:', isBabDetected);
 
+		if (!isBabDetected) {
+			isBabDetected = await this.checkDomainBlocking();
+			utils.logger(logGroup, 'GAM domain blocking detection - detected:', isBabDetected);
```
szymon-nawrocki, line 33: Or it will be even better to move this log to the method `checkDomainBlocking()`, this way we can call in line 27 `const isBabDetected: boolean = await this.checkBlocking() || await this.checkDomainBlocking()` and the log in line 29 can stay as it is.

src/core/services/bab-detection.ts
```diff
@@ -24,10 +24,15 @@ class BabDetection {
 	}
 
 	async run(): Promise<boolean> {
-		const isBabDetected: boolean = await this.checkBlocking();
+		let isBabDetected: boolean = await this.checkBlocking();
 
 		utils.logger(logGroup, 'BAB detection, AB detected:', isBabDetected);
```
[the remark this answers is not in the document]
nandy-andy: Why do you think it's misleading? It clearly states it's the BAB detection that succeed or failed. If it fails then you'll receive another log message about domain blocking detection (not BAB detection) succeed or failed 🤔

src/core/services/bab-detection.ts
nandy-andy: Won't this way make unnecessary HTTP request from `this.checkDomainBlocking()` when `this.checkBlocking()` returns `true`? 🤔 The logic here is simple: if BAB detected ad block we trust it and we don't make additional HTTP request. If BAB fails, let's check domain blocking.

src/core/services/bab-detection.ts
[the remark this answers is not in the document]
szymon-nawrocki: Okay, I think I mistaken BAB detection with ad block detection. Maybe because result of `checkDomainBlocking()` is assigned to variable `isBabDetected`. But this is just a nitpicking

src/core/services/bab-detection.ts
szymon-nawrocki: JS won't execute second part if the first is true, but this will require moving the log message from line 33, so maybe it will be better to leave it as it is

src/core/services/bab-detection.ts
```diff
@@ -69,14 +74,39 @@ class BabDetection {
 		});
 	}
 
+	public doesBlockAdBlockExist() {
```
szymon-nawrocki, line 77: How about just `blockAdBlockExist()` so more literally in English and easier to read?
(`if (doesBlockAdBlockExists)` vs `if (blockAdBlockExists)`)

src/platforms/gamefaqs/gamefaqs-platform.ts
```diff
@@ -28,6 +29,7 @@ export class GamefaqsPlatform {
 			// once we have Geo cookie set on varnishes we can parallel bootstrapAndGetConsent and InstantConfigSetup
 			() => bootstrapAndGetConsent(),
 			InstantConfigSetup,
+			GameFaqsWadSetup,
```
szymon-nawrocki, line 32: It would be nice to change the class name to GamefaqsWadSetup (with small "f") too keep the class names consistent.

spec/platforms/shared/services/wad-runner.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+		babDetectionStub.isEnabled.returns(true);
+		babDetectionStub.run.returns(babDetectionRunStub);
+
+		const wadRunner = new WadRunner(babDetectionStub, onDetected);
+		await wadRunner.call();
+
+		expect(onDetected.called).to.equal(true);
+	});
+
+	function createDetectionRunStub(returnValue: boolean) {
```
szymon-nawrocki, line 42: It would be nice to move the function outside `describe` block so the tests are not cluttered with unnecessary things

spec/core/services/bab-detection.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+		createCreateBlockAdBlockStub(detector);
+		sinon.stub(detector, 'checkDomainBlocking').returns(createCheckDomainBlockingStub(true));
+
+		return await detector.run().then((detected) => {
+			expect(detected).to.be.true;
+		});
+	});
+
+	function createCreateBlockAdBlockStub(detector) {
+		// @ts-ignore just for tests
```
szymon-nawrocki, line 79: This `// @ts-ignore` is unnecessary

spec/ad-services/bt-rec/bt-rec.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { expect } from 'chai';
+import { createSandbox } from 'sinon';
+
+import { context, utils } from '../../../src/core';
+import { btRec } from '../../../src/ad-services/bt-rec';
```
szymon-nawrocki, line 5: Import can be shortened with:
`import { btRec } from '@wikia/ad-services';`

spec/core/services/bab-detection.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { BabDetection } from '../../../src/core/services/bab-detection';
```
szymon-nawrocki, line 4: Import can be shortened with:
`import { BabDetection } from '@wikia/core';`

szymon-nawrocki (dismissed):
👌

szymon-nawrocki (approved):

szymon-nawrocki (dismissed):

szymon-nawrocki (approved):

PR #1670, opened 2022-12-27
ADEN-12573

lnawrotek:


nandy-andy (approved):

src/platforms/shared/bootstrap.ts
```diff
@@ -4,6 +4,8 @@ import { trackingOptInWrapper } from './consent/tracking-opt-in-wrapper';
 export async function bootstrapAndGetConsent(): Promise<void> {
 	logVersion();
 
+	utils.geoService.setUpGeoData();
+
 	await trackingOptInWrapper.init();
 
 	utils.geoService.setUpGeoData();
```
nandy-andy, line 11: Maybe it's worth leaving a comment here that 2nd time is the charm ;) Meaning we call it second time because `tracking-opt-in` will set the cookie if it won't find it.

dmnsobczak (approved):

PR #1676, opened 2022-12-30
ADEN-12232 - Added CrossDomain storage to Identity Service

domjanik:


src/ad-services/user-identity/adms-identity-repository/index.ts
```diff
@@ -1,13 +1,18 @@
 import { admsService } from './adms-service';
 import { ActionType, IdentityAction } from './adms-actions';
 import { IdentityRepositoryInterface } from '../identity-repositories';
-import { utils } from '@ad-engine/core';
+import { UniversalStorage, utils } from '@ad-engine/core';
 import { UserIdentity } from '../';
 
 class AdmsIdentityRepository implements IdentityRepositoryInterface {
+	storage = new UniversalStorage();
```
nandy-andy, line 8: We could pass it in the constructor to make unit testing it easier :)

src/ad-services/user-identity/global-identity-repository/global-identity-provider/index.ts
```diff
@@ -0,0 +1,46 @@
+import { utils } from '@ad-engine/core';
+import {
+	CrossDomainClient,
+	IframeMessage,
+	Messages,
+} from '@fandom-frontend/cross-domain-storage/dist';
```
nandy-andy, line 6: nit-picking: we could maybe create an alias and get rid of `/dist` 🤔

src/ad-services/user-identity/global-identity-repository/global-identity-provider/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+import { GlobalIdentityProviderStatus } from './GlobalIdentityProviderStatus';
+
+export class GlobalIdentityProvider {
+	public client: CrossDomainClient;
+	private logGroup = 'GlobalIdentityProvider';
+	public status = GlobalIdentityProviderStatus.INITIALIZING;
+
+	private isFandomUrl(): boolean {
+		return window.location.href.includes('.fandom.com');
```
nandy-andy, line 16: Consider using `window.location.host` 🤔

src/ad-services/user-identity/global-identity-repository/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+import { Messages } from '@fandom-frontend/cross-domain-storage/dist';
+import { UniversalStorage, utils } from '@ad-engine/core';
+import { GlobalIdentityProvider } from './global-identity-provider';
+import { GlobalIdentityProviderStatus } from './global-identity-provider/GlobalIdentityProviderStatus';
+
+class GlobalIdentityRepository implements IdentityRepositoryInterface {
+	logGroup = 'GlobalIdentityRepository';
+	storage = new UniversalStorage();
+	client = new GlobalIdentityProvider();
```
nandy-andy, line 11: What do you think of renaming `client` to `identityProvider`? It's more descriptive as `client` can mean many different things 🤔

nandy-andy (dismissed):

src/ad-services/user-identity/global-identity-repository/global-identity-provider/index.ts
domjanik: I've tried to do that, however I've failed - It seems that there are some problems with fandom-frontend configuration. When I've checked in their repo, all of their projects that are loaded as dependencies seems to be loaded with /dist :/ If you know about package without that in fandom-frontend please, let me know. Otherwise I would leave it as it is for now.

jakubbanas (changes requested):
Looks good overall, but I believe `package-lock` should not be removed.

domjanik:
> Looks good overall, but I believe `package-lock` should not be removed.

Fixed

jakubbanas (approved):

PR #1677, opened 2022-12-30
ADEN-12563 Add partners to N&R sites

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12563


src/platforms/news-and-ratings/metacritic/ad-context.ts
```diff
@@ -6,6 +6,18 @@ export const basicContext = {
...
 	services: {
+		confiant: {
+			enabled: false,
+			propertyId: 'd-aIf3ibf0cYxCLB1HTWfBQOFEA',
+		},
+		durationMedia: {
+			enabled: false,
+			libraryUrl: '//tag.durationmedia.net/sites/10651/dm.js',
+		},
+		iasPublisherOptimization: {
```
nandy-andy, line 17: I wonder if we could move `confiant` and `durationMedia` settings to shared ones 🤔 Maybe even IAS with an exception of slots list. On the other hand this way it's less complicated. @dmnsobczak @domjanik what do you think?

nandy-andy: Damian found a nice solution in 🐟 changes, so let's go with this now and once we'll have conflicts to resolve we'll change it 👌 https://github.com/Wikia/ad-engine/blob/04d3068c1b8d4e8109315f037941b938f7e2e64f/src/platforms/f2/modes/f2-ads-mode.service.ts#L26

nandy-andy (dismissed):

nandy-andy (approved):

nandy-andy:
No testing #wżtr We'll check the services later while testing the rest N&R sites integration.

PR #1689, opened 2023-01-11
ADEN-12620 Fixing top-ads-container collapsing and collapse proxy for IX tags

dmnsobczak:


src/platforms/shared/utils/placeholder-service-helper.ts
```diff
@@ -46,6 +46,10 @@ export class PlaceholderServiceHelper {
 	hidePlaceholder = (placeholder: HTMLElement): void => {
 		if (this.shouldHidePlaceholder) {
 			placeholder.classList.add('hide');
+
+			if (placeholder?.parentElement.className.includes('-ads-container')) {
```
nandy-andy, line 50: And now `hidePlaceholder()` not only hides the placeholder but also its parent 🙈

src/communication/event-types.ts
```diff
@@ -293,6 +293,11 @@ export const eventsRepository: Dictionary<EventOptions> = {
 		name: 'Ad intervention',
 		payload: props<AdIntervention>(),
 	},
+	GAM_AD_DELAYED_COLLAPSE: {
```
nandy-andy, line 296: Why can't it be just `GAM_AD_COLLAPSE`? 🤔

src/core/models/ad-slot.ts
```diff
@@ -472,6 +472,20 @@ export class AdSlot {
...
+
+		this.setupDelayedCollapse();
+	}
+
+	private setupDelayedCollapse() {
+		communicationService.on(
+			eventsRepository.GAM_AD_DELAYED_COLLAPSE,
+			(payload) => {
+				if (payload.source.includes(this.getSlotName())) {
+					this.collapse();
```
nandy-andy, line 484: I wonder if `this.collapse()` is enough here. We had the agreement that we collapse/hide slots only when it's a direct campaign and "forced collapse" happen -- then the gain from direct ad is bigger than SEO loss from content shifts. Do we need to introduce a new event instead of using "forced collapse" in the creatives? 🤔

dmnsobczak: This is not a force_collapse, but regular collapse: this will leave the placeholder and just hide the label and set proper status in slots repository :)

src/communication/event-types.ts
dmnsobczak: Because initially slot gets `success` and then `collapse` when IX tag won't get an ad.

nandy-andy: This is how forced collapse work, isn't it? 🤔 We got a success but in the creative template there is "a signal" that makes AdEngine forcefully collapse the slot ;)

src/platforms/shared/utils/placeholder-service-helper.ts
dmnsobczak: Technically it's still placeholder 🙈 It's just consist of two divs. Without it the black div without content was left in the DOM and it was visible due to a margins applied.

src/core/models/ad-slot.ts
nandy-andy: Oh, darn ;-)

src/communication/event-types.ts
dmnsobczak: Yes, but no :D It's not force collapse, it's just collapse. In force_collapse logic we load a creative to close the slot using collapsinator. Here we actually want to render the ad, but IX tag is not giving us any :(

nandy-andy: Hm, yeah, a simple behaviour became really complicated 🙈 One more idea from the "naming" category that comes to my mind is to get rid of "collapse" word in the event name then 🤔 Something more generic like `GAM_EMPTY` and then handle it in the code with a collapse, force-collapse, flowers however we want to...

src/platforms/shared/utils/placeholder-service-helper.ts
nandy-andy: Yeah, I'm not buyin' it ;-) Placeholder is only one per slot: it holds the place for an ad. The part with `'-ads-container'` is a custom wrapper of ours, so all the things are styled properly and look nice. That's my PoV :D

dmnsobczak: I understand, but this is how it is xD This wrapped contains only placeholder and nothing else and placeholder cannot exist without a wrapper as well as wrapper cannot exist without a placeholder. 
If I replace 2 divs where one has black background with one div and more complex margin settings and/or `:after` elements then:
1. It will be hard to handle UAP properly
2. It will be one placeholder in your definition
But actually it won't change anything from product perspective. This placeholder is bugged: Seba is adding `display: none` attribute to when there is a campaign equipped with `force_collapse` signal. With this PR Seba's hack is no longer needed.
Also, parent bubbling works similar way in UAP's DOM Manager:
https://github.com/Wikia/ad-engine/blob/dev/src/platforms/shared/templates/helpers/uap-dom-manager.ts#L76

nandy-andy: I know how it is and I'm voting for changing it so the code is cleaner :D What holds us from having two simple and clear methods: `hidePlaceholder()` and `hideWrapper()`? 🤔 If we have ugly code in `uap-dom-manager.ts`, let's clean it up and if the placeholder is buggy let's also fix it ;)

src/platforms/f2/styles.scss
```diff
@@ -232,12 +111,136 @@
...
 #bottom_leaderboard {
 	@include uap-advertisement-label();
 }
 
-@include f2-styles();
+.gpt-ad {
+	text-align: center;
+
+	&.hide,
+	& .hide {
```
preludek, line 141: that looks odd 🙈

src/platforms/f2/styles.scss
```diff
@@ -232,12 +111,136 @@
...
+		.feed-layout__right-rail {
+			top: 0 !important;
+		}
+	}
+}
+
+.feed-item .gpt-ad {
+	padding: 24px 24px 30px;
+	position: sticky;
+	top: 55px;
```
preludek, line 184: If that's not an `expert value` it would be nice to make some descriptive variable names out of such values.

src/platforms/f2/styles.scss
dmnsobczak: This styles are moved from place to place - I didn't modify them 🤷

src/platforms/f2/styles.scss
dmnsobczak: This styles are moved from place to place - I didn't modify them 🤷

src/platforms/shared/utils/placeholder-service-helper.ts
dmnsobczak: I'm not a big fan of extracting one liners to a separate methods when they are self explaining and their functionality fits the original method 🤔 I was hoping for @rafalgiemza to give deciding opinion :D

nandy-andy: I understand. I suggest to take a look at it from a different perspective. A perspective of someone, who just takes a glance at the code and does not read line by line of each method. She sees `hidePlaceholder()` and decides to use it thinking it'll hide just the placeholder. To her surprise there is something more going on the parent element after reusing the method, so she needs to invest more time to go back and indeed read the method to know what it does. The idea behind clean code is to the have small methods doing one thing with descriptive names.

src/platforms/f2/styles.scss
preludek: Be a good scout lad 😸

src/platforms/shared/utils/placeholder-service-helper.ts
dmnsobczak: Rafal didn't reply :( So I give up: change pushed :D

nandy-andy (dismissed):

nandy-andy (approved):

PR #1690, opened 2023-01-11
ADEN-12623 Add Appnexus to News&Ratings

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12623


src/ad-bidders/prebid/adapters/appnexus.ts
```diff
@@ -61,6 +61,18 @@ export class Appnexus extends PrebidAdapter {
 	}
 
 	getBids(code, { placementId, position = 'mobile' }: PrebidAdSlotConfig): PrebidBid[] {
+		if (Array.isArray(placementId) && placementId.length > 0) {
```
nandy-andy, line 64: Let's create a follow-up ticket to fix mess in this bidder 😊

nandy-andy (approved):

src/ad-bidders/prebid/adapters/appnexus.ts
szymon-nawrocki: Great idea!

PR #1696, opened 2023-01-16
ADEN-12624 Add Megnite (Rubicon) to News&Ratings bundles

nandy-andy:


src/platforms/news-and-ratings/gamespot/bidders/rubicon-display.ts
```diff
@@ -0,0 +1,217 @@
...
+		},
+		'mobile-highimpact-plus': {
+			sizes: [[300, 250]],
+			siteId: '462824',
+			zoneId: '2719798',
+		},
+	};
+
+	return {
+		enabled: false,
```
mfarul, line 213: what does it mean here? it will not be enabled by default and then override via ICBM or some else variable?

mfarul (approved):

src/platforms/news-and-ratings/gamespot/bidders/rubicon-display.ts
nandy-andy: Yep, correct. It's unified with how it works now on FCP:
* https://github.com/Wikia/ad-engine/blob/dev/src/platforms/ucp-mobile/bidders/prebid/rubicon-display.ts#L3
* https://github.com/Wikia/ad-engine/blob/dev/src/platforms/ucp-desktop/bidders/prebid/rubicon-display.ts#L3
* https://github.com/Wikia/ad-engine/blob/dev/src/platforms/sports/bidders/prebid/rubicon-muthead.ts#L6

We can change it in the future if we decide it makes only a little bit of sense and adds unnecessary bits to our bundles 😊

PR #1702, opened 2023-01-17
ADEN-12649 Port ad unit structure to N&R bundles

nandy-andy:


src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
```diff
@@ -44,4 +48,25 @@ export class NewsAndRatingsBaseContextSetup implements DiProcess {
...
 	}
+
+	private getPagePathFromMetatags(): string {
+		const dataSettingsJson = this.getDataSettingsFromMetaTag();
+
+		if (!dataSettingsJson) {
+			return '';
+		}
+
+		const pagePath = JSON.parse(dataSettingsJson)?.unit_name?.split('/').pop();
```
szymon-nawrocki, line 59: Have you tried taking `pagePath` from `window.utag_data`? There is a key-val `siteSection` where the page path is defined. This way it will be also available on TVGuide and Metacritic, and you won't have to do the parsing of JSONs

nandy-andy: I can try if it's fast enough 🤞

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
```diff
@@ -33,4 +37,44 @@ export class NewsAndRatingsBaseContextSetup implements DiProcess {
 	private shouldSwitchSrcToTest() {
 		return utils.queryString.get('switch_src_to_test') === '1';
 	}
+
+	private getPagePath(): string {
+		let pagePath,
```
szymon-nawrocki, line 42: I like the solution. What do you think about writing it as, to make the logic more readible/clear?

`const pagePath = this.getPagePathFromMetaTag || this.getPagePathFromUtagData`

nandy-andy: Discussed on slack: such change brakes one of the unit tests. Let's keep it as most likely we'll rebuild the ad unit structure within first weeks after the migration.

szymon-nawrocki (dismissed):

szymon-nawrocki (approved):

PR #1716, opened 2023-01-24
ADEN-12567 Map N&R targeting keys

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12567


src/platforms/news-and-ratings/shared/setup/context/targeting/phoenix-sites-and-gamefaqs-targeting.setup.ts
```diff
@@ -0,0 +1,51 @@
...
+		return targeting;
+	}
+
+	getMetadataTargetingParams(): TargetingParams {
+		const dataSettingsElement = document.head
+			.querySelector('[id=ad-settings]')
+			?.getAttribute('data-settings');
+
+		if (dataSettingsElement) {
+			return JSON.parse(dataSettingsElement)?.target_params;
```
preludek, line 46: I'm guessing `?` is here in case JSON.parse fails. If yes, I think this will not work as intended and produce an error. I'd recommend using a try/ catch here and returning value accordingly.

src/platforms/news-and-ratings/tvguide/setup/context/targeting/tvguide-targeting.setup.ts
```diff
@@ -2,6 +2,7 @@ import { context, DiProcess } from '@wikia/ad-engine';
 
 export class TvGuideTargetingSetup implements DiProcess {
 	execute(): void {
+		// TODO: Map key 'network' to 'tv' (ADEN-12567)
```
preludek, line 5: 😞

src/platforms/news-and-ratings/comicvine/comicvine-platform.ts
```diff
@@ -25,8 +26,9 @@ export class ComicvinePlatform {
 			InstantConfigSetup,
 			NewsAndRatingsBaseContextSetup,
 			NewsAndRatingsWadSetup,
-			ComicvineTargetingSetup,
 			NewsAndRatingsTargetingSetup,
+			PhoenixSitesAndGamefaqsTargetingSetup,
```
preludek, line 30: Is this a part of comicvine platform setup? 🤔

src/platforms/news-and-ratings/gamefaqs/gamefaqs-platform.ts
```diff
@@ -26,8 +27,9 @@ export class GamefaqsPlatform {
 			InstantConfigSetup,
 			NewsAndRatingsBaseContextSetup,
 			NewsAndRatingsWadSetup,
-			GamefaqsTargetingSetup,
 			NewsAndRatingsTargetingSetup,
+			PhoenixSitesAndGamefaqsTargetingSetup,
+			GamefaqsTargetingSetup,
```
preludek, line 32: It seems like there's like two GamefaqsTargetingSetups, or at least one and a half 😄

preludek (changes requested):
Left some comments. Also the tickets AC calls for unit tests to be added.

src/platforms/news-and-ratings/shared/setup/context/targeting/phoenix-sites-and-gamefaqs-targeting.setup.ts
```diff
@@ -0,0 +1,51 @@
+import { context, DiProcess } from '@wikia/ad-engine';
+import { TargetingParams } from './interfaces/targeting-params';
+
+export class PhoenixSitesAndGamefaqsTargetingSetup implements DiProcess {
```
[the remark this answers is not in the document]
szymon-nawrocki: So rename it to `GamefaqsGamespotGiantbombComicvineTargetingSetup` ? That does not look good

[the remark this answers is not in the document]
szymon-nawrocki: I'd rather rename it to `PhoenixBasedSitesAndGamefaqsTargetingSetup` or just `PhoenixBasedSitesTargetingSetup` Then we can see from the usage that Gamefaqs is using the same TargetingSetup as Phoenix based sites

src/platforms/news-and-ratings/shared/setup/context/targeting/phoenix-sites-and-gamefaqs-targeting.setup.ts
szymon-nawrocki: Good catch!

src/platforms/news-and-ratings/comicvine/comicvine-platform.ts
szymon-nawrocki: yes, Comicvine is Phoenix based site

src/platforms/news-and-ratings/gamefaqs/gamefaqs-platform.ts
szymon-nawrocki: I don't understand, this is one `GamefaqsTargetingSetups` that shares some of the targeting parts from:
* phoenix based sites
* all news and ratings sites

src/platforms/news-and-ratings/shared/setup/context/targeting/phoenix-sites-and-gamefaqs-targeting.setup.ts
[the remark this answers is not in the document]
preludek: I'm thinking if this should be in the shared directory anyway. It's not being shared between all the n&r platforms hence I'd argue this code should be in each TargetingSetup respectively. It's not a problem bundle size-wise since we are generating a bundle based on a platform. It would be pretty straightforward which platforms use these different keys, and finally, we wouldn't have to initialize multiple targeting setups. (for example for Gamefaqs we have 3):
```
   NewsAndRatingsTargetingSetup,
   PhoenixSitesAndGamefaqsTargetingSetup,
   GamefaqsTargetingSetup,
```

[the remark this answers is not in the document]
szymon-nawrocki: `PhoenixBasedSitesTargetingSetup` is shared between 4/6 platforms so for me it shared :)
I don't want to put the exactly the same code to 4 different places  according to the DRY rule.
Plus this class will be removed, and the code will be moved to `NewsAndRatingsTargetingSetup` once Metacritic and TVGuide will implement `adtags` metadata object that contains all targeting params on their sites

preludek (approved):

preludek (approved):

PR #1671, opened 2022-12-27
ADEN-12519 Get ICBM config per country

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12519

PR #1719, opened 2023-01-25
ADEN-12656 Adding BFAA on Gamespot

dmnsobczak:


src/platforms/news-and-ratings/gamespot/gamespot-platform.ts
```diff
@@ -20,6 +26,7 @@ export class GameSpotPlatform {
 	execute(): void {
 		this.pipeline.add(
 			() => context.extend(basicContext),
+			() => context.set('state.isMobile', getDeviceMode() === 'mobile'),
```
szymon-nawrocki, line 29: It would be nice, to refactor this `isDesktop` variable: https://github.com/Wikia/ad-engine/blob/31aaaa3a7a045e8b42a1414235f376157fe31885/src/platforms/news-and-ratings/gamespot/setup/context/prebid/gamespot-prebid-config.setup.ts#L23 to use one you set here.

Plus it would be even nicer to apply this later to the rest of the platforms ;-)

src/platforms/news-and-ratings/gamespot/setup/context/targeting/gamespot-targeting.setup.ts
```diff
@@ -4,6 +4,9 @@ export class GamespotTargetingSetup implements DiProcess {
 	execute(): void {
 		const targeting = {
 			s0: 'gaming',
+			uap: 'none',
```
szymon-nawrocki, line 7: If this will be shared (set up the same) across all N&R platforms, it would be nice to add it to news-and-ratings-targeting.setup.ts

dmnsobczak: Done!

src/platforms/news-and-ratings/gamespot/gamespot-platform.ts
dmnsobczak: Done!

src/core/utils/client.ts
```diff
@@ -77,6 +78,10 @@ class Client {
 		return 'desktop';
 	}
 
+	getDeviceMode(): DeviceMode {
```
szymon-nawrocki, line 81: Why can't we use `getDeviceType` that already exists in this class?

szymon-nawrocki (approved):

src/core/utils/client.ts
dmnsobczak: This method works a bit different and rather than checking screen size it is detecting real device. I don't know the potential consequences of that change and I didn't want to expand the scope of this ticket, so I left it for the future refactor :)

src/platforms/news-and-ratings/comicvine/comicvine-platform.ts
```diff
@@ -20,6 +21,7 @@ export class ComicvinePlatform {
 	execute(): void {
 		this.pipeline.add(
 			() => context.extend(basicContext),
+			() => context.set('state.isMobile', !utils.client.isDesktop()),
```
nandy-andy, line 24: What do you think of setting this in the `NewsAndRatingsBaseContextSetup`, so we don't have to repeat it in each N&R `*-platform.ts` file? 🤔

dmnsobczak: Targeting setup is executed quite late in the pipeline and mobile detection is/can be needed faster. I was thinking about some other pipe step for all platforms (including Fandom) to handle this, but for now I don't have any better idea :(

szymon-nawrocki: I tried to address this issue in this PR: https://github.com/Wikia/ad-engine/pull/1671/files where I set `isMobile` in `Bootstrap.setupContext()` which is called at the beginning of the pipeline

nandy-andy: Cool, let's postpone it and address it with Szymon's `Bootstrap` idea in the other PR 👌

szymon-nawrocki (dismissed):

PR #1724, opened 2023-01-30
ADEN-12692 Fixing SkyBoxes on all N&R platforms

dmnsobczak:


nandy-andy (dismissed):

src/platforms/news-and-ratings/giantbomb/styles.scss
```diff
@@ -18,3 +18,13 @@ body:not(.skybox-loaded) {
 		min-height: 0;
 	}
 }
+
+.skybox-loaded {
+	.mapped-skybox-nav {
+		min-height: 19vw;
```
szymon-nawrocki, line 24: Why only on Giantbomb there is defined `min-heigh > 0` ? Maybe it will be worth to add a comment explaining it to avoid "tribal knowledge"? 😄

dmnsobczak: Oh, yeah :D That's the hotfix for SkyBox on GiantBomb, because sometimes CSS for this unit are broken and this exists on production, so in theory we don't need to fix this, but we can keep this hack until SkyBox feature sunset :)

dmnsobczak: Unfortunately this causes problems here:
https://www.giantbomb.com/shows/forspoken/2970-22112?cid=TEST-SKYBOX&adengine_version=dev/ADEN-12692
Reverting.

szymon-nawrocki (approved):

PR #1725, opened 2023-01-30
ADEN-12626 Add Kargo to N&R bundles

nandy-andy:


src/ad-bidders/prebid/adapters/kargo.ts
```diff
@@ -20,14 +20,27 @@ export class Kargo extends PrebidAdapter {
...
-						placementId,
-					},
-				},
-			],
+			bids: this.getBids(placementId),
 		};
 	}
+
+	getBids(placementId) {
+		if (Array.isArray(placementId) && placementId.length > 0) {
```
szymon-nawrocki, line 28: As far as I can see, the `placementId` is always a single string, so we probably don't need this `if` ?

nandy-andy: Unfortunately, there are two unique cases when it's not true: gamespot and metacritic :/

szymon-nawrocki (approved):

szymon-nawrocki (approved):

PR #1731, opened 2023-01-31
ADEN-12721 Decode adtags on N&R platforms

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-12721

### DIFF
--- a/src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
+++ b/src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
@@ -31,9 +31,7 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
 	getPageLevelTargeting(): TargetingParams {
 		const adTags = this.getAdTags();
 		const parsedAdTags = this.parseAdTags(adTags);
-		const mappedAdTags = this.getMappedAdTags(parsedAdTags);
-
-		return mappedAdTags;
+		return this.getMappedAdTags(parsedAdTags);
 	}
 
 	getAdTags(): string {
@@ -52,7 +50,7 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
 
 			if (Array.isArray(parts) && parts.length === 2) {
 				const key = parts[0];
-				const value = decodeURIComponent(parts[1]).replace(/\+/g, ' ');
+				const value = decodeURIComponent(parts[1]).replaceAll('+', ' ');
 
 				adTagsObj[key] = value;
 			}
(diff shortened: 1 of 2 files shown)


src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
```diff
@@ -51,7 +51,10 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
 			const parts = keyval.split('=');
 
 			if (Array.isArray(parts) && parts.length === 2) {
-				adTagsObj[parts[0]] = parts[1];
+				const key = parts[0];
+				const value = decodeURIComponent(parts[1]).replace(/\+/g, ' ');
```
nandy-andy, line 55: We don't need a regular expression to replace `'+'` with `' '`, do we? Consider using just `.replace('+', ' ')` 🤔

szymon-nawrocki: The value can contain multiple plus signs. The `replace` method will replace only first match

szymon-nawrocki: > A string pattern will only be replaced once. To perform a global search and replace, use a regular expression with the g flag, or use [replaceAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) instead.

Taken from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

nandy-andy (approved):

src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
```diff
@@ -51,7 +49,10 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
 			const parts = keyval.split('=');
 
 			if (Array.isArray(parts) && parts.length === 2) {
-				adTagsObj[parts[0]] = parts[1];
+				const key = parts[0];
```
mfarul, line 52: the whole forEach could be replaced with this:
`for (const [key, value] of new URLSearchParams(adTags)) { 
    adTagsObj[key] = value;
}`
or some map function

mfarul (changes requested):

src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
szymon-nawrocki: Wow, thanks for the suggestion. Now this complicated splitting and mapping became much more clear one liner

mfarul (approved):

PR #1732, opened 2023-01-31
ADEN-12700 - Fixed invalid setting NPA value 

domjanik:
[Ticket](https://fandom.atlassian.net/browse/ADEN-12700)


mfarul (dismissed):

src/core/providers/gpt-provider.ts
```diff
@@ -209,7 +209,9 @@ export class GptProvider implements Provider {
 		const settings: Dictionary = {
 			restrictDataProcessing: trackingOptIn.isOptOutSale(),
 		};
-
+		if (context.get('options.geoRequiresConsent') && !trackingOptIn.isOptedIn()) {
+			settings.nonPersonalizedAds = true;
```
dmnsobczak, line 213: This is risky on so many levels. First: this goes global for all traffic, including Fandom.com. If this impacts revenue, then it will as hard as it can, so using rollout variable is a good idea.
Second: this approach is a bit deprecated. At the beginning Google was using `npa` keyval to mark non-personalised traffic. Then they dropped the `npa` keyval logic, because they switched fully to TCF and now they are base personalisation detection on Consent String and `npa` keyval is only for targeting purposes - Sebastian is using it for campaigns that not support TCF. Long story short: `npa=1` means "full consent" aka we're safe no matter what we do and/or track.
Third: `isOptedIn()` is not the same as `npa`. This method returns true (for TCF implementation) only if user accept all vendors and all purposes, without questions. Opting out one single purpose will switch to `npa` mode, when for Google it might mean, that personalised ads still can be served: see **Requirements to serve personalized ads**:
https://support.google.com/admanager/answer/9805023?hl=en&ref_topic=9760861

dmnsobczak (changes requested):
This is a major, dangerous change and it shouldn't be rolled out globally in one commit. Extended data check and analysis is a must-have :/

PR #1734, opened 2023-02-01
ADEN-12729 Load Anyclip player on GameFAQs and ComicVine

nandy-andy:


src/platforms/shared/setup/anyclip-player.setup.ts
```diff
@@ -11,23 +12,32 @@ import {
 
 export class AnyclipPlayerSetup extends BaseServiceSetup {
```
dmnsobczak, line 13: I don't like that Anyclip is still using `PlayerSetup` proxy class, but I think that we can handle it in one of the other Anyclip tickets (mobile integration or outstream removal) 🤔

src/platforms/news-and-ratings/gamefaqs/ad-context.ts
```diff
@@ -26,6 +26,13 @@ export const basicContext = {
 		},
 	},
 	services: {
+		anyclip: {
+			enabled: true,
```
dmnsobczak, line 30: Why `enabled` is hardcoded? Don't we want to use ICBM variable? 🤔

src/platforms/shared/setup/anyclip-player.setup.ts
```diff
@@ -11,23 +12,32 @@ import {
...
 	private initIncontentPlayer(incontentPlayer) {
-		const anyclip = new Anyclip(new AnyclipTracker(Anyclip.SUBSCRIBE_FUNC_NAME));
+		if (incontentPlayer) {
+			slotDataParamsUpdater.updateOnCreate(incontentPlayer);
+		}
 
-		if (!incontentPlayer) return;
```
dmnsobczak, line 16: I see potential change for Fandom Anyclip integration. Previously if `incontentPlayer` variable was falseable - logic stopped. Imagine the scenario when this happens and Anyclip is enabled globally in ICBM variable: this will create `Anyclip` and `AnyclipTracker` classes, previously this part was skipped. What will be the result for Fandom?

src/platforms/news-and-ratings/shared/modes/news-and-ratings-ads-mode.ts
```diff
@@ -18,6 +18,7 @@ import { Injectable } from '@wikia/dependency-injection';
 @Injectable()
 export class NewsAndRatingsAdsMode implements DiProcess {
 	constructor(
+		private anyclipPlayerSetup: AnyclipPlayerSetup,
```
dmnsobczak, line 21: Just for readability we are putting `private pipeline: PartnerPipeline` first and then all other dependencies in alphabetical order. Can you move this one line down? :D

dmnsobczak (changes requested):
Just a few questions 🥺

src/platforms/news-and-ratings/shared/modes/news-and-ratings-ads-mode.ts
nandy-andy: Yeah, I saw it! It was odd and I went with the "alphabetical order approach" ;) Sure, if this is the convention, I can change it 👍

src/platforms/news-and-ratings/gamefaqs/ad-context.ts
nandy-andy: Well, I thought Anyclip is supposed to be enabled globally anyway, so why bother with icVar 😅 and since we want to keep the payload from instant-config as minimal as possible it made even more sense. On the other hand, the `icAnyclipPlayer` exists and it includes even a list of countries, so it makes sense to reuse it for N&R. I'll also ask James about the list of countries if we want to keep the same or new but there may be a list of countries needed. Thanks for pointing this out!

src/platforms/shared/setup/anyclip-player.setup.ts
nandy-andy: I did the happy-path on our LongPage and nothing exploded ;) It creates the instances of the classes but it doesn't affect the page nor the user until `anyclip.loadPlayerAsset()` is called. But you're right. Maybe I can think of a way so it doesn't create the instances.

nandy-andy: Actually, it wasn't skipped before as the initialisation happened in line #14 and the early return happened in line #16. So, I think we're safe the way it is 🤞

dmnsobczak (dismissed):

src/platforms/news-and-ratings/comicvine/setup/context/video/comicvine-anyclip-applier.setup.ts
```diff
@@ -0,0 +1,8 @@
+import { context, DiProcess } from '@wikia/ad-engine';
+
+export class ComicvineAnyclipApplierSetup implements DiProcess {
+	execute(): void {
+		// TODO: change it to the flag in <meta name="ad-settings" /> once it's available
+		context.set('services.anyclip.isApplicable', () => true);
```
szymon-nawrocki, line 6: Can we stick with the convention `services.<service name>.enabled` ?

szymon-nawrocki: But this would require the assumption, that Anyclip is not enabled when in fact is enabled but not applicable 🙈

src/ad-services/anyclip/index.ts
```diff
@@ -7,28 +7,37 @@ export class Anyclip {
 	public static SUBSCRIBE_FUNC_NAME = 'lreSubscribe';
 
 	constructor(
+		private pubname: string = 'fandomcom',
+		private widgetname: string = '001w000001Y8ud2_19593',
+		private libraryUrl: string = '//player.anyclip.com/anyclip-widget/lre-widget/prod/v1/src/lre.js',
```
szymon-nawrocki, line 12: For me the code would be more clear if these values were moved to ad-context files - so it will be unified between all platforms

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
```diff
@@ -63,6 +63,8 @@ export class NewsAndRatingsBaseContextSetup implements DiProcess {
 			this.instantConfig.get('icComscoreJwpTracking'),
 		);
 		context.set('options.video.pauseJWPlayerAd', this.instantConfig.get('icPauseJWPlayerAd'));
+
+		context.set('services.anyclip.enabled', this.instantConfig.get('icAnyclipPlayer'));
```
szymon-nawrocki, line 67: Part of the Anyclip context is set up here, part of is set in `<Platform>AnyclipApplierSetup` which can be confusing.

I'd probably try to move the logic from `<Platform>AnyclipApplierSetup` into this file - so it is in one place where different service contexts are set up (here we even have the `setVideoOptions`)

src/platforms/shared/setup/anyclip-player.setup.ts
```diff
@@ -3,24 +3,38 @@ import {
...
+				context.get('services.anyclip.widgetname'),
+				context.get('services.anyclip.libraryUrl'),
+				context.get('services.anyclip.isApplicable'),
+				new AnyclipTracker(Anyclip.SUBSCRIBE_FUNC_NAME),
+			).loadPlayerAsset();
 		}
 	}
 
-	call() {
+	private registerAnyclipToLoadOnUapLoadStatus() {
```
szymon-nawrocki, line 37: Can you move this function under `call()` so it is easier to read?

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
szymon-nawrocki: Or if the logic conncted with TODOs `ad-settings` will overcomplicate this file, maybe it will be better to move link 67 from this file into `*AnyclipApplier` so Anyclip setup is in the one place

szymon-nawrocki: But here we set`videoOptions` 🙈  - maybe we can leave it for anyclip development improvements then

src/platforms/news-and-ratings/gamefaqs/setup/context/video/gamefaqs-anyclip-applier.setup.ts
```diff
@@ -0,0 +1,12 @@
+import { context, DiProcess } from '@wikia/ad-engine';
+
+export class GamefaqsAnyclipApplierSetup implements DiProcess {
+	private isUserLoggedInBasedOnTargeting() {
```
szymon-nawrocki, line 4: Moving this method under `execute` will make  it easier to read

szymon-nawrocki (dismissed):

dmnsobczak (approved):

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
```diff
@@ -68,6 +68,12 @@ export class NewsAndRatingsBaseContextSetup implements DiProcess {
...
+		context.set('services.anyclip.isApplicable', () => {
```
szymon-nawrocki, line 73: Btw. why `isApplicable` is a function, if the callback is not used later? Why isn't it just boolean?

nandy-andy: Yeah, it can be for now 🤔 In the future when we'll introduce more advanced logic on the frontend checking state of JWPlayer/UAP etc. it may make sense to be a function but now just a Boolean flag should be fine.

szymon-nawrocki (dismissed):

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
nandy-andy: After discussion on slack: let's leave it for now as we're still waiting for info on how it'll work on ComicVine. I've created a follow-up we can discuss after it's live on GameFAQs and ComicVine: https://fandom.atlassian.net/browse/ADEN-12753

kwfandom (approved):

PR #1737, opened 2023-02-02
ADEN-12693 Adding A9 integration to N&R platforms

dmnsobczak:


src/platforms/news-and-ratings/gamefaqs/setup/context/prebid/gamefaqs-prebid-config.setup.ts
```diff
@@ -22,11 +21,7 @@ export class GamefaqsPrebidConfigSetup implements DiProcess {
 	constructor(private bidders: Bidders) {}
 
 	execute(): void {
-		const isDesktop = utils.client.isDesktop();
-
-		if (!this.isPrebidEnabled()) {
```
nandy-andy, line 27: Why don't we want the safety-net here? Because it's being checked in the shared bidder setup?

nandy-andy (approved):

src/platforms/news-and-ratings/gamefaqs/setup/context/prebid/gamefaqs-prebid-config.setup.ts
dmnsobczak: Yeah, and in the `ad-bidders/index` - Prebid class won't be created if Prebid is not enabled 🤷

PR #1741, opened 2023-02-06
ADEN-12739 Add new script for Confiant on N&R

nandy-andy:


src/ad-services/confiant/index.ts
```diff
@@ -32,24 +20,41 @@ function trackBlock(blockingType, blockingId, isBlocked, wrapperId, tagId, impre
  * Confiant service handler
  */
 export class Confiant extends BaseServiceSetup {
+	protected scriptDomain = 'cdn.confiant-integrations.net';
```
szymon-nawrocki, line 23: why these fields are not `private`?

nandy-andy: it's useful for tests for example but I didn't use it in this way here, so I'll change it 👌

PR #1745, opened 2023-02-07
ADEN-12720 use tracking params from window object

jakubbanas:


src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -3,7 +3,7 @@ import { Injectable } from '@wikia/dependency-injection';
 import { TrackingParams } from './models/tracking-params';
 
 const logGroup = 'data-warehouse-trackingParams';
-const eventUrl = 'https://beacon.wikia-services.com/__track/special/trackingevent';
+const eventUrl = 'https://beacon.wikia-services.com/"__track/special/trackingevent"';
```
domjanik, line 6: @jakubbanas Why do we need to add those quotation marks?

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -1,29 +1,55 @@
-import { context, DiProcess } from '@wikia/ad-engine';
+import { context, DiProcess, InstantConfigService, utils } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
 import Cookies from 'js-cookie';
 import { getMediaWikiVariable } from '../utils/get-media-wiki-variable';
 
+function getLegacyTrackingParameters(): ITrackingParameters {
```
domjanik, line 6: I think it would be cleaner, to keep those methods inside TrackingParametersSetup if we do not need them outside of that class 🤔

domjanik (approved):

src/platforms/shared/tracking/data-warehouse.ts
jakubbanas: No idea, I put it by mistake. Good catch!

domjanik (approved):

PR #1758, opened 2023-02-10
ADEN-12754 Fix seamless content loading

nandy-andy:


@types/window/index.d.ts
```diff
@@ -48,4 +48,5 @@ interface Window {
 	utag_data?: any;
 	wgCookiePath?: string;
 	XMLHttpRequest?: any;
+	Phoenix?: any;
```
szymon-nawrocki, line 51: Can Phoenix be type of `Phoenix` interface (not any)?

szymon-nawrocki: Okay, I don't think we should do the `import` here

szymon-nawrocki (approved):

szymon-nawrocki (approved):

PR #1765, opened 2023-02-14
ADEN-12746 Use @wikia/instant-config-loader library

jbienkowski-fandom:
Lock mechanism will be available in version 0.2.0 -> see: https://github.com/Wikia/instant-config-loader/pull/3


domjanik (dismissed):

jbienkowski-fandom:
Total code coverage went down, because we moved some highly covered code out of this codebase. It has been agreed that it is not an issue for this PR.

src/core/utils/query-string.ts
```diff
@@ -48,9 +48,6 @@ class QueryString {
 		try {
 			return JSON.parse(value);
 		} catch (ignore) {
-			if (value.startsWith('[') && value.endsWith(']')) {
```
jbienkowski-fandom, line 45: It is handled by `JSON.parse` therefore this code was never reached

jakubbanas (approved):

PR #1777, opened 2023-02-21
ADEN-12660 Make video ads ad units differ from display ads ad units

nandy-andy:


spec/platforms/news-and-ratings/shared/context/base/news-and-ratings-base-context.setup.spec.ts
```diff
@@ -8,17 +8,21 @@ describe('News and Ratings base context setup', () => {
 
 	beforeEach(() => {
 		instantConfigStub = global.sandbox.createStubInstance(InstantConfigService);
+		context.set('custom.region', 'test');
```
szymon-nawrocki, line 11: It would be nice to differentiate this `test` value from the value of `custom.property`, so it is visible where each part is used in building video adunit id

src/platforms/news-and-ratings/shared/setup/context/base/news-and-ratings-base-context.setup.ts
```diff
@@ -9,7 +9,7 @@ export class NewsAndRatingsBaseContextSetup implements DiProcess {
 		this.setBaseState();
 		this.setupIdentityOptions();
 		this.setupServicesOptions();
-		this.setupVideoOptions();
+		this.setupVideo();
```
szymon-nawrocki, line 12: I think this logic is only connected with JWPlayer - if this is the case I think it would be nice to change this method name to `setupJWPlayer` so the method name says clearly which video player is being set up there

nandy-andy: If it is so, we have a problem! I was under an impression it sets up context for JWPlayer AND Anyclip 🙀

szymon-nawrocki: You're right my bad

szymon-nawrocki (approved):

PR #1791, opened 2023-02-28
ADEN-12839 Refresh bids for lazy-loaded slots on TV Guide

nandy-andy:


src/platforms/news-and-ratings/tvguide/setup/context/slots/tvguide-slots-context.setup.ts
```diff
@@ -244,6 +244,7 @@ export class TvGuideSlotsContextSetup implements DiProcess {
 				},
 			},
 			'incontent-leader-plus-bottom': {
+				bidderAlias: 'incontent-leader-plus-bottom',
```
dmnsobczak, line 247: Bidder aliases are not required: they are used only if Prebid ad unit code is different from slot name or two slots share same Prebid ad unit code. Here those two values are equal, so bidderAlias is redundant :)

dmnsobczak (dismissed):

src/platforms/news-and-ratings/tvguide/setup/context/slots/tvguide-slots-context.setup.ts
nandy-andy: I tried but it won't work since there are no incremental slots in the bidders' contexts and for `mpu-middle-3` we want bidders to refresh `mpu-middle` thus aliasing :)

dmnsobczak (approved):

szymon-nawrocki (approved):

src/platforms/news-and-ratings/tvguide/setup/context/slots/tvguide-slots-context.setup.ts
dmnsobczak: Yes, and we need to set `bidderAlias`, because otherwise `mpu-middle-3` won't know where to look bids for itself 👍

PR #1794, opened 2023-03-01
ADEN-12846 Bump @wikia/instant-config-loader - fixes loading Instant Config from Local Storage

jbienkowski-fandom:


webpack.platforms.js
```diff
@@ -32,6 +32,11 @@ const platforms = ({ entry }) => ({
 	},
 
 	devServer: {
+		allowedHosts: 'all', // FIXME: maybe specify platform subdomains?
```
jbienkowski-fandom, line 35: It fixes HMR while using `adengine_version=local`, otherwise webpack would use wrong host (e.g. `wss://www.fandom.com:9000/ws`) and would fail. We could be more specific here, but maybe it is overkill?

jbienkowski-fandom:
```suggestion
		allowedHosts: 'all',
```

dmnsobczak (approved):

domjanik (approved):

domjanik (approved):

PR #1798, opened 2023-03-02
ADEN-12854 -  gathering and sending data to prometheus throug…

adi-wesolowski-fandom:
…h pandora service, for monitoring purpose


src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,129 @@
...
+		};
+	}
+
+	private getEndpointUrlFor(endpointSpecific: string, { baseUrl, service }): string {
+		return [baseUrl, service, 'api', 'adengine', 'meter', endpointSpecific]
+			.filter((element) => !!element)
+			.join('/');
+	}
+
+	private clickHookOnIframe(prevTriggeredId: string): void {
```
andrzejkupczyk, line 98: It seems that the first call always passes an empty string. If so, I would suggest making `prevTriggeredId` optional or giving it a default value.

src/platforms/shared/setup/monitoring.setup.ts
```diff
@@ -0,0 +1,32 @@
...
+
+		return queryParams.adengine_version === 'local';
+	}
+
+	private getServiceName(): string {
+		return this.isLocal() ? '' : 'icbm';
+	}
+
+	private getServiceBaseURL(): string {
+		return this.isLocal() ? 'http://localhost:8080' : utils.getServicesBaseURL();
```
andrzejkupczyk, line 30: I wonder if it is possible to avoid hard coding localhost? Maybe for example `InstantConfigService` could be used for that?

src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,129 @@
...
+		let i = 1;
+		let currentElement = element;
+		while (i <= 3) {
+			i++;
+			currentElement = currentElement.parentElement;
+			if (currentElement && currentElement.id && currentElement.id.split('/').length > 1) {
+				continue;
+			}
+			return currentElement.id;
+		}
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: I dont understand, what you mean by "exacly same result". Yes, result should be the same and this is the case. But in `currentElement` we dont have the same element all the time - it skip up the ladder in node tree. 

The reason for it is, I want parent element, that match the pattern described as if statement. It is because, somethimes we got iframe or div of the Ad, put deeper by googletag, but we want to catch our id, not googletag id.

src/platforms/shared/tracking/monitoring.ts
adi-wesolowski-fandom: Yes, it is empty. But can be null I guess. This is a leftover from early version of PoC, and I didnt enhance it :) The reason is to get a click event for new Ad, or else we would stuck on only one click-event

src/platforms/shared/setup/monitoring.setup.ts
adi-wesolowski-fandom: It is used to get the prod behaviour. But still, many link in AdEngine are hardcoded. Here I put it to be able to send data to local instance of Pandora. But I can use other query param as well. As for ICBM - I dont see oportunity to do so, because I get config from prod on my dev-box, even with "local" param. Thats why I did it this way.

src/platforms/shared/tracking/monitoring.ts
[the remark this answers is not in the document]
andrzejkupczyk: Gotcha. `Element.closest()` accepts any CSS selector. It's pretty powerful so imho it might meet your needs 🙂 

```js
const closestAdElement = element.closest('[id]:not([id*="/"])');
```

☝🏻 finds closest ancestor element having non-empty id attribute, which does not contain "/" character (or `null` if there is no such element).

https://jsfiddle.net/w03c54xj/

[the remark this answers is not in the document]
andrzejkupczyk: Thank u!

src/platforms/shared/setup/monitoring.setup.ts
adi-wesolowski-fandom: Not, port should not change, at least as I am aware

andrzejkupczyk (approved):

src/core/utils/logger.ts
```diff
@@ -8,10 +8,15 @@ if (debugGroup !== '') {
...
-export function getTimeDelta(): string {
+export function getTimeDelta(): number {
 	const currentTimeStamp = Date.now();
 	const originTime = window.performance?.timeOrigin ?? BACKUP_ORIGIN_TIME;
-	const delta = currentTimeStamp - originTime;
+
+	return currentTimeStamp - originTime;
+}
+
+export function getTimeDeltaInSeconds(): string {
```
David-Iacovone, line 18: I would suggest to change the return type to number, since this function aims to return seconds.

David-Iacovone (changes requested):
Looks good, just a few cosmetic suggestions

src/core/utils/logger.ts
adi-wesolowski-fandom: I think I wanted to keep string as it was there before, and to keep campatibility with what it was before. But in the end, it seems to be unnecessary. Ok, I will change it. thx

adi-wesolowski-fandom: Well... actually it return seconds, but not full seconds, there is comma. Also, function return the result of the "toFixed", which force to be "string" type.

David-Iacovone (changes requested):

src/core/utils/outbound-traffic-restrict.ts
```diff
@@ -0,0 +1,28 @@
+import { isLocal } from '../../platforms/shared';
+import { context } from '../services';
+
+class OutboundTrafficRestrict {
+	private static MIN = 0;
+	private static MAX = 100;
+	private static DEFAULT_THRESHOLD = 2;
```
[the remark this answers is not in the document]
David-Iacovone: It seems that they still can be modified. It looks kinda wired to me to use `private` and `static` on one property, since static suggest external usage of those properties

David-Iacovone (dismissed):

andrzejkupczyk (dismissed):

jakubbanas (approved):

mfarul (changes requested):

src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,96 @@
+import { AdSlot, communicationService, context, eventsRepository, utils } from '@wikia/ad-engine';
+
+interface EndpointInfo {
+	baseUrl: string;
+	service: string;
+	appName?: string;
+}
+
+export class Monitoring {
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: ok

src/core/utils/outbound-traffic-restrict.ts
```diff
@@ -1,13 +1,32 @@
...
-			this.getSeed() < (context.get(`services.${serviceName}.threshold`) || this.DEFAULT_THRESHOLD)
+		const isAllowedFromContext = context.get(`services.${serviceName}.allowed`);
+		if (typeof isAllowedFromContext === 'boolean') {
+			return isAllowedFromContext;
+		}
+
+		const threshold =
+			context.get(`services.${serviceName}.threshold`) || this.ALWAYS_ALLOWED_THRESHOLD;
+		let isAllowed = true;
+		if (threshold !== this.ALWAYS_ALLOWED_THRESHOLD) {
```
andrzejkupczyk: The name of the constant and the condition parameter imply that the threshold value cannot be larger than 250, which might cause confusion for anyone working with the code in the future. Would it be possible to consider renaming the constant and parameter to something more generic that doesn't imply a specific threshold value? This would help avoid any confusion or potential bugs in the future. Thanks for considering this suggestion!

src/platforms/shared/setup/monitoring.setup.ts
```diff
@@ -0,0 +1,31 @@
...
+			'services.monitoring.threshold',
+			this.instantConfig.get('icMonitorTrafficThreshold'),
+		);
+	}
+
+	private getServiceName(): string {
+		return isLocal() ? '' : 'icbm';
+	}
+
+	private getServiceBaseURL(): string {
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: It was for 2 reasons: 
- not send metric data during tests or development using local version (as it may have some errors and not show ads, etc
- - to not send data to prod pandora and therefor, do not mess with data there if something went wrong during development
- to send data to local instance of Pandora during development, and have local copy of this data

src/core/utils/outbound-traffic-restrict.ts
andrzejkupczyk: Okay, it looks like the code I reviewed has been already outdated 🤔

adi-wesolowski-fandom: I used current "outbound-traffic-restricter" from "dev", that was created and deployed with task for DW-limitation

src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,96 @@
...
+		communicationService.on(eventsRepository.AD_ENGINE_AD_CLICKED, (data) => {
+			this.sendSlotStateToMeteringSystem(data.slotName, 'click');
+		});
+	}
+
+	private sendSlotStateToMeteringSystem(slotName: string, state: string): void {
+		const endpointInfo = this.getEndpointInfoFromContext();
+		const endpointUrl = this.getEndpointUrlFor('slot', endpointInfo);
+
+		fetch(`${endpointUrl}?app=${endpointInfo.appName}&slot=${slotName}&state=${state}`);
```
andrzejkupczyk, line 55: To ensure that the parameters are properly encoded, may I suggest using the `URLSearchParams` method to encode the parameters before passing them to the fetch function? This will help avoid any issues with special characters or spaces in the parameters. Same for the `sendTimeDataToMeteringSystem` method.

andrzejkupczyk (approved):

src/platforms/shared/setup/monitoring.setup.ts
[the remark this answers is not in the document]
adi-wesolowski-fandom: it can be

mfarul (dismissed):

src/platforms/ucp-desktop/ucp-desktop-platform.ts
```diff
@@ -43,6 +45,8 @@ export class UcpDesktopPlatform {
 			PlatformContextSetup,
 			parallel(InstantConfigSetup, () => bootstrapAndGetConsent()),
 			TrackingParametersSetup,
+			MetricReporterSetup,
```
mfarul, line 48: this can be added to ucp-mobile and to other platforms as sell

adi-wesolowski-fandom: ok, done

src/platforms/shared/tracking/monitoring.ts
adi-wesolowski-fandom: I added encoding, and change building url, into one place only

src/platforms/shared/setup/monitoring.setup.ts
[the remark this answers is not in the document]
adi-wesolowski-fandom: I removed all "isLocal" references

src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,96 @@
...
+		};
+	}
+
+	private getEndpointUrlFor(endpointSpecific: string, { baseUrl, service }): string {
+		return [baseUrl, service, 'api', 'adengine', 'meter', endpointSpecific]
+			.filter((element) => !!element)
+			.join('/');
+	}
+
+	private clickHookOnIframe(prevTriggeredId: null | string = null): void {
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: extracted

src/platforms/shared/tracking/monitoring.ts
```diff
@@ -0,0 +1,96 @@
...
+		const endpointUrl = this.getEndpointUrlFor('slot', endpointInfo);
+
+		fetch(`${endpointUrl}?app=${endpointInfo.appName}&slot=${slotName}&state=${state}`);
+	}
+
+	private sendTimeDataToMeteringSystem(action: string, duration: number): void {
+		const endpointInfo = this.getEndpointInfoFromContext();
+		const endpointUrl = this.getEndpointUrlFor('time', endpointInfo);
+
+		fetch(`${endpointUrl}?app=${endpointInfo.appName}&action=${action}&duration=${duration}`);
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: done - I added fetch with support for timeout (default = 2 sec)

mfarul (approved):

PR #1815, opened 2023-03-07
ADEN-12836 - Modified pvUniqueId source on N&R

domjanik:


src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -12,7 +11,7 @@ export class TrackingParametersSetup implements DiProcess {
 		return (
 			getMediaWikiVariable('pvUID') || // UCP
 			window.pvUID || // F2
-			v4() // N+R
+			window.fandomContext.tracking.pvUID // N+R
```
jakubbanas, line 14: This line will be redundant if `pvUID` is already in `tracking` object.
Also, you should remove line 47, and it should work fine.

domjanik: After brief discussion we decided to leave it as a safety fallback.

jakubbanas (approved):

PR #1816, opened 2023-03-07
ADEN-12771 - Add new API for BlockThrough

domjanik:


src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+	constructor() {
+		if (!window.fandomContext) {
+			this.createEmptyContext();
+		}
+	}
+
+	private createEmptyContext(): void {
+		utils.logger(this.logGroup, 'Creating empty fandomContext');
+		window.fandomContext = {
```
jbienkowski-fandom, line 21: I don't like this, it breaks `WindowFandomContext`'s interface and is a source of potential bugs... I would rather extend WindowFandomContext to allow empty objects by Partial<IXYZParameters>`. Otherwise I bet otherwise sooner or later someone will use `window.fandomContext.tracking.something.nested` and it will fail

@types/fandom/index.d.ts
```diff
@@ -20,25 +20,38 @@ interface ITrackingParameters {
...
-		lang: string;
-		pageId: number;
-		pageName: string;
-		pageType: string;
-		tags: TaxonomyTags;
-		wordCount: number;
-	};
+	site: ISiteParameters;
+	page: IPageParameters;
 	tracking?: ITrackingParameters;
```
jbienkowski-fandom, line 55: I guess `tracking` and `partners` are not optional anymore but should me marked with `Partial` as `GlobalContextService` will initialize them to empty object

@types/fandom/index.d.ts
```diff
@@ -20,25 +20,38 @@ interface ITrackingParameters {
...
-		articleId: number;
-		lang: string;
-		pageId: number;
-		pageName: string;
-		pageType: string;
-		tags: TaxonomyTags;
-		wordCount: number;
-	};
+	site: ISiteParameters;
+	page: IPageParameters;
```
jbienkowski-fandom, line 54:
```suggestion
	page: Partial<IPageParameters>;
```

jbienkowski-fandom (changes requested):

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,70 @@
...
+			site: {} as ISiteParameters,
+			page: {} as IPageParameters,
+			tracking: {} as ITrackingParameters,
```
jbienkowski-fandom, line 24:
```suggestion
			tracking: {},
```
should work as tracking is `Partial`, same for `partners` (and could be same for site and page as they should be `Partial` too imho)

spec/ad-services/eyeota/eyeota.spec.ts
```diff
@@ -95,7 +95,12 @@ describe('Eyeota', () => {
 	});
 
 	it('constructs proper url with context', async () => {
-		window.fandomContext = { site: null, page: null } as FandomContext;
+		window.fandomContext = {
+			site: null,
```
jbienkowski-fandom, line 99:
```suggestion
			site: {},
```
and same with others, then you will not have to use `as FandomContext`

src/ad-services/eyeota/index.ts
```diff
@@ -17,6 +17,9 @@ export function parseContextTags(tags: TaxonomyTags): string {
 }
 
 export class Eyeota extends BaseServiceSetup {
+	constructor(props) {
```
jbienkowski-fandom, line 20: this is unnecessary imho

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,70 @@
...
+			...replaced,
+		} as A & B;
+	}
+
+	setValue<K extends keyof WindowFandomContext, V extends Required<WindowFandomContext[K]>>(
+		category: K,
+		value: V,
+	): void {
+		if (!window.fandomContext[category]) {
+			window.fandomContext[category] = {} as any;
```
jbienkowski-fandom, line 55:
```suggestion
			window.fandomContext[category] = {};
```
should work without `as any` after changes to WindowFandomContext types (see previous comments)

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,70 @@
...
+		}
+		window.fandomContext[category] = this.merge(window.fandomContext[category], value) as any;
```
jbienkowski-fandom, line 57:
```suggestion
		window.fandomContext[category] = this.merge(window.fandomContext[category], value);
```
same here

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	getValue(category: GlobalContextCategories, key?: string): unknown {
```
jbienkowski-fandom, line 60: this method typings' could be improved too:
```suggestion
	getValue<K extends keyof WindowFandomContext, I extends keyof WindowFandomContext[K]>(
		category: K,
		key?: I,
	): WindowFandomContext[K][I] | undefined {
```

jbienkowski-fandom (changes requested):

src/ad-services/eyeota/index.ts
domjanik: This change has been introduced, because Intellij sometimes has problems with our typings. I know it is unnecessary, but it looks better in our IDE :D If you insist on removing it - I can do it :)

jbienkowski-fandom: okay, then it is fine by me ;)

spec/ad-services/eyeota/eyeota.spec.ts
```diff
@@ -95,7 +95,12 @@ describe('Eyeota', () => {
 	});
 
 	it('constructs proper url with context', async () => {
-		window.fandomContext = { site: null, page: null } as FandomContext;
+		window.fandomContext = {
+			site: {},
+			page: null,
```
jbienkowski-fandom, line 100:
```suggestion
			page: {},
```

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,73 @@
...
+	constructor() {
+		if (!window.fandomContext) {
+			this.createEmptyContext();
+		}
+	}
+
+	private createEmptyContext(): void {
+		utils.logger(this.logGroup, 'Creating empty fandomContext');
+		window.fandomContext = {
+			site: {} as ISiteParameters,
```
jbienkowski-fandom, line 22:
```suggestion
			site: {},
```

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,73 @@
...
+			page: {} as IPageParameters,
```
jbienkowski-fandom, line 23:
```suggestion
			page: {},
```

src/core/services/global-context-service.ts
```diff
@@ -0,0 +1,73 @@
...
+			tracking: {},
+			partners: {} as IPartnersParameters,
```
jbienkowski-fandom, line 25:
```suggestion
			partners: {},
```

jbienkowski-fandom (approved):

PR #1818, opened 2023-03-09
ADEN-12433 Removing PorvataTemplate and Outstream player

dmnsobczak:


src/ad-services/anyclip/index.ts
```diff
@@ -25,27 +45,38 @@ export class Anyclip {
 		};
 	}
 
-	loadPlayerAsset() {
+	private setupConfig() {
+		this.pubname = context.get('services.anyclip.pubname') || 'fandomcom';
+		this.widgetname = context.get('services.anyclip.widgetname') || '001w000001Y8ud2_19593';
+		this.libraryUrl =
```
nandy-andy, line 51: Consider writing getters for `pubname`, `widgetname` and especially `libraryUrl`. They'll get the value from the context or fallback to the hardcoded defaults. Using here getters will make the three lines of code more readable:
```ts
this.pubname = this.getPubName();
this.widgetname = this.getWidgetName();
this.libraryUrl = this.getLibraryUrl();
```
instead of:
```ts
this.pubname = context.get('services.anyclip.pubname') || 'fandomcom';
this.widgetname = context.get('services.anyclip.widgetname') || '001w000001Y8ud2_19593';
this.libraryUrl = context.get('services.anyclip.libraryUrl') || '//player.anyclip.com/anyclip-widget/lre-widget/prod/v1/src/lre.js';
```

src/ad-services/anyclip/index.ts
```diff
@@ -55,11 +86,38 @@ export class Anyclip {
 			});
 	}
 
+	private loadOnUapReadyStatus() {
```
nandy-andy, line 89: It does not really load Anyclip on UAP ready status. It registers a callback for the action. Consider renaming the method so it better describes what it does.

src/ad-services/anyclip/index.ts
```diff
@@ -55,11 +86,38 @@ export class Anyclip {
...
+		communicationService.on(
+			eventsRepository.AD_ENGINE_UAP_LOAD_STATUS,
+			({ isLoaded, adProduct }: UapLoadStatus) => {
```
nandy-andy, line 92: Consider extracting the callback, so the part below will be more readable.

src/ad-tracking/slot-tracker.ts
```diff
@@ -28,7 +28,7 @@ class SlotTracker extends BaseTracker implements BaseTrackerInterface {
 	compilers = [slotPropertiesTrackingCompiler, slotTrackingCompiler, pageTrackingCompiler];
 
 	isEnabled(): boolean {
-		return context.get('options.tracking.slot.status');
+		return true;
```
nandy-andy, line 31: Good catch. By removing it we could also remove the instant-config variable: [icSlotTracking](https://services.fandom.com/icbm/dashboard/feature_flag/icSlotTracking). However, maybe it's worth holding up with it and create a follow-up ticket to discuss if we want to track it all since we're recently lookings for savings 🤔

src/core/video/video-event-provider.ts
```diff
@@ -43,4 +39,19 @@ export class VideoEventDataProvider {
 			video_number: videoData.video_number || '',
 		};
 	}
+
+	/**
+	 * Emit single event
```
nandy-andy, line 44: I think we can remove this comment. It does not give any value.

src/platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -58,4 +62,15 @@ export class UcpTargetingSetup implements DiProcess {
 
 		return createSelectedStrategy(selectedStrategy, createFandomContext()).get();
 	}
+
+	private setCustomPlayerRuntimeAdUnit(slotName = 'incontent_player'): void {
```
nandy-andy, line 66: Could you move it to a separate setup step? It's not really a targeting, it's something we agreed with our partners to share. I suggest creating `UcpRuntimeSetup` and moving it there.

nandy-andy (commented):
I left some minor comments. A bigger concern for me is that with these changes we've expanded responsibilities of `Anyclip` class to not only load 3rd-party asset and integrate with it via the event but also we moved there creation/initiation of `incontent_player` slot. I think we need to separate it in case in the future we'll have one more 3rd-party outstream player integration. However, since this is about minimising outstream players to 1 we could create a follow-up and discuss it at our technical refinement.

src/ad-services/anyclip/index.ts
dmnsobczak: Good idea, but I compressed it even more :D I always forgetting that getter are in JavaScript ;_;

src/core/video/video-event-provider.ts
dmnsobczak: Agree

src/ad-services/anyclip/index.ts
dmnsobczak: Done

src/ad-services/anyclip/index.ts
dmnsobczak: Done!

src/platforms/shared/context/targeting/ucp-targeting.setup.ts
dmnsobczak: Moved to a different place

dmnsobczak:
@nandy-andy This follow up will simplify the IncontentPlayer creation logic:
https://github.com/Wikia/ad-engine/pull/1844
Also we have another follow up to rethink Anyclip logic one more:
https://fandom.atlassian.net/browse/ADEN-12884

nandy-andy (approved):

nandy-andy (approved):

PR #1819, opened 2023-03-09
ADEN-12559 - Removed UserIdentity from AdEngine

domjanik:


src/ad-services/identity-setup/index.ts
```diff
@@ -0,0 +1,24 @@
...
+		const identityPromise = new Promise<void>((res) => {
+			this.identityReady = res;
+		});
+
+		communicationService.on(eventsRepository.IDENTITY_ENGINE_READY, () => {
+			this.identityReady();
+			clearTimeout(this.fallback);
+			utils.logger(this.logGroup, 'initialized');
+		});
+		return identityPromise;
```
jbienkowski-fandom, line 22:
```suggestion
		return new Promise((resolve) => {
		    communicationService.on(eventsRepository.IDENTITY_ENGINE_READY, () => {
			resolve();
			clearTimeout(this.fallback);
			utils.logger(this.logGroup, 'initialized');
		    });
		});
```
it should work just the same, but you won't need that `identityReady` property which I guess will reduce bundle size a bit

src/ad-services/identity-setup/index.ts
```diff
@@ -0,0 +1,22 @@
+import { communicationService, eventsRepository } from '@ad-engine/communication';
+import { BaseServiceSetup, utils } from '@ad-engine/core';
+
+export class IdentitySetup extends BaseServiceSetup {
+	private logGroup = 'identity-setup';
+	private identityReady: () => void;
+	private fallback = setTimeout(() => {
+		utils.logger(this.logGroup, 'Fallback launched');
+		this.identityReady();
+	}, 2500);
```
jakubbanas, line 10: Is it a good idea to wait for Identity-engine that long?

jakubbanas (approved):

jakubbanas (dismissed):

jakubbanas (approved):

PR #1821, opened 2023-03-10
ADEN-12864 | Add 3 video bidders to N&R

rafalgiemza:
https://fandom.atlassian.net/browse/ADEN-12864

- bidders: appnexus, pubmatic, rubicon
- index exchange was already done
- for Metacritic-neutron & ComicVine there was no video prebbid ids
- A9 will be done in separate ticket


src/platforms/news-and-ratings/gamefaqs/bidders/rubicon-display.ts
```diff
@@ -46,6 +46,9 @@ export function getRubiconDisplayContext(isDesktop: boolean): object {
 			siteId: '462802',
 			zoneId: '2719822',
 		},
+		video: {
```
nandy-andy, line 49: This may not work because of two reasons:
1. we're missing the zoneIds
2. Magnite (old name Rubicon) video has probably a separate implementation, so we need a new config file for https://github.com/Wikia/ad-engine/blob/dev/src/ad-bidders/prebid/adapters/rubicon.ts

src/platforms/news-and-ratings/gamefaqs/bidders/appnexus.ts
```diff
@@ -38,6 +38,9 @@ export function getAppnexusContext(isDesktop: boolean): object {
 			sizes: [[728, 90]],
 			placementId: ['28574403'],
 		},
+		video: {
```
dmnsobczak, line 41: `appnexus` is not a video adapter - use `appnexusAst`

dmnsobczak, line 41: `appnexus` is not a video adapter - use `appnexusAst`

src/platforms/news-and-ratings/gamefaqs/bidders/rubicon-display.ts
dmnsobczak: Yep, you need to use `rubicon` adapter - those two adapter generate ad units in different structure.

dmnsobczak (changes requested):

dmnsobczak:
Also: if we want to support video bidders on N&R we need GAM VAST redirect catch all order to properly pass bid VAST content to the player 🤔

nandy-andy:
Right, because now we've got only IX video lines: https://admanager.google.com/5441#delivery/order/order_overview/order_id=3146768554

rafalgiemza:
Pubmatic part was moved here: https://github.com/Wikia/ad-engine/pull/1823

PR #1835, opened 2023-03-13
ADEN-12877 | Add Rubicon prebid video on N&R

rafalgiemza:
https://fandom.atlassian.net/browse/ADEN-12877
ADEN-12877 | Add Rubicon (Magnite) prebid video on N&R
affected sites: metacritic, gamefaqs, gamespot, giantbomb & tvguide

### DIFF
--- a/src/platforms/news-and-ratings/gamefaqs/bidders/rubicon.ts
+++ b/src/platforms/news-and-ratings/gamefaqs/bidders/rubicon.ts
@@ -1,0 +1,37 @@
+export function getRubiconContext(isDesktop: boolean): object {
+	const desktopSlots = {
+		slots: {
+			featured: {
+				siteId: '462802',
+				sizeId: '201',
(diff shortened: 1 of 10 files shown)


src/platforms/news-and-ratings/gamefaqs/bidders/rubicon.ts
```diff
@@ -0,0 +1,37 @@
+export function getRubiconContext(isDesktop: boolean): object {
+	const desktopSlots = {
+		slots: {
+			featured: {
```
nandy-andy, line 4: Will it work? 🤔 There are no such slots on N&R AFAIK. I recall there is a video one for example: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/news-and-ratings/gamefaqs/setup/context/slots/gamefaqs-slots-context.setup.ts#L290-L292

PR #1841, opened 2023-03-17
ADEN-12878 | Integrate A9 video prebid on N&R

rafalgiemza:


src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -36,7 +36,8 @@ export class BiddersStateSetup implements DiProcess {
 	}
 
 	setupBidders(): void {
-		const hasFeaturedVideo = context.get('custom.hasFeaturedVideo');
+		const hasFeaturedVideo =
+			context.get('custom.hasFeaturedVideo') || context.get('slots.video.isVideo');
```
dmnsobczak, line 40: `context.get('slots.video.isVideo')` is basically hardcoded `true` because it takes value from context:
https://github.com/Wikia/ad-engine/blob/dev/src/platforms/news-and-ratings/giantbomb/setup/context/slots/giantbomb-slots-context.setup.ts#L168

dmnsobczak: So the real logic here is:

* If there is a FeaturedVideo on UCP
* Or News&Ratings

Second part you could replace with skin comparison or anything else easier to read or understand but that still don't fix the issue that we don't want to do heavy, slot video requests on pages where video doesn't exist. You need to find a way to be sure whether video is on the page or not.

src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -69,6 +73,16 @@ export class BiddersStateSetup implements DiProcess {
 			'bidders.enabled',
 			context.get('bidders.prebid.enabled') || context.get('bidders.a9.enabled'),
 		);
+
+		function isAnyVideo(): boolean {
```
dmnsobczak, line 77: Can you convert it to a private class function?

rafalgiemza: sure

src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -69,6 +73,16 @@ export class BiddersStateSetup implements DiProcess {
...
+			// Is there is a FeaturedVideo on UCP?
```
dmnsobczak: We usually don't add comments to the code: code should be self-commenting via properly named variables and methods.

src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -69,6 +73,16 @@ export class BiddersStateSetup implements DiProcess {
...
+			const hasFeaturedVideo = context.get('custom.hasFeaturedVideo');
```
dmnsobczak, line 79: What about `if (context.get('custom.hasFeaturedVideo')) return true;` ?

src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -69,6 +73,16 @@ export class BiddersStateSetup implements DiProcess {
...
+
+			// Is there a video on News&Ratings?
+			const ptypeIncludesVideo = targetingService.get('ptype')?.includes('video_');
```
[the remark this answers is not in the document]
dmnsobczak: Bonus points: currently Prebid also bids for video all the time. It can be easily disabled for non-video pages.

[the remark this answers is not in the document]
dmnsobczak: https://github.com/Wikia/ad-engine/blob/dev/src/ad-bidders/prebid/prebid-helper.ts#L35

src/platforms/shared/bidders/bidders-state.setup.ts
rafalgiemza: good idea

src/ad-products/utils/video.ts
```diff
@@ -0,0 +1,6 @@
+import { context, targetingService } from '@ad-engine/core';
```
dmnsobczak, line 1: This is used only in N&R: let's keep it in `src/platforms/news-and-ratings/shared` :)

rafalgiemza: moved :)

dmnsobczak (dismissed):

src/platforms/shared/bidders/bidders-state.setup.ts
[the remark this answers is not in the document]
rafalgiemza: follow up ticket: https://fandom.atlassian.net/browse/ADEN-12964

PR #1844, opened 2023-03-20
ADEN-12849 Change incontent ads injection policy

dmnsobczak:


src/core/services/slot-creator.ts
```diff
@@ -40,6 +42,8 @@ const groupName = 'slot-creator';
 
 @Injectable()
 export class SlotCreator {
+	private slotRepeater = new SlotRepeater();
```
nandy-andy, line 45: Consider initialising an instance of `SlotRepeater` in a different way, so it's easy to stub/mock it.

src/core/services/slot-creator.ts
```diff
@@ -178,6 +186,48 @@ export class SlotCreator {
 		placeholder.appendChild(div);
 	}
 
+	private setupSlotRepeat(slotConfig: Required<SlotCreatorConfig>): void {
```
nandy-andy, line 189: Consider renaming the method to `registerOnSlotRenderHandler()` as it describes better what it does ;-)

src/core/services/slot-repeater.ts
```diff
@@ -1,80 +1,56 @@
...
+			return;
+		}
 
-		return false;
-	}
+		context.set(`slots.${slotName}`, newSlotDefinition);
+		context.set(`slots.${slotName}.uid`, generateUniqueId());
 
-	context.set(`slots.${slotName}`, newSlotDefinition);
+		if (repeatConfig.updateProperties) {
```
nandy-andy, line 33: Consider moving the body of the condition to a separate method, so the whole `repeatSlot()` method is more readable.

src/core/services/slot-creator.ts
```diff
@@ -178,6 +186,48 @@ export class SlotCreator {
...
+		communicationService.onSlotEvent(
+			AdSlot.SLOT_RENDERED_EVENT,
+			({ slot }) => {
```
nandy-andy, line 192: Or if you extract the slot event handler to a separate method it will be easier to read and also you won't need `setupSlotRepeat()` nor its renamed version `registerOnSlotRenderHandler()` as you'll just use `communicationService.onSlotEvent()` with your extracted handler logic in the place you call now `setupSlotRepeat()` 🤔

src/platforms/news-and-ratings/metacritic-neutron/setup/dynamic-slots/metacritic-neutron-dynamic-slots.setup.ts
```diff
@@ -59,6 +60,14 @@ export class MetacriticNeutronDynamicSlotsSetup implements DiProcess {
 	}
 
 	private setupRepeatableSlot(slotName, slotNameBase = '') {
+		const slotRepeater = new SlotRepeater();
+
+		communicationService.onSlotEvent(AdSlot.SLOT_RENDERED_EVENT, ({ slot }) => {
```
nandy-andy, line 65: Why do we need the `communicationService.onSlotEvent()` handlers below if we're introducing the new way here? 🤔

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
```diff
@@ -57,7 +57,7 @@ export class CommonTags implements TargetingProvider<Partial<SlotTargeting>> {
 		const hasFeaturedVideo = !!videoStatus.hasVideoOnPage;
 		const hasIncontentPlayer =
 			!hasFeaturedVideo &&
-			!!document.querySelector(context.get('slots.incontent_player.insertBeforeSelector'));
+			!!document.querySelector(context.get('templates.incontentAnchorSelector'));
```
nandy-andy, line 60: Could you update [the unit tests for common-tags.ts](https://github.com/Wikia/ad-engine/blob/dev/spec/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.spec.ts) to cover this case?

src/platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
```diff
@@ -138,16 +137,11 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 			},
 			featured: {
 				adProduct: 'featured',
+				isVideo: true,
```
nandy-andy, line 140: ❤️

src/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -16,21 +15,9 @@ import {
...
 import { Injectable } from '@wikia/dependency-injection';
 
-interface SlotCreatorInsertionParamsType {
-	anchorSelector: string;
-	insertMethod: insertMethodType;
-}
-
 @Injectable()
 export class UcpMobileSlotsDefinitionRepository {
-	protected incontentsAnchorSelector = '.mw-parser-output > h2';
```
nandy-andy, line 26: ❤️

src/core/models/ad-slot.ts
```diff
@@ -18,9 +17,8 @@ export interface RepeatConfig {
 	slotNamePattern: string;
 	limit: number;
 	updateProperties: Dictionary;
-	additionalClasses?: string;
+	updateCreator?: Dictionary;
```
nandy-andy, line 20: It was weird for me to see that behind `updateCreator` name we expect an optional `Dictionary`. After reading the code I suggest renaming it to `creatorPropsToUpdate` or `creatorConfigToUpdate` 🤔

spec/core/services/slot-repeater.spec.ts
```diff
@@ -1,108 +1,44 @@
...
+import { SlotRepeater, targetingService, TargetingService } from '@wikia/core/services';
 import { expect } from 'chai';
+import { SinonStubbedInstance } from 'sinon';
 import { adSlotFake } from '../ad-slot-fake';
 
 describe('slot-repeater', () => {
 	let adSlot;
-	let injectedContainer;
-	let handleSlotRepeating;
+	let slotRepeater;
```
nandy-andy, line 8: With the new `SlotRepeater` we have only 2 unit tests out of 6. I think they were valid business cases to test:
* ad-slot is not repeated when it is disabled,
* ad-slot is not repeated when it is not configured as repeatable,
* ad-slot is not repeated when it is configured as repeatable and sibling is too close,
* ad-slot is repeated when it is configured as repeatable and sibling is far away.

Could you cover this test scenarios maybe with in testing dynamic-slots if it's not possible in with the new `SlotRepeater`?

nandy-andy (commented):
Overall it looks good -- less logic in too-big-anyway `AdSlot` class, less line of code = less potential for bugs. I left some small coding style/naming/unit testing comments and I have two concerns:
1. Didn't we want to AB test the changes (the description in the ticket mentions AB testing)? If we wanted the code is not ready for AB testing 🙈 
2. Shouldn't we update all N&R bundles for repeatable slots? My understanding is that now even if backend responses with divs for ad slots we need to run through them with `SlotCreator` or `SlotRepeater` (if these are repeatable slots) to alter the divs from backend.

src/core/services/slot-creator.ts
dmnsobczak: Good idea, moved into constructor 👍

src/platforms/news-and-ratings/metacritic-neutron/setup/dynamic-slots/metacritic-neutron-dynamic-slots.setup.ts
dmnsobczak: Old repeatable slots logic had its own logic and it was listening for `SLOT_RENDERED_EVENT` to create a slot definition in the context before pushing it into the ads queue. New logic requires use of SlotCreator to trigger slot repeat. At the same time we are testing platform events in TVGuide and the same system will be applied to Metacritic Neutron, which is also a Neutron app. It's a waste of time to reimplement the slot injection logic if in the near future we'll going to change it again, probably unify it with TVGuide logic. So that's why I extracted removed code here as a temporary hack/workaround to keep repeatable slots on Metacritic and this code will be replaced with a solid SlotCreator usage once N&R team add events to Metacritic Neutron :)

src/core/services/slot-repeater.ts
```diff
@@ -1,80 +1,56 @@
...
-	});
-}
-
-function repeatSlot(adSlot: AdSlot): boolean {
-	const newSlotDefinition = adSlot.getCopy();
-	const repeatConfig = newSlotDefinition.repeat;
-
-	repeatConfig.index += 1;
+export class SlotRepeater {
+	repeatSlot(adSlot: AdSlot, repeatConfig: RepeatConfig): string {
```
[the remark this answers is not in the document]
dmnsobczak: The slot name also served this purpose here: `Boolean(slotName) === true` - no slot name means that slot repeat was not success, for example because of repeat limit. This slot name is then used in `SlotCreator` and I did it that way not to extract any more logic from `SlotRepeater` to `SlotCreator` while keeping `context` unused - strict rules are required to finally reach the point where context is no longer a God Object with everything :/

spec/core/services/slot-repeater.spec.ts
dmnsobczak: Some of those test were testing just a slightly different variants and were covering the same code. Some were checking side effects of `SlotRepeater` class which now I removed. But I do agree that some tests can be added to `SlotCreator` - I will do this in the separate ticket, because I want to unlock the QA stage asap and first I need to do some other things today :( 
I'll ping you when I'll push the ticket and I'll keep this comment open - resolving all is required to merge the pull request 🤞

src/core/models/ad-slot.ts
dmnsobczak: Same naming convention we have for `updateProperties` :D Creator is basically an object with all data required by `SlotCreator` to handle AdSlot creation. The one of few usages of this type points to a `SlotCreator` class where object creation-destructuration explains what happens to this Dictionary. Also when editing it you need to jump and check possible `SlotCreatorConfig` params, so it's hard to be confused 🤔

src/core/services/slot-repeater.ts
dmnsobczak: Done.

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
dmnsobczak: Yes, I'll add unit tests in the next commit.

src/core/services/slot-creator.ts
dmnsobczak: I don't like connecting everything with events and from perspective of somebody willing to use this creator - I don't need to know that this is a on render ended handler registering, `setupSlotRepeat()` will just setup slot repeat and handle this, no more details required. Also if I rename it like that there will be: `if (slotConfig.repeat) then registerOnSlotRenderHandler` - why registering on render ended handler if `SlotCreatorConfig` has `repeat` property? For me this change is worse descriptive 🤔

src/core/services/slot-creator.ts
dmnsobczak: Described above - I like the `setupSlotRepeat` name, it's easy to understand. Also extracting `communicationService` callback to a separate method would require a binding to pass `slotConfig` and I think this syntax is complicated, although fancy.

nandy-andy: How do you understand `setupSlotRepeat()`? My point is that it's not easy to understand an async logic stands behind this name. Let's reach out to the team and see if it's only me -- this will help us come to with a team-oriented solution.

dmnsobczak:
@nandy-andy 
1. Description is incoherent - first part mentions AB tests, but rollout plan says that we will roll this out geo by geo, which can be done using AELoader. I'll discuss it with Lisandro. I prefer going without ICBM variable, because most of this code is a refactor that can't be icVared and should be tested well and functional part is easy to revert (since it's mostly selectors and new `incontent_leaderboard` repeat config).
2. I updated all N&R platforms that handles repeatable slots 🤔 
4. I'll add more unit tests in the next commit - goals for 2023 first 🙈

src/core/models/ad-slot.ts
nandy-andy: Imagine a name when you don't need to jump to other classes to see creation-destruction logic to understand what's going on ;-)

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
nandy-andy: You mean the next ticket, a follow-up, don't you? :)

nandy-andy (dismissed):
Let's test it and the unit test will be addressed in a follow-up ticket @dmnsobczak will create 🤞

src/core/services/slot-creator.ts
Jorewin: I don't know how the rest of the codebase compares to this, but this could be an important factor. When I look at this function it looks more like piece of code was removed just so everything is not in one place, because the important logic itself is hidden after a few indentations. I also don't think I would be able to tell what `setupSlotRepeat` does in this context, without looking at its implementation, so the name argument doesn't speak to me. With that said I agree that using binding as a general practise can introduce many errors so I would be wary of implementing it that way. I would stick with the proposed solution or, more preferably, implement a function returning a callback that can be passed to `onSlotEvent`. What do you think about that?

kwfandom (approved):

PR #1851, opened 2023-03-21
ADEN-12944 | backfill top nav slots

mfarul:


kwfandom (approved):

src/platforms/news-and-ratings/metacritic/setup/context/slots/metacritic-slots-context.setup.ts
```diff
@@ -105,6 +111,8 @@ export class MetacriticSlotsContextSetup implements DiProcess {
 			},
 			mobile_nav_ad_omni: {
 				defaultSizes: [
+					[320, 100],
+					[320, 50],
 					[5, 5],
 					[6, 6],
```
kwfandom, line 117: i believe [6, 6] is unnecessary here

spec/platforms/shared/setup/slots.config.extender.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+			},
+			slot2: {
+				targeting: {
+					s0: 'ent',
+					genre: 'gaming',
+				},
+			},
+			slot3: {},
+		};
+		context.set('slots', slots);
```
nandy-andy, line 122: We could add `afterEach()` hook and reset the `context`, so it doesn't pollute the rest of the tests. Unless after "karpik changes" this practise isn't needed anymore 🤔

src/platforms/shared/setup/slots.config.extender.ts
```diff
@@ -0,0 +1,57 @@
...
+export class SlotsConfigurationExtender implements DiProcess {
+	constructor(private instantConfig: InstantConfigService) {}
+
+	execute(): Promise<void> | void {
+		const slotsConfig: Dictionary<BaseSlotConfig> = this.instantConfig.get('icSlotsConfig');
+
+		if (slotsConfig) {
+			SlotsConfigurationExtender.extendSlotsConfiguration(context.get('slots'), slotsConfig);
+		}
+		return undefined;
```
nandy-andy, line 20: Why do we need to return `undefined` here? 🤔

src/platforms/shared/setup/slots.config.extender.ts
```diff
@@ -0,0 +1,57 @@
...
+				SlotsConfigurationExtender.extendBaseSlotConfiguration(
+					slots[slotName],
+					slotsConfigExtension[slotName],
+				);
+			}
+		});
+	}
+
+	private static extendBaseSlotConfiguration(slot: BaseSlotConfig, config: BaseSlotConfig) {
+		if (Array.isArray(config.defaultSizes)) {
```
nandy-andy, line 38: We could create a follow-up to extend it and handle GPT capabilities fully. I think we only use it in f2: https://developers.google.com/publisher-tag/guides/ad-sizes#responsive_ads

src/platforms/shared/setup/slots.config.extender.ts
nandy-andy, line 1: I wanted to suggest renaming it to `slots-config-extender.setup.ts` but after looking at [the directory](https://github.com/Wikia/ad-engine/tree/912f013bc436290e9069f724e679daf910d1e289/src/platforms/shared/setup) I don't feel that strong as the convention seem not to be followed anymore 🤷‍♂️ I'm leaving it up to you to decide which way to go ;-)

src/platforms/shared/setup/slots.config.extender.ts
mfarul: good catch - that was generated by IDE when implementing interface ;)

src/platforms/shared/setup/slots.config.extender.ts
mfarul: I was thinking even to go with deep merge of the whole config, but this will be tricky and might 
a) produce outages if someone will brake the slot structure 
b) lead to overuse of ICBM config -> Fastly costs
So i just decided to add 2 fields handled more carefully.

nandy-andy: Good point. Maybe we could add a comment somewhere that the extender is suppose to be use only for small tweaks and once we're done with the tweaks we should replace the ICBM config with actual code in the platform's files 🤔

nandy-andy (approved):

PR #1853, opened 2023-03-22
ADEN-12905 - Added AgeGate handling

domjanik:
In this PR I've refactored a bit our approach to detecting COPPA subjects. Currently, we were just reading wiki.targeting.directedAtChildren property from context. After introducing AgeGate ( and maybe over_18 in the future ) this approach is not enough, as it can't be easily extended. I've created a new method - isCoppaSubject, which checks LocalStorage for age_gate object, and if there is such object - it will parse it. If there is not - it will return wiki.targeting.directedAtChildren as it used to.


src/core/utils/is-coppa-subject.ts
```diff
@@ -0,0 +1,21 @@
...
+	timestamp: number;
+}
+
+export function isCoppaSubject(): boolean {
+	try {
+		const ageGateResult = localStorage.getItem('age_gate');
+		if (ageGateResult) {
+			const parsedAgeGateResult = JSON.parse(ageGateResult) as AgeGateResult;
+			return parsedAgeGateResult?.dialogShown
+				? !parsedAgeGateResult.adult
```
szymon-nawrocki, line 15: If `adult` property is not available it means that someone clicked "I am a kid". In this case the code should return true (but currently it will throw an error)

szymon-nawrocki: Okay it won't throw an error. It will return undefined, and `!undefined = true` 🙈

szymon-nawrocki (approved):

src/core/utils/is-coppa-subject.ts
```diff
@@ -0,0 +1,21 @@
...
+
+interface AgeGateResult {
+	dialogShown: boolean;
+	adult?: boolean;
+	timestamp: number;
+}
+
+export function isCoppaSubject(): boolean {
+	try {
+		const ageGateResult = localStorage.getItem('age_gate');
```
jbienkowski-fandom, line 11: In the meantime, we have changed the storage of AgeGate result from localStorage to cookie: https://github.com/Wikia/identity-engine/pull/50

Moreover, we could use the same feature flag as in the IdentityEngine which is `icAgeGate` in this util function (if the flag is set to false, just take the value from `wiki.targeting.directedAtChilden` context).

jbienkowski-fandom (changes requested):

src/core/utils/is-coppa-subject.ts
```diff
@@ -0,0 +1,18 @@
+import { context, CookieStorageAdapter } from '../services';
+
+export function isCoppaSubject(): boolean {
+	if (!context.get('services.ageGateHandling')) {
+		return context.get('wiki.targeting.directedAtChildren');
+	} else {
```
jbienkowski-fandom, line 6: else is redundant

src/core/utils/is-coppa-subject.ts
```diff
@@ -0,0 +1,18 @@
...
+		try {
+			const cookieStorage = new CookieStorageAdapter();
+			const ageGateResult = cookieStorage.getItem('age_gate');
+			if (ageGateResult) {
+				const [dialogShown, adult] = ageGateResult.split('|');
+				return dialogShown ? adult === '0' : context.get('wiki.targeting.directedAtChildren');
```
jbienkowski-fandom, line 12:
```suggestion
				return dialogShown === '1' ? adult === '0' : context.get('wiki.targeting.directedAtChildren');
```

src/core/utils/is-coppa-subject.ts
```diff
@@ -0,0 +1,18 @@
...
+			}
```
jbienkowski-fandom, line 13: if there is no cookie we should return something as well

jbienkowski-fandom (approved):

PR #1814, opened 2023-03-07
ADEN-12861 Refactor of TVGuide slots logic based on events

dmnsobczak:

PR #1855, opened 2023-03-22
ADEN-12863 - add observer for galleries, to rotate ads together with …

adi-wesolowski-fandom:
…gallery content + fix destroy-event name


src/core/models/ad-slot.ts
```diff
@@ -75,6 +75,7 @@ export class AdSlot {
 	static SLOT_LEFT_VIEWPORT = 'slotLeftViewport';
 	static SLOT_STATUS_CHANGED = 'slotStatusChanged';
 	static DESTROYED_EVENT = 'slotDestroyed';
+	static DESTROY_EVENT = 'slotDestroy';
```
David-Iacovone, line 78: Is this meant to replace 'slotDestroyed' ?

David-Iacovone (approved):

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
```diff
@@ -0,0 +1,50 @@
...
+			return;
+		}
+
+		this.currentUrl = location.href;
+
+		this.slotsNamesToHandle.forEach((slotNameToHandle) => {
+			slotService.get(slotNameToHandle).destroy();
+		});
+
+		communicationService.onSlotEvent(AdSlot.DESTROYED_EVENT, ({ slot }) => {
```
andrzejkupczyk, line 43:
```suggestion
		communicationService.onSlotEvent(AdSlot.DESTROY_EVENT, ({ slot }) => {
```

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
```diff
@@ -0,0 +1,50 @@
...
+			return;
+		}
+
+		const observer = new MutationObserver(() => this.handleMutation());
+
+		observer.observe(elementToObserveMutation, config);
+	}
+
+	private handleMutation() {
+		if (!this.currentUrl) {
```
andrzejkupczyk, line 28: Out of pure curiosity, when will it be true? According to [ln#14](https://github.com/Wikia/ad-engine/pull/1855/files#diff-879ef870fd8aa9d7993c5f2948d46ccc520f0f7132d9ddb097a4dfbf2198a968R14) `currentUrl` should always be set, shouldn't it? 🤔

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
andrzejkupczyk, line 1: Have you considered keeping consistent file naming with the `.setup` suffix?

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
adi-wesolowski-fandom: No. I divided DESTROY and DESTROYED for a reason. Here, should be DESTROYED, because we want to create new slot _after_ destroying is complete, not before or there will be collision in GPT.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
adi-wesolowski-fandom: Should. I copyied the basic idea from SeamlessObserver, and they have it there also. And I think, its ok to be here. I believe, there could be (very rare) situation, where location.href could give as "null". But I cant tell and image, how it could happen in this case. So maybe you are right and this can be deleted.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
adi-wesolowski-fandom: I can rename but I believe this is not a classic setup, dont affect on the AdEngine in the matter of lib behaviour. Its more like domain exception and observer itself. Just like TvGuidePageChangeAdsObserver is included in the TVguide domain.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
andrzejkupczyk: Oh, my mistake - I thought this is the same event.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
andrzejkupczyk: Well, if the `location.href` could be null (according to docs [it's always a string](https://developer.mozilla.org/en-US/docs/Web/API/Location/href#value)), then there are probably more problems with this code.

Imagine what would happen if:

this.currentUrl == null
location.href == null

In ☝🏻 case, the `this.currentUrl` will get a null value either way.

adi-wesolowski-fandom: Yes indeed, there will be null. Maybe thats why we got if ;) 

But ok, sure, I will remove it then.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
adi-wesolowski-fandom: This is a place where all observers are put in other platforms directories so I followed this pattern. 
I will add "setup" suffix for you ;)

dmnsobczak: Filenames was unified in this ticket: https://github.com/Wikia/ad-engine/pull/1814
Please match this convention and avoid conflict while merging :D

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
dmnsobczak: I guess the problem here was with the events order. Communication service handles callbacks in the order they were registered - that's why this callback is executed before the one in the `gpt-provider` - second one waits for GPT load. You can solve this by postponing this callback register too, for example by wrapping it with this:
`communicationService.on(eventsRepository.AD_ENGINE_STACK_START, () => {
    communicationService.onSlotEvent(AdSlot.DESTROY_EVENT, ({ slot }) => {});
});`
Then you don't need to duplicate events and change their logic 🤔

andrzejkupczyk (dismissed):

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
adi-wesolowski-fandom: I did not change logic for existing events, I just rename it to be more relevant to what's going on there (action to mark slot TO BE destroyed on the event (already) "Destroyed", seems weird to me. So I rename to, to future tense. 

And I used old name in Past tense, to message thing that happened: slots has been destroyed. And then I can create a new one - just like we discuss on slack :) 

Making this listener inside listener does not seems to be clear (I mean, it might not be in the future)

andrzejkupczyk: Introducing a new event can provide more flexibility and improve readability, so if it's not an invalid approach, I'll stick with it.

src/core/models/ad-slot.ts
```diff
@@ -78,6 +78,7 @@ export class AdSlot {
 	static SLOT_LEFT_VIEWPORT = 'slotLeftViewport';
 	static SLOT_STATUS_CHANGED = 'slotStatusChanged';
 	static DESTROYED_EVENT = 'slotDestroyed';
```
David-Iacovone, line 80: Is this event still in use?

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.setup.ts
```diff
@@ -0,0 +1,64 @@
...
+		if (!elementToObserveMutation) {
+			return;
+		}
+
+		const observer = new MutationObserver(() => this.handleMutation());
+
+		observer.observe(elementToObserveMutation, config);
+	}
+
+	private handleMutation() {
```
David-Iacovone, line 34: I would add a return type here just to keep it clear

David-Iacovone (changes requested):
Generally, it looks good. I would just consider these few minor tweaks before merging

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.setup.ts
adi-wesolowski-fandom: ok, done

David-Iacovone (approved):

andrzejkupczyk (approved):

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.setup.ts
```diff
@@ -0,0 +1,64 @@
...
+					slotNameToHandle,
+					true,
+				);
+			});
+		});
+	}
+
+	private handleSlotDestroyed(eventData): void {
+		const slot: AdSlot = eventData.slot;
+		if (!this.slotsNamesToHandle.includes(slot.getSlotName())) {
```
Jorewin, line 57: Is this still necessary if the handler is subscribed to the `AdSlot.DESTROYED_EVENT` on `slotNameToHandle` slot? I don't see a possibility of it running with a slot outside of `slotNamesToHandle` list.

Jorewin (dismissed):
One question, but other than that it looks good to me.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.setup.ts
adi-wesolowski-fandom: It may not. It was done, because we didnt wanted to handle other on "destroyed" event. But calling it in onSlotEvent was simpler. I will check if we can remove it safely and do so

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
```diff
@@ -0,0 +1,50 @@
...
+		const config = { childList: true, subtree: false };
+		this.currentUrl = location.href;
+
+		const elementToObserveMutation = document.querySelector(this.elementToObserveMutationSelector);
+
+		if (!elementToObserveMutation) {
+			return;
+		}
+
+		const observer = new MutationObserver(() => this.handleMutation());
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: this change causing an error on "forEach" in "handleMutation". Propably on "this.slotNamesToHandle" could not see it. So I leave it as it is.

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.setup.ts
adi-wesolowski-fandom: Ok, removed

src/platforms/news-and-ratings/metacritic/setup/page-observers/metacritic-page-change-gallery-observer.ts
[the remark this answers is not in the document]
andrzejkupczyk: Oh, this is strange 🤔  I'm sorry for the trouble! 🙇🏻

andrzejkupczyk (approved):

PR #1868, opened 2023-03-30
ADEN-12977 add sampling service for outbound traffic for DW track req…

adi-wesolowski-fandom:
…uests

### DIFF
--- a/src/core/utils/outbound-traffic-restrict.ts
+++ b/src/core/utils/outbound-traffic-restrict.ts
@@ -1,0 +1,38 @@
+import { utils } from '../index';
+import { context } from '../services';
+
+class OutboundTrafficRestrict {
+	private static MIN = 0;
+	private static MAX = 100;
+	private static DEFAULT_THRESHOLD = 2;
+
+	public isOutboundTrafficAllowed(serviceName = 'default'): boolean {
+		const isAllowedFromContext = context.get(`services.${serviceName}.allowed`);
+		if (typeof isAllowedFromContext === 'boolean') {
+			return isAllowedFromContext;
+		}
+
+		const isAllowed =
+			this.getSeed() <
+			(context.get(`services.${serviceName}.threshold`) ||
+				OutboundTrafficRestrict.DEFAULT_THRESHOLD);
+
+		utils.logger(
+			'outbound-traffic-restrict',
+			`Outbound traffic for: "${serviceName}" is allowed: ${isAllowed}`,
+		);
+
+		context.set(`services.${serviceName}.allowed`, isAllowed);
+
+		return isAllowed;
+	}
+
+	private getSeed(): number {
+		return (
+			Math.random() * (OutboundTrafficRestrict.MAX - OutboundTrafficRestrict.MIN) +
+			OutboundTrafficRestrict.MIN
+		);
+	}
+}
+
+export const outboundTrafficRestrict = new OutboundTrafficRestrict();
--- a/src/platforms/shared/setup/tracking-parameters.setup.ts
+++ b/src/platforms/shared/setup/tracking-parameters.setup.ts
@@ -59,5 +59,10 @@ export class TrackingParametersSetup implements DiProcess {
 		const trackingParameters = await this.getTrackingParameters(legacyEnabled);
 
 		context.set('wiki', { ...context.get('wiki'), ...trackingParameters });
+
+		context.set(
+			'services.dw-tracker.threshold',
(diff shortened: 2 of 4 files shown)


src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -59,5 +59,10 @@ export class TrackingParametersSetup implements DiProcess {
 		const trackingParameters = await this.getTrackingParameters(legacyEnabled);
 
 		context.set('wiki', { ...context.get('wiki'), ...trackingParameters });
+
+		context.set(
+			'services.dw-tracker.threshold',
+			this.instantConfig.get('icDwTrackerTrafficThreshold') ?? 50,
```
andrzejkupczyk, line 65: Could we move this magic number into a properly named const such as `DEFAULT_TRAFFIC_THRESHOLD`?

src/core/utils/outbound-traffic-restrict.ts
```diff
@@ -0,0 +1,38 @@
...
+		context.set(`services.${serviceName}.allowed`, isAllowed);
+
+		return isAllowed;
+	}
+
+	private getSeed(): number {
+		return (
+			Math.random() * (OutboundTrafficRestrict.MAX - OutboundTrafficRestrict.MIN) +
+			OutboundTrafficRestrict.MIN
+		);
```
andrzejkupczyk, line 34: It would be great if we could improve the readability of this code 🙏🏻

At first glance, these calculations could be simplified to `Math.random() * OutboundTrafficRestrict.MAX` since `100 - 0` gives `100`, and `x + 0` gives `x`. In conjunction with inconsistent code formatting (each or none of the operations should be on a separate line),  I'm afraid this can be hard to read and understand.

adi-wesolowski-fandom: Yes, you are right. It is silly right now. I wanted to give a possibility to change minimal and maximum of the new randomized value range. but it is no longer needed, because we wont change the scope, as it if fixed to 0-100%.

andrzejkupczyk (approved):

src/core/utils/outbound-traffic-restrict.ts
```diff
@@ -0,0 +1,34 @@
+import { utils } from '../index';
+import { context } from '../services';
+
+class OutboundTrafficRestrict {
+	private static MAX = 100;
+	private static DEFAULT_THRESHOLD = 2;
+
+	public isOutboundTrafficAllowed(serviceName = 'default'): boolean {
+		const isAllowedFromContext = context.get(`services.${serviceName}.allowed`);
```
jakinluk, line 9: I am thinking how it can be tested from e2e perspective sice if it is disabled it is disabled for the whole session.

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -19,6 +19,10 @@ export class DataWarehouseTracker {
 	 * Call all of the setup trackers
 	 */
 	track(options: TrackingParams, trackingURL?: string): void {
+		if (!utils.outboundTrafficRestrict.isOutboundTrafficAllowed('dw-tracker')) {
```
jakinluk, line 22: according ticket description I think restriction should apply only to requests made to https://beacon.wikia-services.com/__track/special/adengloadtimes
so I think sampling should be applied to LoadTimesTracker

jakinluk:
what about unit tests?

src/platforms/shared/tracking/data-warehouse.ts
adi-wesolowski-fandom: IMO, to get satisfying results, we should reduce all of the trafic, not only 10% of this traffic.

adi-wesolowski-fandom: lets wait till monday for bigger discution

src/core/utils/outbound-traffic-restrict.ts
adi-wesolowski-fandom: I think, disabling whole session in needed, to keep consistency of data in DW, for each slot, etc. 

as for e2e and other testing matter, you are right, I will add some params to give ability to set it up

src/platforms/shared/tracking/data-warehouse.ts
adi-wesolowski-fandom: i made changes, and now, every track can be configured separately

jakinluk (approved):
thx for implementing all remarks

PR #1869, opened 2023-03-31
ADEN-12988 Fix UAP glitch on mobile Safari

nandy-andy:


src/platforms/ucp-mobile/templates/handlers/slot/mobile-slot-size-impact-to-resolved-handler.ts
```diff
@@ -0,0 +1,34 @@
+import { UapDomManager } from '@platforms/shared';
+import { DomListener, tapOnce, TemplateStateHandler } from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+import { merge, Subject } from 'rxjs';
+import { startWith, takeUntil, tap } from 'rxjs/operators';
+
+@Injectable({ autobind: false })
+export class MobileSlotSizeImpactToResolvedHandler implements TemplateStateHandler {
```
nandy-andy, line 8: I thought of just extending `SlotSizeImpactToResolvedHandler` here instead of copy&pasting most of the code but after second thoughts I decided not to do it. Inheritance here could bring more pain if we'd have decided to change anything for desktop in `SlotSizeImpactToResolvedHandler` we'd need to make sure it behaves properly on mobile.

src/platforms/ucp-mobile/templates/handlers/slot/mobile-slot-size-impact-to-resolved-handler.ts
```diff
@@ -0,0 +1,34 @@
...
+				startWith({}),
+				tap(() => this.manager.setSlotHeightImpactToResolved()),
+				takeUntil(this.unsubscribe$),
+			)
+			.subscribe();
+
+		this.domListener.scroll$
+			.pipe(
+				tapOnce(() => {
+					this.manager.setImpactImage();
```
dmnsobczak, line 25: If I remember correctly this case is required only for UAP going back to impact, for example after clicking `Replay` on desktop. On mobile though there is no different image for impact, because impact is a video with 100% width and thumbnail below that handles the placeholder role when video still loads, so there is just resolved image added to the creative.
Creative template suggest it differently: https://github.com/Wikia/dfp-creative-templates/blob/master/uap-hivi.html
But actually Adops don't add mobile impact images: 
https://admanager.google.com/5441#creatives/creative/detail/line_item_id=4973088146&creative_id=138261896621
I would go a little bit further a get rid of impact image for mobile UAP. This could be the value required for conditional logic in UAP 🤔

dmnsobczak (changes requested):
Something doesn't seems right. I would try more brave changes with removing mobile impact image from the configuration if UAP is in mobile mode.
But, this can be done in the follow up if we won't split the handlers logic to a different files. RxJS allows many way to do this in the same, conditional pipe without losing a readability. We spent lot of time to unify UAP logic, make it really universal and get rid of branches, alternative handlers for this, additional steps for that or hacked logic there. If you want, I can prepare a commit with my proposition, but I need your help with testing it 🤔

src/platforms/ucp-mobile/templates/handlers/slot/mobile-slot-size-impact-to-resolved-handler.ts
nandy-andy: I couldn't reproduce it on f2.

nandy-andy:
> If you want, I can prepare a commit with my proposition, but I need your help with testing it 🤔
Be my guest and implement changes you're thinking about. You can test it on your local machine with Safari mobile emulation. The only downside is that `?adengine_version=local` does not work and once you have changes you think will work you need to push them to s3 and then use proper `?adengine_version`

dmnsobczak (dismissed):

dmnsobczak (approved):

src/platforms/shared/templates/handlers/slot/slot-size-impact-with-placeholder-handler.ts
```diff
@@ -11,7 +11,19 @@ export class SlotSizeImpactWithPlaceholderHandler implements TemplateStateHandle
 	constructor(private domListener: DomListener, private manager: UapDomManager) {}
 
 	async onEnter(): Promise<void> {
-		this.manager.setImpactImage();
+		if (context.get('state.isMobile')) {
+			this.domListener.scroll$
```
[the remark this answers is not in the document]
dmnsobczak: Done!

PR #1870, opened 2023-03-31
ADEN-12909 remove legacy tracking context

jakubbanas:
Since legacy tracking context is disabled in all the platform globally, we could remove it completely.

Coverage is lower, because there are fewer lines to cover 🤷


src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -1,62 +1,16 @@
...
-		};
-	}
-
-	private async getTrackingParameters(
-		legacyEnabled: boolean,
-	): Promise<Partial<ITrackingParameters>> {
-		return legacyEnabled
-			? this.getLegacyTrackingParameters()
-			: await this.getNewTrackingParameters();
+		return window.fandomContext.tracking;
```
jbienkowski-fandom, line 9: this way we return object by reference (and it may get mutated somewhere else), using object destructuring creates shallow object copy

jakubbanas: Good point, I thought that it would not be a problem, but I reverted it just to be safe.

jbienkowski-fandom (dismissed):

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -1,62 +1,16 @@
...
-		legacyEnabled: boolean,
-	): Promise<Partial<ITrackingParameters>> {
-		return legacyEnabled
-			? this.getLegacyTrackingParameters()
-			: await this.getNewTrackingParameters();
+		return { ...window.fandomContext.tracking };
 	}
 
 	async execute() {
-		const legacyEnabled = !this.instantConfig.get('icDisableLegacyTrackingParameters', false);
```
szymon-nawrocki, line 58: This ICBM variable can be removed then: https://services.fandom.com/icbm/dashboard/feature_flag/icDisableLegacyTrackingParameters

spec/platforms/shared/setup/tracking-parameters.setup.spec.ts
```diff
@@ -15,22 +15,8 @@ describe('TrackingParametersSetup', () => {
...
-
-		assert.called(cookiesGetStub);
-	});
-
-	it('should set new context when flag is set to true', () => {
-		const instantConfig = {
-			get: () => true,
-		} as any;
-		const experimentSetup = new TrackingParametersSetup(instantConfig);
+	it('should always set new context', () => {
```
szymon-nawrocki, line 18: There is no concept of "new" and "old" context anymore, so this test case can be removed. Instead of that I'd add a test case checking that tracking parameters are being set properly

src/platforms/shared/setup/tracking-parameters.setup.ts
jakubbanas: We could remove it after releasing this version. For now, we need it to be set to `true`, removed variable is considered `false`.

spec/platforms/shared/setup/tracking-parameters.setup.spec.ts
jakubbanas, line 18:
```suggestion
	it('should always set tracking context', () => {
```

jakubbanas: Actually, setting proper parameters is not AdEngine's responsibility any more, IdentityEngine take care of it.
AdEngine just need to set what IE provides, so IMO we only need to test if parameters are set at all.

I just renamed the test. What do you think?

szymon-nawrocki: For me this case is no longer relevant, because it tests whether the Cookie.get was called. Class TrackingParametersSetup does not even use cookies anymore, so this test does not make much sense.

Instead I thought about adding a test that:
* mocks `getTrackingParameters()`
* and checks if `execute()` method sets context's wiki object properly based on that mock 

@jakubbanas What do you think?

jakubbanas: Good point 👍

szymon-nawrocki (dismissed):

jakubbanas:
> 

Actually, we do set `wiki.pvUID` here (https://github.com/Wikia/ad-engine/blob/267d191cde1b128bdcf00589672ce1b7eaca4e99/src/platforms/shared/setup/tracking-parameters.setup.ts#L16)
But maybe the code you mentioned is executed before that.

PR #1872, opened 2023-04-03
Aden 12962

adi-wesolowski-fandom:


src/core/utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+		this.count = 0;
+
+		this.start();
+	}
+
+	public isAggregatorActive(): boolean {
+		return this.enabled;
+	}
+
+	public push(track: string, url: string, element: DataWarehouseParams): void {
```
jakinluk, line 31: what about reducing parameter no to 
`(url: string, element: DataWarehouseParams)`
 since there must be exactly one queue per URL exposing a queue name through the interface may cause misuse.
In other words I think a queue name extraction from URL is a responsibility of this class

src/core/utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+
+		window.addEventListener('visibilitychange', () => {
+			if (document.hidden) {
+				this.fireAggregatedQueue();
+			}
+		});
+	}
+}
+
+export const dwTrafficAggregator = new DwTrafficAggregator();
```
jakinluk, line 104: for consistency, we can use dependency injection (@Injectable())

src/platforms/shared/setup/tracking-urls.ts
```diff
@@ -0,0 +1,103 @@
...
+	url: string;
+	config: TrackingConfig;
+}
+
+export const trackingUrls: Dictionary<TrackingUrl> = {
+	AD_ENG_LOAD_TIMES: {
+		name: 'AdEngLoadTimes',
+		url: 'https://beacon.wikia-services.com/__track/special/adengloadtimes',
+		config: {
+			threshold: true,
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: yes, we can do that. I wasnt sure about this names, and tried to avoid camelcase names. Maybe change "config" to "allowed", and then change "threshold" to "sampling", leaving "aggregation" as is?

src/core/utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: We can try to reduce, but I dont agree that this responsibility should be put here, it break solid "s" rule - too much would be inserted here. And also, some other parts of library may also use it now, or in the near future (eg. monitoring task)

adi-wesolowski-fandom: I'll try to reduce no of params thou

jakinluk: I disagree that with my suggestion we break SRP. But we indeed break encapsulation by allowing the external consumers to affect the aggregation mechanism. Determining dictionary keys values internal implementation details of DwTrafficAggregator.

src/core/utils/dw-traffic-aggregator.ts
jakinluk: You are right.
However, utils might not be the best place for that class. DwTrafficAggregator name and DataWarehouseParams type make this class specific to aggregate DW-related traffic. Can we keep it next to DataWarehouse where it is used?

src/core/utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: I stand by my opinion, this would affect of single responsibility of this class.... but!, I have solution that (i hope) will satisfy both of us. See now, after my last commits - now there is now two params and no messing with srp - now there is passing full TrackingUrl object, that include all data and nothing will go messy - and this also should force, to be less prone to errors

src/core/utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: I moved aggregator class closer to data-warehouse

jakinluk (approved):

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+		this.sendRequest(dwTrackingUrlV2, paramsAggregated);
+	}
+
+	private sendRequest(url: string, params: DataWarehouseParams[]): void {
+		fetch(url, {
+			method: 'POST',
+			body: JSON.stringify(params),
+		}).then((response) => {
+			if (response.status === 200) {
+				utils.logger(`DW - Track Aggregated Success`, { url, params });
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: i used the same call as it is in the "data-warehouse", with just some changes: use fetch instead of http-xml-thing and fixed param in logger to "aggregated", instead of event, which IMO is unnecessary. But sure, I will check it

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -59,12 +70,20 @@ export class DataWarehouseTracker {
...
 			...options,
 			...this.getTimeBasedParams(),
 		};
-		const url = this.buildDataWarehouseUrl(params, trackingURL);
+
+		if (
+			context.get(`services.dw-tracker-${trackingURL.name.toLowerCase()}.aggregate`) &&
+			dwTrafficAggregator.isAggregatorActive()
+		) {
+			dwTrafficAggregator.push(trackingURL, params);
```
mfarul, line 83: shouldn't we return from here if the aggregator will be also responsible for sending the event

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
+import { Dictionary, utils } from '@wikia/ad-engine';
+import { TrackingUrl } from '../../setup/tracking-urls';
+import { DataWarehouseParams } from '../data-warehouse';
+
+interface AggregateQueue {
+	track: TrackingUrl;
+	params: DataWarehouseParams[];
+}
+
+class DwTrafficAggregator {
```
mfarul, line 10: Unit test is missing

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+	private fireAggregatedQueue(): void {
+		const keys = Object.keys(this.queue);
+		keys.forEach((queueKey) => {
+			this.fireAggregatedQueueByTrack(this.queue[queueKey].track);
+		});
+	}
+
+	private fireAggregatedQueueByTrack(track: TrackingUrl): void {
+		const paramsAggregated = [];
+		this.queue[track.name]?.params?.forEach((options) => paramsAggregated.push(options));
```
mfarul, line 54: shouldn't we mark each parameters list per event so that backend will be able to deduce where the next event is starting?

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	public push(track: TrackingUrl, element: DataWarehouseParams): void {
+		if (!this.queue[track.name]) {
+			this.queue[track.name] = {
+				track,
+				params: [],
+			};
+		}
+		this.queue[track.name].params.push(element);
```
mfarul, line 38: what about 2 events with the same name? for example impression for slot?

mfarul (changes requested):

src/platforms/shared/tracking/data-warehouse.ts
adi-wesolowski-fandom: yes, we should. I want to place a `return` here, when DW separation will be ready. Until that moment, putting `return` here will cause losing data, as it will be pushed only to new endpoints, which is not parsing yet.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: I will add

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: our (my) observation (mentioned in task) made me think that they are often single or only double, and quite distant from each other in time, and it was also hard for me to decide how best to combine them (combine by event-type, like `requested` or by slot name like `top-leaderboard`?). 

This approach seems to be way more efficient to achieve what we want: cost reduction.

mfarul: ok, so the queue is actually not aggregating all events, but it is a map of tracks that will be aggregated. I would change the field name to make it more clear, f.e. trackQueue

src/platforms/shared/tracking/data-warehouse.ts
mfarul: pls add a comment to indicate the code is incomplete

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
mfarul: ok, so my thinking was from a misunderstanding of the queue field purpose
we might change its name to indicate that it only aggregates events of the same type (track)

mfarul (approved):

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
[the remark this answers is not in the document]
jakinluk: Of course, it can play the role of logGroup but from the source code, I see the convention is the log group is a string without whitespaces to be easily enabled via query params.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -0,0 +1,104 @@
...
+
+interface AggregateQueue {
+	track: TrackingUrl;
+	params: DataWarehouseParams[];
+}
+
+class DwTrafficAggregator {
+	private readonly AGGREGATION_SECONDS_OFF = 5;
+	private readonly SEND_DATA_INTERVAL = 1000;
+	private readonly AGGREGATION_LIMIT_PER_TRACK = 5;
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: we could

mfarul (approved):

jakinluk (approved):

PR #1884, opened 2023-04-13
ADEN-13013 extend tracking data aggregation with gzip compression

jakinluk:
to verify run:
https://www.tvguide.com/?adengine_debug=1&adengine_version=local&icbm__dwTrafficLimits=%7B%22loadTimes%22%3A%7B%22agg%22%3Atrue,%22sample%22%3A100%7D%7D&icbm__dwTrafficCompression=true

where icbm__dwTrafficCompression=true enables compression
will work only if aggregation is enabled in the first place


src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -9,14 +20,14 @@ interface AggregateQueue {
 
 export class DwTrafficAggregator {
 	private readonly AGGREGATION_SECONDS_OFF = 5;
-	private readonly SEND_DATA_INTERVAL = 1000;
-	private readonly AGGREGATION_LIMIT_PER_TRACK = 5;
+	private readonly SEND_DATA_INTERVAL = 10000;
```
adi-wesolowski-fandom, line 23: pls, do not change this. 10 sec is very long. User can close page-tab sooner, and data will be lost. Ads load much quicker and we want to send all data, including requesting and success of receiving it. 

So IMO, here should be time which is half of the average time when ads being loaded to the browser, and it is around 1 second, as I measure it

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -1,6 +1,17 @@
...
 import { TrackingUrl } from '../../setup/tracking-urls';
 import { DataWarehouseParams } from '../data-warehouse';
+import { GzipCompressor } from './gzip-compressor';
+
+export type Compressed = {
+	contentEncoding?: string;
+	compressed: string | Uint8Array;
+};
+
+export interface DataCompressor {
```
adi-wesolowski-fandom, line 11: I am not sure, this definition should be here, in aggregator. Looks like very different responsibility. Aggregation can be done without compression, and IMO should be treated like a module, not functional part of aggregator (even we want to use it only here - but only here for now - later on compression can be implemented somewhere else).

I suggest to create new file for interface and type above, aside of aggregator.ts and gzip-compressor.ts

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -68,20 +79,40 @@ export class DwTrafficAggregator {
...
+		return (
+			context.get('services.dw-tracker.compression') && this.dataCompressor.compressionSupported()
+		);
+	}
+
 	private sendTrackData(trackingUrl: string, paramsAggregated: DataWarehouseParams[]): void {
 		const dwTrackingUrlV2 = `${trackingUrl}_v2`;
-		this.sendRequest(dwTrackingUrlV2, paramsAggregated);
+		if (this.isCompressionEnabled()) {
+			utils.logger('adengine_debug', 'DwTrafficAggregator', 'Compression enabled', {
```
adi-wesolowski-fandom, line 91: this key is to general. Make it more precise, like "aggregation-compression"

adi-wesolowski-fandom (changes requested):

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
jakinluk: I planned to remove those changes. I increased them only for test purposes. However, you mentioned subscribing to _visibilitychange_ would prevent data loss.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
jakinluk: Since DwTrafficAggregator is currently responsible for sending aggregated data (if these two things are separated we can talk about running compression outside of DwTrafficAggregator) compression must happen as well in the DwTrafficAggregator. Naturally, the compression module is injected as a dependency into DwTrafficAggregator. But since the compression module is only meant to fit DwTrafficAggregator needs, it's DwTrafficAggregator the owner of that interface. I followed here the Inversion of Control principle. And I would keep that interface here to underline the fact interface is owned by DwTrafficAggregator

src/platforms/shared/tracking/data-warehouse-utils/gzip-compressor.ts
```diff
@@ -0,0 +1,81 @@
...
+	readonly contentEncoding = 'gzip';
+
+	public compressionSupported(): boolean {
+		const compressionSupported =
+			!!(window as unknown as any).CompressionStream && !!window.TextEncoder;
+		if (!compressionSupported) {
+			utils.logger(
+				'adengine_debug',
+				'GzipCompressor',
+				'Compression not supported, returning uncompressed data instead.',
```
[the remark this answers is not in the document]
jakinluk: I proposed something in the middle

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: No, I disagree. As you mentioned, Aggregator class is reponsible for aggregation and sending data, not to compress. Is different task. And as I mentioned, this compression can be in future used in other classes, eg monitoring. 
So it is natural, this should be separated, and moved to separate file, where all of things related to "compression" should be placed. 
Aggregator can own this, but its not its responsibility, to compress, only to aggregate. Even sending data is in fact a little to much for this class to do. Compression is definitely too much. 
I understand you followed to Inversion of Control, but still, "compression" placed here break single-reponsibility. 

Look at it that way: there more than one reason to change this class: change algorithm of how we aggregate data, and change comporession algorithm. Just like Uncle Bob said "Gather together the things that change for the same reasons. Separate those things that change for different reasons."
(https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html)

Other example: when we would like to use compression in monitoring mechanism, we will be forced to implement aggregator, to get to this interface. And this would be so much confusing. 

"Inversion of Control" principle would be still here present, even if you moved this "compression" interfaces and responsibility out of this class. They will still be imported and used here.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
```diff
@@ -68,20 +79,40 @@ export class DwTrafficAggregator {
...
+				aggregatedItemsNo: paramsAggregated.length,
+			});
+			this.dataCompressor.compress(JSON.stringify(paramsAggregated)).then((result) => {
+				this.sendRequest(dwTrackingUrlV2 + result.contentEncoding || '', result.compressed);
```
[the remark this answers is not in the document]
adi-wesolowski-fandom: I think (maybe I am wrong), every call, including query-params would be placed separately. No need to create new endpoint IMO. 

Did you asked them, we could use query-param to determine we send compressed data or not?

src/platforms/shared/tracking/data-warehouse-utils/gzip-compressor.ts
```diff
@@ -0,0 +1,81 @@
...
+		return input.pipeThrough(new CompressionStream('gzip')) as ReadableStream<Uint8Array>;
+	}
+
+	private async readStream(input: ReadableStream<Uint8Array>): Promise<Uint8Array> {
+		const chunks: Uint8Array[] = [];
+		const reader = input.getReader();
+		// eslint-disable-next-line no-constant-condition
+		while (true) {
+			const { done, value } = await reader.read();
+			if (done) {
```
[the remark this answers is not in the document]
jakinluk: Before I start changing can you please first try your idea to see if it really looks better? 
If you want me to avoid disabling eslint rule I can propose something like that:
```
let finished;
while (!finished) {
	const { done, value } = await reader.read();
	if (!done) {
		chunks.push(value);
	}
	finished = done;
}
```
is your request something more than a matter of taste?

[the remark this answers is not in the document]
adi-wesolowski-fandom: it was an idea. Yes, this chunk.push part could be problematic and this does not look good either in this way. First version was better; 

I just wonder, if we could push empty value or null? To be something like this: 
`chunks.push(done ? value : null);`

```suggestion
let done, value;
while (!done) {
        { done, value } = await reader.read();
	chunks.push(done ? value : null);
}
```

If it cannot be done, then leave as it was at the first place

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
adi-wesolowski-fandom: Do as you want. I'd like just to keep description part in browser console as short as possible, to be easily to read, like "AdSlot" logs etc

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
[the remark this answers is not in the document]
adi-wesolowski-fandom: Ok, if you insist, leave it as is.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
jakinluk: I am not going to argue more about that and won't change my mind. Since as you said this class should not in the first place perform sending and should be only responsible for aggregation extracting interface outside does not change anything about the responsibility of DwTrafficAggregator, it would be just an irrelevant move.

src/platforms/shared/tracking/data-warehouse-utils/gzip-compressor.ts
```diff
@@ -0,0 +1,81 @@
+import { utils } from '@wikia/ad-engine';
+import { Compressed, DataCompressor } from './dw-traffic-aggregator';
+
+export class GzipCompressor implements DataCompressor {
+	readonly contentEncoding = 'gzip';
+
+	public compressionSupported(): boolean {
+		const compressionSupported =
+			!!(window as unknown as any).CompressionStream && !!window.TextEncoder;
```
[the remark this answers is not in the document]
jakinluk: I don't like putting any logic into a constructor since then I would need to add try-catch there as well. But I will make it to be calculated only once.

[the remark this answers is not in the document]
jakinluk: Your suggestions are valid and would prevent any unexpected errors but since every logic might be changed in the future and become more complicated I would prefer to keep it outside the constructor.

src/platforms/shared/tracking/data-warehouse-utils/dw-traffic-aggregator.ts
jakinluk: Let's move the discussion outside and come back here with conclusions

andrzejkupczyk (approved):

src/core/utils/get-global.ts
```diff
@@ -0,0 +1,24 @@
...
+	if (typeof globalThis !== 'undefined') {
+		return globalThis;
+	}
+	if (typeof self !== 'undefined') {
+		return self;
+	}
+	if (typeof global !== 'undefined') {
+		return global;
+	}
+	return undefined;
```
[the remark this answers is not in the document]
jakinluk: good point. I will do that

src/platforms/shared/tracking/data-warehouse-utils/dw-aggregated-data-sender.ts
```diff
@@ -0,0 +1,60 @@
+import { context, utils } from '@wikia/ad-engine';
+import { DwAggregatedDataGzipCompressor } from './dw-aggregated-data-gzip-compressor';
+
+export type Compressed = {
```
adi-wesolowski-fandom, line 4: 👍

adi-wesolowski-fandom (approved):

andrzejkupczyk (approved):

PR #1906, opened 2023-04-25
Lower coverage levels so test-coverage command does not fail

jbienkowski-fandom:


dmnsobczak (approved):
@jbienkowski-fandom What caused such drastic decrease in unit tests coverage? ;_;

jbienkowski-fandom:
It is a result of few things... First, the coverage went a little below 55% few days ago (which caused Codecov to fail when uploading the coverage reports as `npm run test-coverage` was exiting with code 1), so all the codecov CI checks were failing. Then, my bundle size optimizations caused a drop of ~9% coverage (god only knows why)... I will try and take a look why it happened and maybe I will be able to fix it.

szymon-nawrocki (approved):

PR #1907, opened 2023-04-25
ADEN-13095 - upgrade logic for 0 values and set default 0, in case of…

adi-wesolowski-fandom:
… absence of ICBM variable

### DIFF
--- a/src/core/utils/outbound-traffic-restrict.ts
+++ b/src/core/utils/outbound-traffic-restrict.ts
@@ -22,6 +22,10 @@ class OutboundTrafficRestrict {
 			isAllowed = this.getSeed() < threshold;
 		}
 
+		if (threshold < 1) {
+			isAllowed = false;
+		}
+
 		utils.logger(
 			'outbound-traffic-restrict',
 			`Outbound traffic for: "${serviceName}" is allowed: ${isAllowed}`,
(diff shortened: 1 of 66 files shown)


mfarul (approved):

mfarul (dismissed):

andrzejkupczyk (approved):

jbienkowski-fandom (approved):
LGTM, but I wonder why `fetchTimeout` became a class?

src/core/utils/outbound-traffic-restrict.ts
```diff
@@ -22,6 +22,10 @@ class OutboundTrafficRestrict {
 			isAllowed = this.getSeed() < threshold;
 		}
 
+		if (threshold < 1) {
```
jakinluk, line 25: why do we need that ? if threshold is 0 , `isAllowed = this.getSeed() < threshold` should return false

src/platforms/shared/setup/metric-reporter.setup.ts
```diff
@@ -11,7 +11,7 @@ export class MetricReporterSetup implements DiProcess {
 		context.set('services.monitoring.service', 'adeng');
 		context.set(
 			'services.monitoring.threshold',
-			this.instantConfig.get('icMonitorTrafficThreshold'),
+			this.instantConfig.get('icMonitorTrafficThreshold', 0),
```
jakinluk, line 14: so if icMonitorTrafficThreshold does not exist on the instantConfig all outbound traffic is disabled by default?

PR #1913, opened 2023-04-28
ADEN-13062 Adding interstitial to all N&R platforms

dmnsobczak:


src/core/services/slot-service.ts
```diff
@@ -254,7 +254,7 @@ class SlotService {
 	 */
 	private getGptAdSlot(id: string): AdSlot {
 		const prefix = context.get('custom.serverPrefix');
```
rafalgiemza, line 256: Maybe put empty string as a fallback here? 

```suggestion
		const prefix = context.get('custom.serverPrefix') ?? '';
```

src/platforms/news-and-ratings/metacritic/setup/context/slots/metacritic-slots-context.setup.ts
```diff
@@ -5,6 +5,19 @@ import { Injectable } from '@wikia/dependency-injection';
 export class MetacriticSlotsContextSetup implements DiProcess {
 	execute(): void {
 		const slots = {
+			interstitial: {
```
rafalgiemza, line 8: This part repeats for all N&R platform. Maybe we can put it in shared folder?

rafalgiemza (approved):
Looks good to me, I have only 2 suggestions

src/core/services/slot-service.ts
dmnsobczak: No, because we want to skip the dot `.` too when server prefix is not set.

kwfandom (approved):

PR #1926, opened 2023-05-10
ADEN-12562 - Migrate Identity, Targeting and Tracking related partners to Identity-Engine

domjanik:
In this PR I've moved TaCo partners from AdEng to IdentityEngine. Those partners are :
- Audigent
- LiveConnect
- LiveRamp ( only its pixel ) 
- Pubmatic Identity Hub


jbienkowski-fandom:
We can remove even more code, for example setting some `context`:
https://github.com/Wikia/ad-engine/blob/170b154d090e1a45c5a5f99ab6e374f77744f17c/src/platforms/shared/setup/base-context.setup.ts#L175-L181

szymon-nawrocki (approved):

PR #1928, opened 2023-05-11
ADEN-12110 - Removed COPPA-related flags

domjanik:
In this PR I've removed Feature Flags, that were responsible for turning on COPPA integration in bidders. Those flags were :
- icCoppaGam
- icCoppaA9
- icCoppaPrebid
After this PR is merged, it should not be possible to turn off that integration by using ICBM


jbienkowski-fandom (approved):
LGTM

PR #1930, opened 2023-05-11
ADEN-13087 - Fixed problem with AgeGate variable in non-kid wikis

domjanik:
## Ticket

https://fandom.atlassian.net/browse/ADEN-13087

## Description

These changes fix behaviour of AgeGate result. At the moment, if the user provided his age declaration it was being used also outside of directed at children wiki. Now, the AgeGate result will only be taken into account if wiki is directed at children.
This behaviour is now being monitored by unit tests.

### DIFF
--- a/src/core/utils/is-coppa-subject.ts
+++ b/src/core/utils/is-coppa-subject.ts
@@ -2,19 +2,19 @@ import { utils } from '../index';
 import { context, CookieStorageAdapter } from '../services';
 
 export function isCoppaSubject(): boolean {
-	const directedAtChildren = context.get('wiki.targeting.directedAtChildren');
-	if (context.get('services.ageGateHandling') && directedAtChildren) {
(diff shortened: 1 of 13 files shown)


jbienkowski-fandom (dismissed):

jbienkowski-fandom (approved):

src/core/utils/is-coppa-subject.ts
```diff
@@ -2,20 +2,19 @@ import { utils } from '../index';
 import { context, CookieStorageAdapter } from '../services';
 
 export function isCoppaSubject(): boolean {
-	if (context.get('services.ageGateHandling')) {
+	const directedAtChildren = context.get('wiki.targeting.directedAtChildren');
```
szymon-nawrocki, line 5: I think changing variable name to `wikiDirectedAtChildren` would improve readability of the code

szymon-nawrocki (approved):

PR #1935, opened 2023-05-16
ADEN-13124 Add 'ag' cookie dependency to isCoppaSubject()

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-13124

### Description
It was decided to change AgeGate cookie name from "age_gate" to "ag".
In this PR I've updated the cookie name so it is backward compatibile.


domjanik (approved):

src/core/utils/is-coppa-subject.ts
```diff
@@ -6,10 +6,9 @@ export function isCoppaSubject(): boolean {
 	if (context.get('services.ageGateHandling') && wikiDirectedAtChildren) {
 		try {
 			const cookieStorage = new CookieStorageAdapter();
-			const ageGateResult = cookieStorage.getItem('age_gate');
+			const ageGateResult = cookieStorage.getItem('ag');
 			if (ageGateResult) {
```
jbienkowski-fandom, line 10: (nit) I don't think that we need this `if (ageGateResult)` anymore, we can just simply return without it

jbienkowski-fandom (approved):
LGTM

PR #1936, opened 2023-05-17
UXFCP-3744 (chore): Remove slot for desktop.

wera-trollcode:
With UCP pull request we will be removing logic for pulling sponsored tile to fan-feed (recirculation footer) on desktop.

Ticket: https://fandom.atlassian.net/browse/UXFCP-3744
Origin: https://github.com/Wikia/unified-platform/pull/14026
Ad-tests: (to be added)

### DIFF
--- a/src/platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
+++ b/src/platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts
@@ -143,11 +143,6 @@ export class UcpDesktopSlotsContextSetup implements DiProcess {
 				trackEachStatus: true,
 				trackingKey: 'featured-video',
 			},
-			ntv_feed_ad: {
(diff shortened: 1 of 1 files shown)


dmnsobczak (changes requested):
This is second out of two Nativo slots that's removed from UCP Desktop, so that's mean we can do a full Nativo cleanup on that platform 🚀

swietlana:
@dmnsobczak Is there a possibility to do such "full" cleanup out of this PR? I believe our team members won't feel comfortable cleaning it up since we don't have enough insights of this repo. WDYT?

houssem-yahiaoui:
@dmnsobczak We want this to be pushed through for the sake of our deployment despite the need for nativo removal :) it's a high priority for us and I don't think we will have the capacity to do the other clean up, can you please help us with getting this approved, else how this can harm our release cc @swietlana

PR #1942, opened 2023-05-19
ADEN-13020 - PoC - reduce number of request to prebid, for unnecessar…

adi-wesolowski-fandom:
…y slots (as there is no placements for slots on current loaded page)


src/platforms/news-and-ratings/comicvine/setup/context/prebid/comicvine-prebid-config.setup.ts
```diff
@@ -33,6 +33,7 @@ export class ComicvinePrebidConfigSetup implements DiProcess {
 		context.set('bidders.prebid.rubicon_display', getRubiconDisplayContext(isDesktop));
 		context.set('bidders.prebid.wikia', getWikiaContext());
 		context.set('bidders.prebid.yahoossp', getYahooSspContext(isDesktop));
+		context.set('bidders.prebid.filter', 'static');
```
preludek, line 36: That flag feels odd and out of place. We are loading all partners' configs/ contexts and all of a sudden a `static` flag on `filter` which I'm guessing it's filtering but what and why? 🤔

src/ad-bidders/prebid/adapters-registry.ts
```diff
@@ -99,6 +99,10 @@ class AdaptersRegistry {
 			}
 		});
 
+		if (context.get('bidders.prebid.filter') === 'static') {
+			adUnits = adUnits.filter((a) => document.querySelector(`div[id="${a.code}"]`) !== null);
```
preludek, line 103: Can't this be filtered out above on lines 95-96? It seems like we push those adUnits there only to filter them here and adUnit.code shouldn't have a falsy value anyway.

preludek (commented):
Left some questions

src/ad-bidders/prebid/adapters-registry.ts
adi-wesolowski-fandom: Yes, it shouldn't and I guess, it wont be. As long as there will be no wrong slot setup. 

I wanted to put them as late as possibly, to filter them at the end, but just before we send bids. (I kept this thinking since I worked with TvGuide SPa pages and I kept them later till the end)

src/platforms/news-and-ratings/comicvine/setup/context/prebid/comicvine-prebid-config.setup.ts
adi-wesolowski-fandom: why do you fell its odd and out of place? I think it is the best place to put it here, because we setup here all prebid things and custom it, overriding default config. 

Why we filter - I described everythink in task Jira comment: https://fandom.atlassian.net/browse/ADEN-13020?focusedCommentId=706444

preludek (dismissed):

preludek (approved):

PR #1962, opened 2023-05-30
ADEN-13247 - Moved directedAtChildren out of Blockthrough scope

domjanik:
In this PR I've moved directedAtChildren flag out of Blockthrough scope. I did that, because we found out that it is being used by other partners as well and its better to keep such values out of single partners scope.


spec/ad-services/bt-rec/bt-rec.spec.ts
```diff
@@ -58,15 +58,15 @@ describe('BlockThrough recovery', () => {
 		loadScriptStub.returns(Promise.resolve());
 		context.set('wiki.targeting.directedAtChildren', true);
 		await btRec.run().then(() => {
-			expect(window.fandomContext.partners.blockthrough.directedAtChildren).to.equal(true);
```
jbienkowski-fandom, line 61: are we not interested in backwards compatibility?

src/ad-services/bt-rec/index.ts
```diff
@@ -20,9 +20,7 @@ class BTRec {
 
 		utils.logger(logGroup, 'loading');
 		globalContextService.setValue(GlobalContextCategories.partners, {
```
szymon-nawrocki, line 22: This code sets up `partners.directedAtChildren` at `fandomContext` object and now is not only BTRec related.
I think it should be moved out of `/ad-services/bt-rec` catalogue

spec/ad-services/bt-rec/bt-rec.spec.ts
domjanik: We're going to inform our partners about that and sync it with them. I think we do not need such compatibility right now 🤔

src/core/utils/is-coppa-subject.ts
```diff
@@ -7,12 +7,25 @@ export function isCoppaSubject(): boolean {
 		try {
 			const cookieStorage = new CookieStorageAdapter();
 			const ageGateResult = cookieStorage.getItem('ag');
+			const isCoppaSubject = ageGateResult === '1' ? false : wikiDirectedAtChildren;
 
-			return ageGateResult === '1' ? false : wikiDirectedAtChildren;
+			globalContextService.setValue(GlobalContextCategories.partners, {
+				directedAtChildren: !!isCoppaSubject,
+				blockthrough: {
```
jbienkowski-fandom, line 14: backward compatible after all? :) let `@types/fandom/index.d.ts` reflect that change too please

src/core/utils/is-coppa-subject.ts
```diff
@@ -1,4 +1,4 @@
-import { utils } from '../index';
+import { GlobalContextCategories, globalContextService, utils } from '../index';
 import { context, CookieStorageAdapter } from '../services';
 
 export function isCoppaSubject(): boolean {
```
jbienkowski-fandom, line 4: we should check `globalContextService.getValue` and use it as "cache" as `utils.isCoppaSubject` is used quite a lot

jbienkowski-fandom (changes requested):

src/core/utils/is-coppa-subject.ts
domjanik: Yeah... I thought about this and after all I'm kinda afraid of letting this only up to our partners xD

src/core/utils/is-coppa-subject.ts
szymon-nawrocki, line 4: Also this logic is duplicated in IdentityEngine, maybe this is a good moment to unify it between the repos?
https://github.com/Wikia/identity-engine/blob/main/src/modules/shared/CoppaSubject/index.ts

szymon-nawrocki (approved):

jbienkowski-fandom (dismissed):

szymon-nawrocki (dismissed):

src/core/services/global-context-service.ts
```diff
@@ -62,6 +65,9 @@ export class GlobalContextService {
 		category: K,
 		key?: I,
 	): WindowFandomContext[K][I] | undefined {
+		if (!window.fandomContext) {
+			return;
```
jbienkowski-fandom, line 69:
```suggestion
			return undefined;
```

src/core/services/global-context-service.ts
```diff
@@ -62,6 +65,9 @@ export class GlobalContextService {
...
+		}
 		const categoryObj = window.fandomContext[category];
 		if (categoryObj) {
 			return categoryObj[key];
```
jbienkowski-fandom, line 73: (nit1) the `else` statement is unnecessary, so let's get rid of it
(nit2) `return undefined;` at the end (explicit is better than implicit)

szymon-nawrocki (approved):

jbienkowski-fandom (approved):
LGTM

PR #1973, opened 2023-06-06
ADEN-13135-fix-floor

lnawrotek:


src/platforms/f2/setup/dynamic-slots/f2-slots-definition-repository.ts
```diff
@@ -53,10 +53,6 @@ export class F2SlotsDefinitionRepository {
 	}
 
 	getFloorAdhesionConfig(): SlotSetupDefinition {
-		if (!context.get('state.isMobile')) {
```
dmnsobczak, line 56: One case to check: FloorAdhesion should be enabled only on mobile for standard ads. Make sure that it won't appear on desktop (use cid to force it). Probably you need to ad `disabled: true` to slot context on desktop when UAP is not loaded.

src/platforms/f2/styles.scss
```diff
@@ -219,24 +228,26 @@ body:not(.uap-sticked) {
...
-		width: 40px;
-		height: 40px;
-		top: -20px;
-
-		&:after {
-			content: '';
-			position: absolute;
-			width: 45px;
-			height: 45px;
+html.mobile {
```
dmnsobczak, line 231: `html.mobile #floor_adhesion {}` - less indentation :D

dmnsobczak (commented):
LGTM 👍 Merged to main feature branch.

PR #1984, opened 2023-06-12
Aden 13827

mfarul:


src/ad-tracking/compilers/slot-bidders-tracking-compiler.ts
```diff
@@ -1,7 +1,8 @@
 import { Dictionary } from '@ad-engine/core';
+import { Bidders } from '../../ad-bidders';
```
dmnsobczak, line 2:
```suggestion
import { Dictionary } from '@ad-engine/core';
```

src/ad-tracking/compilers/slot-bidders-tracking-compiler.ts
```diff
@@ -1,7 +1,8 @@
...
 import { CompilerPartial } from '../base-tracker';
 
-async function getBiddersPrices(slotName: string, bidders): Promise<Dictionary<string>> {
+async function getBiddersPrices(slotName: string, bidders: Bidders): Promise<Dictionary<string>> {
```
dmnsobczak, line 5:
```suggestion
async function getBiddersPrices(slotName: string, bidders): Promise<Dictionary<string>> {
```

src/ad-tracking/compilers/slot-bidders-tracking-compiler.ts
dmnsobczak: We should avoid cross importing from different subbundles:
https://github.com/Wikia/ad-engine/blob/3504862077155a372a2ca3e8410e6e343a1a5246/src/ad-tracking/.eslintrc.json#L16

dmnsobczak (approved):

PR #1986, opened 2023-06-13
ADEN-13243 change before_page_change event to page changed

preludek:


preludek:
I made before page change event work on neutron so those changes are redundant

PR #1989, opened 2023-06-14
ADEN-13294 Handle icMonitorTrafficThreshold being JSON object

jbienkowski-fandom:
### Links
https://fandom.atlassian.net/browse/ADEN-13294

### Description
[icMonitorTrafficThreshold](https://services.fandom.com/icbm/dashboard/feature_flag/icMonitorTrafficThreshold) is a great variable, but for now it had a very simple format (integer). While implementing some of the reporting logic in IdentityEngine, I wanted to use the same ICBM variable to control sampling, but I would like to manage percentages and geos in Identity monitoring independent of AdEngine monitoring settings. That's why the ICBM variable will be moving from integer to JSON object with more complex format.
This PR provides a backward compatible implementation to AdEngine, so it will be able to handle icMonitorTrafficThreshold being both integer and JSON object.

### DIFF
--- a/spec/platforms/shared/setup/metric-reporter.setup.spec.ts
+++ b/spec/platforms/shared/setup/metric-reporter.setup.spec.ts
@@ -3,9 +3,9 @@ import { MetricReporterSetup } from '@wikia/platforms/shared';
 import { expect } from 'chai';
 
 describe('MetricReporterSetup', () => {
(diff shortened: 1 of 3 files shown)


domjanik:
Nice, great idea how to handle this! :)

adi-wesolowski-fandom (approved):

PR #1991, opened 2023-06-14
ADEN-13283 - Modified IE to handle icIdentityPartners flag

domjanik:
In this PR I've added a new Feature Flag: icIdentityPartners. This flag will be responsible for toggling partners in AdEngine and Identity. Flag set as true, should turn off Audigent, LiveIntent, LiveRamp, Eyeota and Pubmatic Identity Hub in AdEngine, and turn them on in IdentityEngine.

### DIFF
--- a/spec/ad-services/live-connect/live-connect.spec.ts
+++ b/spec/ad-services/live-connect/live-connect.spec.ts
@@ -1,6 +1,6 @@
 import { LiveConnect } from '@wikia/ad-services';
 import { communicationService, eventsRepository } from '@wikia/communication';
-import { context, utils } from '@wikia/core';
+import { context, InstantConfigService, utils } from '@wikia/core';
 import { expect } from 'chai';
 import { SinonSpy } from 'sinon';
 
@@ -11,14 +11,16 @@ const mockedStorageStrategyVariable = {
 };
 
 describe('LiveConnect', () => {
-	const liveConnect = new LiveConnect();
-	let loadScriptStub;
+	let liveConnect;
+	let loadScriptStub, instantConfigStub;
 
 	before(() => {
 		context.set('services.liveConnect.cachingStrategy', mockedStorageStrategyVariable);
 	});
 
 	beforeEach(() => {
+		instantConfigStub = global.sandbox.createStubInstance(InstantConfigService);
+		instantConfigStub.get.withArgs('icIdentityPartners').returns(false);
 		loadScriptStub = global.sandbox
 			.stub(utils.scriptLoader, 'loadScript')
 			.returns(Promise.resolve({} as any));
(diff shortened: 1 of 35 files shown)


src/ad-services/identity-setup/index.ts
```diff
@@ -18,6 +18,9 @@ export class IdentitySetup extends BaseServiceSetup {
 			if (ppid) {
 				targetingService.set('ppid', ppid);
 			}
+			const segments = globalContextService.getValue('targeting', 'AU_SEG');
```
jbienkowski-fandom, line 21: I suppose it should run only when `icIdentityPartners` is set to false (you could use BaseContextSetup and use `context` object to not load InstantConfigService here)

spec/ad-services/live-connect/live-connect.spec.ts
```diff
@@ -11,21 +11,24 @@ const mockedStorageStrategyVariable = {
...
+		instantConfigStub = global.sandbox.createStubInstance(InstantConfigService);
+		instantConfigStub.get.withArgs('icIdentityPartners').returns(false);
 		loadScriptStub = global.sandbox
 			.stub(utils.scriptLoader, 'loadScript')
 			.returns(Promise.resolve({} as any));
 		context.set('services.liveConnect.enabled', true);
 		context.set('options.trackingOptIn', true);
 		context.set('options.optOutSale', false);
 		context.set('wiki.targeting.directedAtChildren', false);
+		liveConnect = new LiveConnect(instantConfigStub);
```
szymon-nawrocki, line 31: LiveConnect does not have constructor and instantConfig as a parameter

src/ad-services/identity-setup/index.ts
```diff
@@ -18,6 +18,11 @@ export class IdentitySetup extends BaseServiceSetup {
 			if (ppid) {
 				targetingService.set('ppid', ppid);
 			}
+			if (this.instantConfig.get('icIdentityPartners', false)) {
```
szymon-nawrocki, line 21: Why this code is added? It looks like the doubled logic from: https://github.com/Wikia/ad-engine/blob/9a529922ea889057bdac437cee9f64cea2c7e906/src/ad-services/audigent/index.ts#L117

jbienkowski-fandom: Because this logic in now being doubled in AdEngine and IdentityEngine. If `icIdentityPartners` is true, then Audigent segments will be loaded by Identity and AdEngine must obtain it somehow. If `icIdentityPartners` is false, then this part will not work, but whole Audigent class will be executed.

jbienkowski-fandom (approved):

szymon-nawrocki (approved):

jbienkowski-fandom (approved):

PR #1992, opened 2023-06-14
ADEN-12211 Wunderkind Integration

lnawrotek:


src/ad-services/wunderkind/index.ts
```diff
@@ -0,0 +1,33 @@
+import { communicationService, eventsRepository, UapLoadStatus } from '@ad-engine/communication';
+import { BaseServiceSetup, context, utils } from '@ad-engine/core';
+
+const logGroup = 'wunderkind';
+
+export class Wunderkind extends BaseServiceSetup {
+	call(): void {
+		if (!this.instantConfig.get('icWunderkind')) {
```
adi-wesolowski-fandom, line 8: I think it would be slightly more robust if we set here default return value. 

_(I faced problem few weeks ago, that we could get false calls without it (0.01% of traffic, but it still create some thousands of activations).)_

src/ad-services/wunderkind/index.ts
```diff
@@ -0,0 +1,33 @@
...
+			utils.logger(logGroup, 'disabled');
+			return;
+		} else if (context.get('state.isLogged')) {
```
adi-wesolowski-fandom, line 11: Maybe this should go first? We dont want to show any ads to logged user, so we can exist much quicker and dont need to check if user is signed in. And, I guess there is no need to "else". These are separate conditions.

src/ad-services/wunderkind/index.ts
```diff
@@ -0,0 +1,33 @@
...
+			utils.logger(logGroup, 'disabled - user is logged');
+			return;
+		}
+
+		communicationService.on(eventsRepository.AD_ENGINE_UAP_LOAD_STATUS, (action: UapLoadStatus) => {
+			if (!action.isLoaded) {
+				this.loadScript();
+			} else {
+				utils.logger(logGroup, 'disabled - UAP is loaded');
+			}
```
adi-wesolowski-fandom, line 21: nitpick: quick return first instead of login inside if condition block? just for readability. 

```suggestion
			if (action.isLoaded) {
			    utils.logger(logGroup, 'disabled - UAP is loaded');
			    return;
			}
			this.loadScript();
```

src/ad-services/wunderkind/index.ts
```diff
@@ -0,0 +1,33 @@
...
+		});
+	}
+
+	private loadScript(): void {
+		const libraryUrl = `//tag.wknd.ai/5047/i.js`;
+		utils.logger(logGroup, 'loading', libraryUrl);
+
+		utils.scriptLoader.loadScript(libraryUrl).then(() => {
+			utils.logger(logGroup, 'ready');
```
adi-wesolowski-fandom, line 30: question: can we catch here error situation and manage it somehow? (rejest/onerror). Maybe different message, instead of always "ready"?

src/ad-services/wunderkind/index.ts
lnawrotek: In my opinion, we don't need a default value here.

src/ad-services/wunderkind/index.ts
lnawrotek: I'm not sure we need that, other ad-services work in a similar way. Logger is only for debugging, they are disabled by default.

adi-wesolowski-fandom: ok

adi-wesolowski-fandom (dismissed):

kwfandom (approved):

PR #2000, opened 2023-06-19
ADEN-13257 LiveRamp ATS 2.0

jbienkowski-fandom:
### Links
[Jira ticket](https://fandom.atlassian.net/browse/ADEN-13257)
[doc: Install ATS](https://docs.liveramp.com/privacy-manager/en/install-ats-and-test-your-implementation.html#install-ats)
[doc: Obtain IDs](https://docs.liveramp.com/privacy-manager/en/configure-how-identifiers-are-obtained.html#how-to-configure-direct-mode)

### Description
Changes to enable standalone LiveRamp ATS 2.0 integration.

1. `PlatformContextSetup` was missing from `UcpNoAdsPlatform`, as the result `context.get('wiki.opts. userEmailHashes')` returned undefined and script was not loaded.
2. Script URL has changed
3. Because we are not using CMP from LiveRamp it is necessary to provide `consentType` and `consentString`.


domjanik (dismissed):

src/ad-bidders/prebid/ats/ats.ts
```diff
@@ -5,31 +5,39 @@ const logGroup = 'ATS';
...
-
-			if (!Array.isArray(userEmailHashes) || userEmailHashes?.length !== 3) {
-				return Promise.resolve();
-			}
+		if (this.isLoaded) {
+			return;
+		}
 
-			const [md5, sha1, sha256] = userEmailHashes;
+		const userEmailHashes: [string, string, string] = context.get('wiki.opts.userEmailHashes');
```
szymon-nawrocki, line 25: I don't see this context variable: `wiki.opts.userEmailHashes` being set anywhere, it looks like it's deprecated

jbienkowski-fandom: Check PlatformContextSetup class, it is being set there (but not directly)
https://github.com/Wikia/ad-engine/blob/cf42b8df2c27b18bcd56f1816f61f86660544ef6/src/platforms/shared/setup/platform-context.setup.ts#L3

src/ad-bidders/prebid/ats/ats.ts
```diff
@@ -1,38 +1,57 @@
...
+		const launchpadBundleScript = utils.scriptLoader.loadScript(this.launchpadBundleScriptUrl);
+
+		return Promise.all([launchpadScript, launchpadBundleScript])
+			.then(() => this.waitForAts())
+			.then(() => {
+				const consentType = context.get('options.geoRequiresConsent') ? 'gdpr' : 'ccpa';
+				const consentString =
+					consentType === 'gdpr' ? Cookies.get('euconsent-v2') : Cookies.get('usprivacy');
+
+				(window as any).ats.setAdditionalData({
```
szymon-nawrocki, line 41: I'd use the benefits of having Typescript in the project and add `ats` to `@types`

jbienkowski-fandom: Great suggestion! Thanks :)

src/ad-bidders/prebid/ats/ats.ts
```diff
@@ -1,38 +1,57 @@
...
 
+				utils.logger(logGroup, 'additional data set');
+
 				this.isLoaded = true;
 			});
-		}
+	}
+
+	private waitForAts(): Promise<boolean> {
+		return new utils.WaitFor(() => (window as any).ats !== undefined, 10, 50).until();
```
szymon-nawrocki, line 55: nitpick: now since the `ats` is added to window types, the `(window as any)` can be changed for just `window` ;)

szymon-nawrocki, line 55:
```suggestion
		return new utils.WaitFor(() => window.ats !== undefined, 10, 50).until();
```

szymon-nawrocki (approved):

domjanik (approved):

PR #2018, opened 2023-06-26
ADEN-13176 - Stickiness on N&R

adi-wesolowski-fandom:
ADEN-13176 - stickiness on N&R:

- apply sticky template to all n&r sites, enabling this template as default, including SPA
- it is not active until lineitem show up in icbm variable
- default behaviour can be still overwritten by "force" icbm variable as on fandom site

Jira task: https://fandom.atlassian.net/browse/ADEN-13176

source branch: ADEN-13135 (!)


dmnsobczak (approved):

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -45,19 +45,27 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
...
+	}
 
-					return (
-						lineId.toString() === this.adSlot.lineItemId.toString() &&
-						(!geo || utils.geoService.isProperGeo([geo]))
-					);
-				});
+	private checkIdsMapBySlotGamId(gamId: string, map: string[]): boolean {
+		if (!Array.isArray(map)) {
+			return true;
```
dmnsobczak, line 59: I suggest `false` here. This check here is to prevent AdEngine from crashing when somebody will break icVariable. In that case it's better to disable sticking rather than enable it globally 🤔 
On the other hand, ICBM have syntax checking now, so this might be redundant now.

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -45,19 +45,27 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
...
 		}
-		return false;
+
+		return map
+			.map((element) => element.toString())
+			.some((element) => {
+				const [mapGamId, geo] = element.split(':', 2);
```
dmnsobczak, line 65: Nobody is using this "per geo feature" :( But whatever.

dmnsobczak (dismissed):
@adi-wesolowski-fandom One suggestion: `true` => `false`

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
adi-wesolowski-fandom: I wasnt sure. It is here, so I thought,maybe somewhere it is. And removing it, change logic, and need to tell business, we simplified it. I can try to simplify it if you insist :)

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
adi-wesolowski-fandom: ok

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
dmnsobczak: Noo, it's fine :)

adi-wesolowski-fandom: done here: https://github.com/Wikia/ad-engine/pull/2029

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -56,7 +56,7 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
 
 	private checkIdsMapBySlotGamId(gamId: string, map: string[]): boolean {
 		if (!Array.isArray(map)) {
-			return true;
```
preludek: This template is used between all platforms, is that change necessary and safe? I see potential regression.

Also if it's safe we can change it to a ternary in return `return Array.isArray(map) ? map : false` something like that

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -56,7 +56,7 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
...
+			return false;
```
preludek, line 59: This template is used between all platforms, is that change necessary and safe? I see potential regression.

Also if it's safe we can change it to a ternary in return `return Array.isArray(map) ? map : false` something like that

src/platforms/news-and-ratings/shared/styles/uap-styles.scss
```diff
@@ -17,7 +17,7 @@
 
 			@include bfaa-slot();
 			@include uap-porvata();
-			@include sticky-tlb-slot();
+			@include sticky-tlb-slot(0px, 5000061);
```
preludek, line 20: I kinda dislike that. Wa are using a random high z-index here and that can bite us in the ass in the future. If it's necessary to change it please move it to some sort of a variable:
1. to keep all of those in one place
2. to have it named and to know right away what it's for

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
preludek: Ok, I got the answer for first question here - https://github.com/Wikia/ad-engine/pull/2018/files/990fe7512e1ad18093bd6ff47053a7ee188f391c#r1250580706 :)

adi-wesolowski-fandom: ok

src/platforms/news-and-ratings/shared/styles/uap-styles.scss
adi-wesolowski-fandom: its the highest across N&R I found. I put it here as it has visible impact. But moving it to ICBM is a good option, but - wouldnt be too much to put such a simple thing there? MAybe you are right. We can do that.

adi-wesolowski-fandom: moved it to configuration per each site separately. should be more clear now

preludek (approved):

kwfandom (approved):

PR #2020, opened 2023-06-27
ADEN-13340 Call LiveConnect resolve method once instead multiple times

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-13340

### Description
It was recommended to us by LiveConnect to call their `window.LiQ.resolve()` method once instead of multiple times.
It can be done by calling this method with array of parameters we want to resolve, so: `["sha1", "sha2", "md5", "unifiedId"]`

### DIFF
--- a/src/ad-services/live-connect/index.ts
+++ b/src/ad-services/live-connect/index.ts
@@ -4,7 +4,13 @@ import { BaseServiceSetup, context, localCache, UniversalStorage, utils } from '
 interface IdConfig {
 	id: string;
 	name: string;
-	params?: LiQParams;
+}
+
+interface LiQResolveResponse {
+	sha1?: string;
+	sha2?: string;
+	md5?: string;
+	unifiedId?: string;
 }
 
 interface CachingStrategyConfig {
@@ -19,9 +25,9 @@ const liveConnectScriptUrl = 'https://b-code.liadm.com/a-07ev.min.js';
 
 const idConfigMapping: IdConfig[] = [
 	{ id: 'unifiedId', name: `${partnerName}-unifiedId` },
-	{ id: 'sha2', name: partnerName, params: { qf: '0.3', resolve: 'sha2' } },
-	{ id: 'md5', name: `${partnerName}-md5`, params: { qf: '0.3', resolve: 'md5' } },
-	{ id: 'sha1', name: `${partnerName}-sha1`, params: { qf: '0.3', resolve: 'sha1' } },
+	{ id: 'sha2', name: partnerName },
+	{ id: 'md5', name: `${partnerName}-md5` },
+	{ id: 'sha1', name: `${partnerName}-sha1` },
 ];
 
 export class LiveConnect extends BaseServiceSetup {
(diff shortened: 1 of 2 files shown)


src/ad-services/live-connect/index.ts
```diff
@@ -47,57 +53,64 @@ export class LiveConnect extends BaseServiceSetup {
...
+		}
+
+		window.liQ.resolve(
+			(result) => {
+				this.trackIds(result);
+			},
+			(err) => {
+				console.error(err);
+			},
+			{ qf: '0.3', resolve: ['sha1', 'sha2', 'md5', 'unifiedId'] },
```
jbienkowski-fandom, line 77:
```suggestion
			{ qf: '0.3', resolve: idConfigMapping.map(config => config.id) },
```

src/ad-services/live-connect/index.ts
```diff
@@ -47,57 +53,64 @@ export class LiveConnect extends BaseServiceSetup {
...
-			utils.logger(logGroup, `id ${idName}: ${partnerIdentityId}`);
+		Object.keys(liQResponse).forEach((key) => {
+			const trackingKeyName = this.getTrackingKeyName(key);
+
+			if (!this.isAvailableInStorage(trackingKeyName)) {
+				return;
+			}
 
-			if (idName === 'unifiedId') {
+			if (liQResponse.unifiedId) {
```
jbienkowski-fandom, line 91: this change will cause that on each `forEach` iteration `LIVE_CONNECT_RESPONDED_UUID` event will be broadcasted
```suggestion
			if (key  === 'unifiedId') {
```

src/ad-services/live-connect/index.ts
```diff
@@ -47,57 +53,64 @@ export class LiveConnect extends BaseServiceSetup {
...
+	}
+
+	trackIds(liQResponse: LiQResolveResponse): void {
+		utils.logger(logGroup, 'resolve response:', liQResponse);
 
-			utils.logger(logGroup, `id ${idName}: ${partnerIdentityId}`);
+		Object.keys(liQResponse).forEach((key) => {
+			const trackingKeyName = this.getTrackingKeyName(key);
+
+			if (!this.isAvailableInStorage(trackingKeyName)) {
```
jbienkowski-fandom, line 87: it makes more sense this way
```suggestion
			if (this.isAvailableInStorage(trackingKeyName)) {
```

src/ad-services/live-connect/index.ts
```diff
@@ -4,7 +4,13 @@ import { BaseServiceSetup, context, localCache, UniversalStorage, utils } from '
 interface IdConfig {
 	id: string;
 	name: string;
-	params?: LiQParams;
+}
+
+interface LiQResolveResponse {
```
jbienkowski-fandom, line 9: this could be moved to `@types/live-connect/index.d.ts` and used to better describe resolve method `success: (nonId: LiQResolveResponse) => void`

jbienkowski-fandom (changes requested):

src/ad-services/live-connect/index.ts
szymon-nawrocki: Great suggestion! Done ✅

jbienkowski-fandom (approved):

jbienkowski-fandom (approved):

jbienkowski-fandom (approved):

PR #2029, opened 2023-07-03
ADEN-13377 - floor wait to appear when TLB received stickiness template

adi-wesolowski-fandom:
- made some refactor
- simplify checking the mapped ids - from now on, we put only Ids, without additional geo mapping - it is not necessary with geo-icbm-mapping.
- extend appear time for floor with TLB sticky to reduce time when showing both sticky slots at the same time
- 
JIRA task: https://fandom.atlassian.net/browse/ADEN-13377


src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
```diff
@@ -5,22 +5,26 @@ import {
...
+						stickyExaminer.shouldStick() &&
 						[
 							universalAdPackage.SLOT_UNSTICKED_STATE,
 							universalAdPackage.SLOT_FORCE_UNSTICK,
 							universalAdPackage.SLOT_STICKY_STATE_SKIPPED,
 							universalAdPackage.SLOT_VIDEO_DONE,
 						].includes(payload.status)
 					) {
-						setTimeout(() => callback(), universalAdPackage.SLIDE_OUT_TIME);
+						setTimeout(() => callback(), universalAdPackage.SLIDE_OUT_TIME + 400);
```
preludek, line 27: Why are we extendint the timeout by 400ms? Is there a race condition or something?

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -13,33 +13,15 @@ import {
 } from '@wikia/ad-engine';
 import { Inject, Injectable } from '@wikia/dependency-injection';
 
-@Injectable({ autobind: false })
-export class StickyTlbBlockingHandler implements TemplateStateHandler {
-	static LOG_GROUP = 'sticky-tlb';
-
-	constructor(@Inject(TEMPLATE.SLOT) private adSlot: AdSlot) {}
+export class StickyTlbAllowanceExaminer {
```
preludek, line 16: please stick to one class per file pattern

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
```diff
@@ -56,17 +38,47 @@ export class StickyTlbBlockingHandler implements TemplateStateHandler {
...
-				const [mapGamId, geo] = element.split(':', 2);
-
-				return mapGamId.toString() === gamId && (!geo || utils.geoService.isProperGeo([geo]));
+				return element.toString() === gamId;
 			});
 	}
+}
+
+@Injectable({ autobind: false })
+export class StickyTlbBlockingHandler implements TemplateStateHandler {
```
preludek, line 53: Maybe we can merge those two classes to something like StickyTlbHandler?

Tomasz-at-Fandom: @preludek what would be the benefit of that?
As a rule of thumb it's better to split to smaller pieces with separate responsibilities, unless we are duplicating something here?

src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
Tomasz-at-Fandom: If there is than please assign this to a const with a name that would hint at it.

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
adi-wesolowski-fandom: ok. 

I didnt realize we have such policy single-class-per-file, as I saw there is some files with multiple declarations.

src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
Tomasz-at-Fandom: If there are no downsides to having this here than let it stay, but put it in a const like ADDITIONAL_SLIDE_OUT_TIME or something like that. You can put a comment next that const for more clarification.

src/platforms/shared/templates/handlers/sticky-tlb/sticky-tlb-blocking-handler.ts
Tomasz-at-Fandom: Let's stick to one class per file, but don't go refactoring the old ones until there is an actual need to modify them.

adi-wesolowski-fandom: already separated

src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
adi-wesolowski-fandom: already done

src/platforms/news-and-ratings/shared/setup/dynamic-slots/news-and-ratings-dynamic-slots-neutron.setup.ts
```diff
@@ -120,8 +120,12 @@ export class NewsAndRatingsDynamicSlotsNeutronSetup implements DiProcess {
 				placementId: slot.getAttribute('data-ad'),
 			});
 		});
+
 		utils.logger(logGroup, 'Reinjecting floor_adhesion');
-		insertSlots([this.slotsDefinitionRepository.getFloorAdhesionConfig()]);
+
+		communicationService.on(eventsRepository.AD_ENGINE_UAP_NTC_LOADED, () =>
```
kwfandom, line 126: I believe It may block floor adhesion without NTC 🤔

adi-wesolowski-fandom: I have no idea. It is not my change. It was on dev and I paste it from conflict resolver, as I saw it was newest version. But, as I test, I saw behaviour we want is kept, nothing changed.

kwfandom: I believe on dev it looks like this https://github.com/Wikia/ad-engine/blob/dev/src/platforms/news-and-ratings/shared/setup/dynamic-slots/news-and-ratings-dynamic-slots-neutron.setup.ts#L124

dmnsobczak: I merged dev and this is still a change. Please review the PR :)

adi-wesolowski-fandom: ok, I change it to code without communication.

preludek (dismissed):

Jorewin (approved):
Looks good to me!

src/core/listeners/scroll-listener.ts
Jorewin, line 1: The way I understand this works right now is like this:
When Top Leaderboard is loaded we check if it should be sticky and based on that we either add a Floor Adhesion callback to the Scroll Listener or do so with a delay. Scroll Listener than omits 5 scroll event detections and only then it places the Adhesion.

I treat the Scroll Listener solution a time delay without explicit delay given, so the thing I'm mostly concerned about is that we have two arbitrary delays defined in two different places.

dmnsobczak:
New approach: https://github.com/Wikia/ad-engine/pull/2060

PR #2046, opened 2023-07-11
Aden 13348 Analysis of Performance Test

lnawrotek:


src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
```diff
@@ -163,4 +170,39 @@ export class UcpDesktopPerformanceAdsDefinitionRepository {
 	private triggerImpressionPixels(): void {
 		utils.scriptLoader.loadAsset(this.widgetData.data.impression, 'blob');
 	}
+
+	private trackElementVisibility(element: HTMLElement, type: PerformanceAdType): void {
+		if (!this.widgetData.type || !this.widgetData.type.includes(type)) {
+			return;
+		}
+
+		const minIntersectioneRatio = 0.5;
```
Jorewin, line 179: I think there is a minor typo in this variable's name.

src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
```diff
@@ -163,4 +170,39 @@ export class UcpDesktopPerformanceAdsDefinitionRepository {
...
+		const minVisibleTime = 500;
```
Jorewin, line 180: Mostly out of curiosity but how did you chose the `minVisibleTime` and the `threshold` parameters?

src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
```diff
@@ -163,4 +170,39 @@ export class UcpDesktopPerformanceAdsDefinitionRepository {
...
+
+		new IntersectionObserver(
+			(entries, observer) => {
+				entries.forEach((entry) => {
+					if (
+						entry.intersectionRatio > minIntersectioneRatio &&
+						entry.isIntersecting &&
```
Jorewin, line 187: If ratio is higher than minimum than I'm pretty sure it also implies intersection so we don't need to additionally check if entry intersect with root.

src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
```diff
@@ -163,4 +170,39 @@ export class UcpDesktopPerformanceAdsDefinitionRepository {
...
+				});
+			},
+			{ threshold: minIntersectioneRatio },
+		).observe(element);
+	}
+
+	private viewabilityCallToDW() {
+		this.dwTracker.track(
+			{
+				creative_id: this.widgetData.data.creativeId,
```
Jorewin, line 202: I couldn't get deep enough to understand what happens later in tracking but will passing `creative_id` or `line_item_id` as `undefined` here, potentially make us send obsolete parameters in a tracking request? What I'm suggesting is not setting the parameters if the values are `undefined` but I'm not sure if it would make a difference.

Jorewin (commented):
I only found some minor things but overall everything seems ok, so after they are resolved we can proceed.

src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
lnawrotek: fixed

src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-performance-ads-definition-repository.ts
lnawrotek: These values are the same as in GPT, I talked about it with Damian

Jorewin (dismissed):

adi-wesolowski-fandom (approved):

PR #2048, opened 2023-07-12
ADEN-13410 - gpt-ready event send to grafana

adi-wesolowski-fandom:


src/platforms/shared/setup/gpt.setup.ts
```diff
@@ -14,6 +15,7 @@ export class GptSetup extends BaseServiceSetup {
 		utils.logger('gpt-provider', 'loading GPT...');
 		return utils.scriptLoader.loadScript(GPT_LIBRARY_URL).then(() => {
 			utils.logger('gpt-provider', 'ready');
+			new MetricReporter().trackGptLibReady();
```
Jorewin, line 18: The only suggestion I have that it could be better to make the Metric Reporter set this as a callback in Communication Service as it seems to be a convention. It might be also overcomplicate things as we are not reacting to this event in any other way.

Jorewin (dismissed):
One suggestion, but everything seems to be ok otherwise.

PR #2051, opened 2023-07-13
ADEN-13372 | Set up incontent_leaderboard slot

rafalgiemza:
- only UCP-desktop
- add bidders setup:
  - indexexchange
  - pubmatic
  - triplelift
  - wikia


lnawrotek (approved):

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
```diff
@@ -28,6 +28,7 @@ export function getAppnexusContext(): object {
 					[160, 600],
 					[300, 600],
 					[300, 250],
+					[728, 90],
```
dmnsobczak, line 31: WTF. Incontent Boxad is in the right rail which is 300px wide. This size is not allowed there. Did you mean `incontent_leaderboard`?

dmnsobczak (dismissed):
Did you mean `incontent_leaderboard`?

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
rafalgiemza: right, fixed

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
```diff
@@ -31,6 +24,17 @@ export function getAppnexusContext(): object {
 				],
 				position: 'hivi',
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				position: 'hivi',
```
dmnsobczak, line 29: If we use `gallery` IDs, then it should be `gallery` to send proper `placementId` to AppNexus.

src/platforms/ucp-desktop/bidders/prebid/index-exchange.ts
```diff
@@ -16,6 +16,10 @@ export function getIndexExchangeContext(): object {
 				],
 				siteId: '183567',
 			},
+			incontent_leaderboard: {
```
dmnsobczak, line 19: Sometimes `incontent_leaderboard` if before `incontent_boxad_1` in the file and sometimes the other way around. Please reorganise it.

src/platforms/ucp-desktop/bidders/prebid/medianet.ts
```diff
@@ -31,6 +31,11 @@ export function getMedianetContext(): object {
 				cid: '8CU5JOKX4',
 				crid: '210414245',
 			},
+			incontent_leaderboard: {
```
dmnsobczak, line 34: I don't see placements for `gallery_leaderboard` in this file. Why did you use `bottom_leaderboard` placements?

src/platforms/ucp-desktop/bidders/prebid/nobid.ts
```diff
@@ -24,6 +24,10 @@ export function getNobidContext(): object {
 				],
 				siteId: 21872987104,
 			},
+			incontent_leaderboard: {
```
dmnsobczak, line 27: Same thing. One time you use `gallery` IDs, but for few you use BLB IDs. Are you trying to make analysis a little bit more exciting and trigger some anomaly alerts? :D

src/platforms/ucp-desktop/bidders/prebid/pubmatic.ts
```diff
@@ -37,6 +37,10 @@ export function getPubmaticContext(): object {
 					'/5441/INCONTENT_BOXAD_1_300x600@300x600',
 				],
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				ids: ['/5441/INCONTENT_LEADERBOARD_728x90@728x90'],
```
dmnsobczak, line 42: This is a constant, not numeric but `string` type. I don't think we can construct it however we like.

src/platforms/ucp-desktop/bidders/prebid/roundel.ts
```diff
@@ -24,6 +24,10 @@ export function getRoundelContext(): object {
 				],
 				siteId: '820933',
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				siteId: '820933',
```
dmnsobczak, line 29: This time you used `top_boxad` IDs, why?

src/platforms/ucp-desktop/bidders/prebid/rubicon-display.ts
```diff
@@ -39,6 +39,14 @@ export function getRubiconDisplayContext(): object {
 				siteId: '148804',
 				zoneId: '704676',
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				targeting: {
+					loc: ['hivi'],
```
dmnsobczak, line 45: `incontent_leaderboard` is not a highly viewable unit - using `hivi` targeting here is a logical error.

src/platforms/ucp-desktop/bidders/prebid/triplelift.ts
```diff
@@ -28,6 +28,10 @@ export function getTripleliftContext(): object {
 					'Fandom_DT_FMR_300x600_hdx_prebid',
 				],
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				inventoryCodes: ['Fandom_DT_GALLERIES_728x90_hdx_prebid'],
```
dmnsobczak, line 33: Are we missing IDs for galeries for this bidder? I see you used IDs for galery_leaderboard @ UCP-Desktop, but there is no config for it in this file.

src/platforms/ucp-desktop/bidders/prebid/verizon.ts
```diff
@@ -15,6 +15,10 @@ export function getVerizonContext(): object {
 				sizes: [[300, 250]],
 				pos: 'incontent_boxad_1',
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				pos: 'incontent_leaderboard',
```
dmnsobczak, line 20: It might look fine, but creating own `pos` values might give 40x errors for requests. Did you check http request to Verizon?

src/platforms/ucp-desktop/bidders/prebid/wikia.ts
```diff
@@ -11,6 +11,9 @@ export function getWikiaContext(): object {
 			incontent_boxad_1: {
 				sizes: [[300, 250]],
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
```
dmnsobczak, line 15: I'm glad you did not forget about wikia adapter :) Galleries are missing here though. Can you add it as a bonus? 🥺

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
rafalgiemza: fixed

src/platforms/ucp-desktop/bidders/prebid/medianet.ts
rafalgiemza: true, that's wrong usage, I'm skipping this partner.

src/platforms/ucp-desktop/bidders/prebid/nobid.ts
rafalgiemza: of course not, good catch, thanks!

src/platforms/ucp-desktop/bidders/prebid/roundel.ts
rafalgiemza: `top_boxad` and `incontent_boxad_1` already got the same `siteId`
I thought that is ok, if not let's exclude this partner

src/platforms/ucp-desktop/bidders/prebid/rubicon-display.ts
rafalgiemza: Now I see, that for _rubicon-display_ there is already configured `fandom_dt_galleries` slot. Skipping this partner

src/platforms/ucp-desktop/bidders/prebid/verizon.ts
rafalgiemza: I wasn't aware, that this pos have to be configured somewhere else.

src/platforms/ucp-desktop/bidders/prebid/wikia.ts
rafalgiemza: sure :)

src/platforms/ucp-desktop/bidders/prebid/pubmatic.ts
rafalgiemza: but it can make analysis wrong, so a safer way is to skip this partner

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
```diff
@@ -31,6 +24,17 @@ export function getAppnexusContext(): object {
 				],
 				position: 'hivi',
 			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				position: 'gallery',
```
rafalgiemza, line 29: I wonder if a gallery is the best choice here. Maybe native will be better since this ad is in the middle of content?

kwfandom:
What about Magnite and 3lift?

src/platforms/ucp-desktop/bidders/prebid/appnexus.ts
kwfandom: Im not an expert, but from the code perspective position doesn't mean that something is atf, btf or in the middle of the content. At first glance position is equal to placement id. We are missing that for incontent leaderboard, so my suggestion is to wait, till id will be provided, because usage of same id for two different slots may cause side effects. But im open for discussion @adi-wesolowski-fandom @lnawrotek

adi-wesolowski-fandom: "position" for appnexus is an optional parameter. So the question should be directed to James Wu. If we dont have it in his spreadsheet, it may be whatever we put here. I guess its mainly for reporting. I recomment asking James directly, does they use, how, and if they do (at least for reporting purpose), what they want here to be set. 

But I guess, Karol advice to put here "middle" is good. This is the value we put in the standard GAM targeting in "loc" KV. So to be consistent, it would be a good direction to put "middle" here, instead "gallery", which IMO should be reserved only for slot in gallery (at least for now, as far as I know)

kwfandom: Yes, but when you dig into our app nexus adapter, there's code which is is mapping position to placement id or am i wrong 🤔

rafalgiemza: let's remove this parnter for now

kwfandom (approved):

PR #2054, opened 2023-07-17
ADEN-12881-system1

lnawrotek:
https://fandom.atlassian.net/browse/ADEN-12881


src/ad-services/system1/index.ts
```diff
@@ -0,0 +1,143 @@
...
+	desktop: 'fan00',
+	darkTheme: 'fandomdark',
+};
+const themes = {
+	dark: 'dark',
+	light: 'light',
+};
+
+const cacheKey = 'adEngine_system1';
+const cacheTtl = 24 * 3600;
```
rafalgiemza, line 18:
```suggestion
const cacheTtl = 86400; // 24 * 3600;
```

src/ad-services/system1/index.ts
```diff
@@ -0,0 +1,143 @@
...
+export class System1 extends BaseServiceSetup {
+	private isLoaded = false;
+
+	call(): Promise<void> {
+		if (!this.isSearchPage()) {
+			utils.logger(logGroup, 'disabled - it is not search page');
+			return Promise.resolve();
+		}
+
+		if (!this.isEnabled('icSystem1', false)) {
```
rafalgiemza, line 29: what do you think about moving icSystem1 to const?

src/ad-services/system1/index.ts
```diff
@@ -0,0 +1,143 @@
...
+				return 'video';
+			case 'imageOnly':
+				return 'image';
+			default:
+				return 'web';
+		}
+	}
+
+	private getSubId(): string {
+		const formattedUrl = window.location.protocol + '//' + this.getHostname();
```
rafalgiemza, line 87:
```suggestion
		const formattedUrl = `${window.location.protocol}//${this.getHostname()}`;
```

src/ad-services/system1/index.ts
```diff
@@ -0,0 +1,143 @@
...
+	private getTheme(): string {
+		return (window.mw as any)?.config?.get('isDarkTheme') ? themes.dark : themes.light;
+	}
+
+	private isThemeChanged() {
+		const cacheValue = localCache.getItem(cacheKey);
+
+		if (!cacheValue || cacheValue !== this.getTheme()) {
+			localCache.setItem(cacheKey, this.getTheme(), cacheTtl);
+			utils.logger(logGroup, 'New session');
```
rafalgiemza, line 127: New session when the theme changed? I don't get it...

src/ad-services/system1/index.ts
```diff
@@ -0,0 +1,143 @@
...
+			return true;
+		}
+
+		return false;
+	}
+
+	private onSetupResolve(): void {
+		utils.logger(logGroup, 'completed');
+		communicationService.emit(eventsRepository.SYSTEM1_LOADER);
```
rafalgiemza, line 136:
```suggestion
		communicationService.emit(eventsRepository.SYSTEM1_STARTED);
```

rafalgiemza (changes requested):
some comments left

src/ad-services/system1/index.ts
rafalgiemza: This needs to be fixed

src/ad-services/system1/index.ts
lnawrotek: We don't use const for names of ICBM variables.

src/ad-services/system1/index.ts
lnawrotek: It's not a browser session. We need to send "newSession" parameter to System1 after the theme change. I add this parameter if the current theme is different than the theme in the cache, it means that the user has changed the theme.

rafalgiemza: My comment was only about the log message - it confused me.
So maybe to avoid extra explanation in the future we can have a log message like this:
Theme changed, 'new session' event emitted to System1

src/ad-services/system1/index.ts
rafalgiemza: 👍

PR #2060, opened 2023-07-20
ADEN-13377-2 Fixing issues with StickyTLB + Floor integration

dmnsobczak:


src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
```diff
@@ -23,10 +26,33 @@ export function activateFloorAdhesionOnUAP(callback: () => void, withLoadedOnly
 						setTimeout(() => callback(), universalAdPackage.SLIDE_OUT_TIME);
 					}
 				},
-				'top_leaderboard',
+				firstCallSlotName,
 			);
 		} else if (!withLoadedOnly) {
-			callback();
+			if (!context.get(`slots.${firstCallSlotName}.defaultTemplates`).includes('stickyTlb')) {
```
adi-wesolowski-fandom, line 32: it does not work properly when we dont have stickyTlb ad. eg., when disabling Tlb, floor will not show up: https://project43.fandom.com/wiki/SyntheticTests/LongPage?icbm__icForceStickyTlb=false&cid=adeng-floor-adhesion&icbm__icUAPJWPlayer=true&adengine_version=local (or maybe that case does not exist?). 

Anyway, IMO, using context-default-template as a main condition for showing or not floor, may be risky, as there is much async going on here, including mixing with default behaviour. I think, it would be more robust, to use same condition as we have for enabling stickiness but reversed (NOT). So in that case, I would suggest refactor and extrude code for checking activation of Tlb in a separate class, like I did here: https://github.com/Wikia/ad-engine/pull/2029/files to have consistent conditions, without chaining condition for each step one after another. Let it be one place to check if we had sticikness, regardless checking if effect of this took place.

dmnsobczak: It was a typo, I was accessing `eventsRepository.SLOT_STICKINESS_DISABLED` instead of `universalAdPackage.SLOT_STICKINESS_DISABLED` and it was undefined :< Fix pushed.

dmnsobczak: I chose different approach, because previous one was over complicated to me and it was not including few edge cases. After adding them class and StickyTLB setup was hard to read, so I scrapped everything and tried from the beginning.

lnawrotek (approved):

src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
adi-wesolowski-fandom: I am not talking about changing all the logic here, but only about refactor and extracting decision logic of that is tlb should or should not be sticky, from oversized class, and changing complicated condition of three elements (with &&) into one. I dont see how this refactor could be "over complicated" compared to what we got now, but ... ok. I wont instist

PR #2068, opened 2023-07-26
TACO-18 Configure AB testing for ID5

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-18

### Description
Configure A/B testing for ID5


src/ad-bidders/prebid/id5/id5.ts
```diff
@@ -1,24 +1,49 @@
...
 
-		return {
+		const config = {
 			name: 'id5Id',
 			params: {
 				partner: 1139,
 				abTesting: {
 					enabled: false,
-					controlGroupPct: 0.1,
+					controlGroupPct: 0.5,
```
domjanik, line 46: Does it say, that our A/B groups will be divided equally ? (50%-50%) If it does - maybe it would be good to somehow move it to ICBM, as it may change in the future 🤔

domjanik (approved):

domjanik (approved):

PR #2076, opened 2023-07-28
ADEN-13382 No ads (or tlb only) on metacritic/news 

lnawrotek:


kwfandom (dismissed):

rafalgiemza (approved):
as a hot fix with followup its fine

PR #2078, opened 2023-08-01
ADEN-13453 Workaround for lack of tlb and floor adhesion issue

kwfandom:
The issue with floor adhesion is caused by lack of tlb, because we are waiting for STICKINESS_DISABLED event from the tlb to inject floor adhesion, but we haven't got tlb, so the event won't be emitted. That's workaround for setting empty template for tlb, if tlb is not exisiting on N&R sites.

### DIFF
--- a/src/platforms/news-and-ratings/shared/setup/dynamic-slots/news-and-ratings-dynamic-slots.setup.ts
+++ b/src/platforms/news-and-ratings/shared/setup/dynamic-slots/news-and-ratings-dynamic-slots.setup.ts
@@ -61,6 +61,7 @@ export class NewsAndRatingsDynamicSlotsSetup implements DiProcess {
 
 		if (!tlbExists) {
 			communicationService.on(eventsRepository.AD_ENGINE_STACK_START, () => {
+				context.set(`slots.top_leaderboard.defaultTemplates`, []);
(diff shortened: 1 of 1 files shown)


src/platforms/news-and-ratings/shared/setup/dynamic-slots/news-and-ratings-dynamic-slots.setup.ts
```diff
@@ -61,6 +61,7 @@ export class NewsAndRatingsDynamicSlotsSetup implements DiProcess {
 
 		if (!tlbExists) {
 			communicationService.on(eventsRepository.AD_ENGINE_STACK_START, () => {
+				context.set(`slots.top_leaderboard.defaultTemplates`, []);
```
dmnsobczak, line 64: I wanted to propose a manual "stickiness-disabled" event trigger, but since it is a slot-connected event, then maybe that's not a good idea. This above is fine, but what about that?
`context.set('templates.stickyTlb.disabled', true);`
And then you can check this value here:
https://github.com/Wikia/ad-engine/blob/546f6e2db8fbb9b92db90e48cce356a020f21dcf/src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts#L33C22-L33C22

PR #2093, opened 2023-08-11
ADEN-13478 - add few slots, as a start for unification process

adi-wesolowski-fandom:


src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
```diff
@@ -283,6 +283,74 @@ export class MetacriticNeutronSlotsContextSetup implements DiProcess {
...
 			},
+			incontent_leaderboard: {
+				sizes: [
+					{
+						viewportSize: [970, 200],
+						sizes: [
+							[728, 90],
+							[970, 66],
+							[5, 5],
+							universalAdPackage.UAP_ADDITIONAL_SIZES.bfaSize.desktop,
```
kwfandom, line 294: incontent leaderboard is not a part of UAP, so those sizes i believe should be deleted

adi-wesolowski-fandom: ok, done

kwfandom:
if we want to use those slots, we should add also bidder configurations

adi-wesolowski-fandom:
> if we want to use those slots, we should add also bidder configurations

We cant at the moment, coz we dont have bidders ID right now. This need to be done in some other task and give James A note to prepare tho. I will create such task

src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
```diff
@@ -283,6 +283,84 @@ export class MetacriticNeutronSlotsContextSetup implements DiProcess {
...
 				},
 			},
+			incontent_leaderboard: {
+				sizes: [
+					{
+						viewportSize: [970, 200],
+						sizes: [
+							[728, 90],
+							[970, 66],
+							[5, 5],
```
dmnsobczak, line 293: Ditto, please remove 5x5

src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
```diff
@@ -283,6 +283,84 @@ export class MetacriticNeutronSlotsContextSetup implements DiProcess {
...
+						],
+					},
+					{
+						viewportSize: [840, 200],
```
dmnsobczak, line 336: In some places we detect mobile using an user agent plugin - we store this value in `state.isMobile`. Here you using just breakpoints. If that's really will be an unification, you should unify that too, because this will be a problem in the future for sure.

adi-wesolowski-fandom: I used same pattern as for TLB. Isnt you who adviced to use this? :) 

If there will be a problem with it, we will fix it in follow up ticked, together with TLB and other slots defined here, and on other platforms as well

src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
adi-wesolowski-fandom: BUT... we have 5x5 sizes on many other slots also, multiple n&r incontent slots for a long time - this is why I added this as well

PR #2094, opened 2023-08-11
COTECH-623 Update s2 with a page type for N&R

nandy-andy:
https://fandom.atlassian.net/browse/COTECH-623


@types/fandom/index.d.ts
```diff
@@ -54,10 +54,16 @@ interface ITargetingParameters {
 	[key: string]: string | number;
 }
 
+interface ITargetingParameters {
```
pbartkowiak-fandom, line 57: The `ITargetingParameters` interface is defined 4 lines above your declaration. Did you declared it for the second time on purpose?

@types/fandom/index.d.ts
```diff
@@ -54,10 +54,16 @@ interface ITargetingParameters {
...
+	playerName: string;
+}
+
 interface WindowFandomContext {
 	site: Partial<ISiteParameters>;
 	page: Partial<IPageParameters>;
 	tracking: Partial<ITrackingParameters>;
 	targeting: Partial<ITargetingParameters>;
 	partners: Partial<IPartnersParameters>;
+	video?: Partial<IVideoParameters>;
```
pbartkowiak-fandom, line 68: `IVideoParameters` interface declaration is missing. Did you mean `IVideoParameters` in line 57?

pbartkowiak-fandom, line 68: I don't think an interface declaration is a proper use case for the `Partial` utility type. The suggested use cases that I found are:
- When you need to only update some fields in an existing object
- When you need to pass constructor values to populate a new instance of the class

(source: https://timmousk.com/blog/typescript-partial/) 

Of course, the rule is not set in stone. Still, I would rather see explicite declarations:

```typescript
interface IVideoParameters {
	playerLoaded?: boolean;
	playerName?: string;
}

interface WindowFandomContext {
	...
	video: IVideoParameters;
}
```

WDYT?

@types/fandom/index.d.ts
nandy-andy: Nope. Good catch, thanks! :)

@types/fandom/index.d.ts
nandy-andy: Yep 🙈

nandy-andy: I have no strong feelings. I'm happy with your suggestion :) 👍 Thank you!

domjanik (approved):

pbartkowiak-fandom (approved):

dmnsobczak (commented):
👍

PR #2095, opened 2023-08-14
TACO-136 IIQ Additional targeting key val

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-136

### Description
Add additional targeting key value `intent_iq_ppid_group` with a copy of the same info (A/B/U) the `intent_iq_group` key-val carries


domjanik (approved):

dmnsobczak (commented):
👍

PR #2097, opened 2023-08-14
Aden 13480

adi-wesolowski-fandom:


lnawrotek (dismissed):

src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
```diff
@@ -283,6 +283,150 @@ export class MetacriticNeutronSlotsContextSetup implements DiProcess {
...
+							[728, 90],
+							[970, 66],
+							[5, 5],
+						],
+					},
+					{
+						viewportSize: [840, 200],
+						sizes: [
+							[728, 90],
+							[5, 5],
```
dmnsobczak, line 300: 5x5 is an old N&R skybox size and we sunsetted this size - please remove it. Moreover: 5x5 was used only for TLB slots

adi-wesolowski-fandom: ok

PR #2100, opened 2023-08-17
Aden 13100 fixing

adi-wesolowski-fandom:


src/platforms/news-and-ratings/metacritic-neutron/setup/context/slots/metacritic-neutron-slots-context.setup.ts
```diff
@@ -283,6 +283,140 @@ export class MetacriticNeutronSlotsContextSetup implements DiProcess {
...
+							[970, 66],
+						],
+					},
+					{
+						viewportSize: [840, 200],
+						sizes: [[728, 90]],
+					},
+					{
+						viewportSize: [320, 200],
+						sizes: [
```
kwfandom, line 301: it's desktop slot

adi-wesolowski-fandom: it may be, on fandom. But, what I did here was to set some secure step: if, for any reason, this slot appear on mobile I want to make sure it wont call for desktop sizes, but for smaller, mobile sizes. We can later, after Goliath release, find this places and fix it, adjusting name of incontents to fit fandom - in the "unification" step. Right now I want to make sure it is working, and working good way

mfarul (approved):

PR #2106, opened 2023-08-22
TACO-35 - Added Ozone integration

domjanik:


src/ad-bidders/prebid/index.ts
```diff
@@ -134,7 +134,16 @@ export class PrebidProvider extends BidderProvider {
 						filter: 'include',
 					},
 				},
-				userIds: [],
+				userIds: [
```
szymon-nawrocki, line 137: This can be moved to `configureUserSync` method :https://github.com/Wikia/ad-engine/blob/9e0cee8681b357e2585a4988633dbead85afaf0e/src/ad-bidders/prebid/index.ts#L198
where the other `userIds` are set. Plus it will enable ability of not adding the config when Ozone is disabled.

src/ad-bidders/prebid/index.ts
```diff
@@ -134,7 +134,16 @@ export class PrebidProvider extends BidderProvider {
...
+					{
```
szymon-nawrocki, line 138: This can be moved to `configureUserSync` method :https://github.com/Wikia/ad-engine/blob/9e0cee8681b357e2585a4988633dbead85afaf0e/src/ad-bidders/prebid/index.ts#L198
where the other `userIds` are set. Plus it will enable ability of not adding the config when Ozone is disabled.

szymon-nawrocki (approved):

szymon-nawrocki (approved):

szymon-nawrocki (approved):

PR #2107, opened 2023-08-23
ADEN-13071 Added force BT behavior with disabled WAS ad block check

verhopara:
JIRA [ADEN-13071](https://fandom.atlassian.net/browse/ADEN-13071)

- Added Blocktrough force behavior which omits our wikia ad block check and forces to load blockthrough script
- To turn on behavrior use instant config `icBTForce` variable (example https://pokemon.fandom.com/wiki/Pok%C3%A9mon_Wiki?adengine_version=local&icbm__icBTForce=true)

[ADEN-13071]: https://fandom.atlassian.net/browse/ADEN-13071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/platforms/shared/services/wad-runner.ts
```diff
@@ -17,17 +18,24 @@ export class WadRunner extends BaseServiceSetup {
...
 
 	async call(): Promise<void> {
-		if (!this.detector.isEnabled()) {
-			return Promise.resolve();
+		if (this.instantConfig.get('icBTForce')) {
+			context.set('options.wad.btForce', true);
 		}
+		if (context.get('options.wad.btForce')) {
+			btForce.run();
+		} else {
```
adi-wesolowski-fandom, line 26: nitpick: `else` might be removed if we do "quick-return/resolve" in the condition block above

adi-wesolowski-fandom (dismissed):

src/platforms/shared/services/wad-runner.ts
adi-wesolowski-fandom: thx

adi-wesolowski-fandom (dismissed):

src/ad-services/bt-force/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * Injects BT script
+	 */
+	private loadScript(): Promise<Event> {
+		const btLibraryUrl = 'https://btloader.com/tag?o=5199505043488768&upapi=true';
+		return utils.scriptLoader.loadScript(btLibraryUrl, true, 'first');
+	}
+
+	private insertSideUnits(): void {
```
preludek, line 67: This is pretty much duplicated from bt-rec. What do you think about moving the logic for example to `platforms/dynamic-slots` and just call it after detecting ad block?

preludek (approved):
LGTM 👍 left one thing for your consideration

src/ad-services/bt-force/index.ts
preludek: 👍

preludek (dismissed):

preludek (approved):

PR #2113, opened 2023-08-25
TACO-104 Add Lotame pixel

jakubbanas:
Lotame pixel integration


src/platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
```diff
@@ -17,6 +17,7 @@ import {
 	jwPlayerInhibitor,
 	LiveConnect,
 	LiveRampPixel,
+	Lotame,
```
domjanik, line 20: What about other platforms? UCP-mobile, N&R etc?

src/ad-services/lotame/index.ts
```diff
@@ -0,0 +1,38 @@
...
+			return;
+		}
+		utils.logger(this.logGroup, 'pixel enabled');
+		await utils.scriptLoader.loadScript(this.PIXEL_URL);
+		communicationService.dispatch(eventsRepository.LOTAME_LOADED);
+
+		const lotameTagInput = {
+			data: {},
+			config: {
+				clientId: Number(this.CLIENT_ID),
```
domjanik, line 27: CLIENT_ID is number anyway, therefore I think it's pointless to do that conversion.

src/ad-services/lotame/index.ts
```diff
@@ -0,0 +1,38 @@
...
+	async call(): Promise<void> {
+		if (!this.isEnabled('icLotame')) {
+			utils.logger(this.logGroup, 'pixel disabled');
+			return;
+		}
+		utils.logger(this.logGroup, 'pixel enabled');
+		await utils.scriptLoader.loadScript(this.PIXEL_URL);
+		communicationService.dispatch(eventsRepository.LOTAME_LOADED);
+
+		const lotameTagInput = {
```
domjanik, line 24: Shouldn't we create that config object BEFORE loading a pixel ? 🤔

jakubbanas: By documentation, it doesn't seem to matter, but it looks like a better idea.
It still doesn't work, but looks better :D

src/ad-services/lotame/index.ts
jakubbanas: Good point

src/platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
jakubbanas: Right, forgot about it

PR #2118, opened 2023-08-29
ADEN-13527 Disable System1 on kid wikis

lnawrotek:


src/ad-services/system1/index.ts
```diff
@@ -26,7 +26,7 @@ export class System1 extends BaseServiceSetup {
 			return Promise.resolve();
 		}
 
-		if (!this.isEnabled('icSystem1', false)) {
+		if (!this.isEnabled('icSystem1', false) || this.isKidWiki()) {
```
mfarul, line 29:
```suggestion
		if (!this.isEnabled('icSystem1', false) || utils.isCoppaSubject()) {
```

mfarul: and we can get rid of the kid wiki direct check - util is handling this plus Age gate, e.g. when adult is comming to kid wiki

mfarul (changes requested):

kwfandom (approved):

PR #2119, opened 2023-08-29
ADEN-13440 - update path to gpt

adi-wesolowski-fandom:


src/platforms/shared/setup/gpt.setup.ts
```diff
@@ -10,7 +10,7 @@ export class GptSetup extends BaseServiceSetup {
 	}
 
 	call(): Promise<void> {
-		const GPT_LIBRARY_URL = '//www.googletagservices.com/tag/js/gpt.js';
+		const GPT_LIBRARY_URL = 'https://securepubads.g.doubleclick.net/tag/js/gpt.js';
```
lnawrotek, line 13: I'm not sure if we use "http" somewhere, but maybe it's safer to use "//"

adi-wesolowski-fandom: https://developers.google.com/publisher-tag/guides/general-best-practices#load_securely

> Always load the GPT library over HTTPS

And I dont think, we have anything working on HTTP anyway.

lnawrotek (approved):

PR #2122, opened 2023-08-30
[ADEN-13484] disabled localStorage/sessionStorage behavior

verhopara:
https://fandom.atlassian.net/browse/ADEN-13484

- Changed constructor provider setup from `(a: b)` to `(a: () => b)` since plain call to the window.localStorage causes errors in case local/session storage is disabled in chrome.
- Added InMemory fallback mechanism
- Updated universal-storage logger to use ad-engine logger instead of plain console.log()


src/core/services/universal-storage.ts
```diff
@@ -1,21 +1,62 @@
...
+	}
+}
 export class UniversalStorage implements Storage<any> {
-	private fallbackStorage: Dictionary = {};
+	private fallbackStorage = new InMemoryStorage();
+	private provider: StorageProvider;
 
-	constructor(private provider: StorageProvider = window.localStorage) {}
+	constructor(storageProvider: StorageAdapter = () => window.localStorage) {
+		let useFallback = false;
```
jbienkowski-fandom, line 42: (nit) this `useFallback` variable is not necessary

src/core/services/universal-storage.ts
```diff
@@ -1,21 +1,62 @@
...
+		try {
+			// Chrome throws error even during simple attempt to access local or session storage if they are disabled
+			// So before setItem / removeItem test let's try setup provider first
+			this.provider = storageProvider();
+			// Next test storage
+			// In case of local / session storage are not available it will throw error during setItem attempt
+			if (!this.isAvailable()) {
+				useFallback = true;
```
jbienkowski-fandom, line 50:
```suggestion
				this.provider = this.fallbackStorage;
```

src/core/services/universal-storage.ts
```diff
@@ -1,21 +1,62 @@
...
+			}
+		} catch {
+			useFallback = true;
```
jbienkowski-fandom, line 53:
```suggestion
			this.provider = this.fallbackStorage;
```

jbienkowski-fandom (dismissed):

src/core/services/universal-storage.ts
```diff
@@ -1,21 +1,59 @@
...
+		this.storage[key] = input;
+	}
+}
 export class UniversalStorage implements Storage<any> {
-	private fallbackStorage: Dictionary = {};
+	private fallbackStorage = new InMemoryStorage();
+	private provider: StorageProvider;
 
-	constructor(private provider: StorageProvider = window.localStorage) {}
+	constructor(storageProvider: StorageAdapter = () => window.localStorage) {
```
Jorewin, line 41: I think `storageProvider` should be called `storageAdapter` because `storageProvider` would be a name of a variable it returns when called.

src/core/services/universal-storage.ts
```diff
@@ -38,7 +76,7 @@ export class UniversalStorage implements Storage<any> {
 			}
 			return value as any;
 		} catch (e) {
```
Jorewin, line 78: Not really in scope of this CR, but this block should probably not include the last 6 lines.

src/core/services/universal-storage.ts
Jorewin, line 1: I'm a little bit concerned in how Universal Storage behaves. On every error we fallback to a dictionary but by going back and forth between it and the normal storage we can end up with weird variables. I get that it's there to ensure the actions always get executed somewhere but imagine for example something first being saved in the fallback storage due to an error, then the normal storage being cleared (no error) and then a value being pulled out from the fallback storage (an error again) which hasn't been cleared.

Jorewin: It's probably a job for an entirely new ticket but I was thinking of making it so when an operation fails, the storage becomes dirty. Then when there is a need to access the storage again, we check if the default option is available again and if so we sync the contents of the fallback with the default.

Jorewin (approved):

PR #2130, opened 2023-09-06
TACO-92 Hide ID5 auctionDelay behind ICBM flag

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-92

### Description
Hide ID5 ID `auctionDelay` parameter behind ICBM flag so we test it gradually


src/platforms/shared/bidders/bidders-state.setup.ts
```diff
@@ -70,6 +70,10 @@ export class BiddersStateSetup implements DiProcess {
 				this.instantConfig.get('icPrebidId5Analytics', false),
 			);
 			context.set('bidders.prebid.id5AbValue', this.instantConfig.get('icPrebidId5AB', 0));
+			context.set(
+				'bidders.prebid.id5AuctionDelay',
+				this.instantConfig.get('icPrebidId5AuctionDelay', 50),
```
mfarul, line 75:
```suggestion
				this.instantConfig.get('icPrebidAuctionDelay', 50),
```

szymon-nawrocki: The suggestion I am assuming is to control IntentIQ auctionDelay with the same iCBM variable. I'm thinking that maybe it would be better to create a separate variable for IntentIQ to give us more flexibility. For example ID5 ID may need a little longer delay than IntentIQ which is preloaded. @mfarul what do you think?

szymon-nawrocki: Great suggestion, done ✅

jakubbanas (approved):

PR #2131, opened 2023-09-07
TACO-148 Map Fandom key-vals to Lotame

jakubbanas:


szymon-nawrocki (dismissed):

src/ad-services/lotame/index.ts
```diff
@@ -17,11 +17,21 @@ export class Lotame extends BaseServiceSetup {
 			utils.logger(this.logGroup, 'pixel disabled');
 			return;
 		}
+
+		const siteTags = { ...window.fandomContext?.site?.tags, bundles: [] };
```
szymon-nawrocki, line 21: I think it wouldn't hurt to move it to a separate method and unit test it

jakubbanas: Good point, done!

szymon-nawrocki (approved):

PR #2138, opened 2023-09-11
TACO-169 - rework IntentIQ to support timeout set via ICBM

mfarul:


spec/ad-bidders/prebid/intent-iq/intent-iq.spec.ts
```diff
@@ -62,42 +46,6 @@ describe('IntentIQ', () => {
 		global.sandbox.restore();
 	});
 
-	describe('initialize', () => {
```
domjanik, line 65: It was removed as whole logic responsible for loading IntentIQ has been changed

domjanik (dismissed):

domjanik:
Changes in this PR:
- Moved loading IIQ earlier
- Removed IIQ PPID Logic
- Added setting IIQ Timeout via ICBM (icIntentIqTimeout)
- Added 3 new tracking events for IIQ

src/ad-bidders/prebid/intent-iq/intent-iq.ts
```diff
@@ -1,85 +1,75 @@
...
-				if (context.get('services.intentIq.ppid.enabled')) {
-					targetingService.set(
-						'intent_iq_ppid_group',
-						this.intentIqObject.intentIqConfig.abTesting.currentTestGroup || 'U',
-					);
-				}
+			this.intentIqObject = new window.IntentIqObject({
+				partner: this.fandomId,
+				pbjs,
+				timeoutInMillis: context.get('services.intentIq.timeout') || 2000,
```
mfarul, line 58: does it make sense to have it different than prebid auction delay ? This call modifies prebid setup, so changing it after first auction will only support incontent slots that are lazy bidded

domjanik: At this point, I think it's better to have 2 separate flags. I'm afraid that if we'll have a single flag to modify more than one thing, results from such tests might be inaccurate. Having 2 separate flags allows us to adjust all values to get the best possible result.

src/ad-bidders/prebid/intent-iq/intent-iq.ts
```diff
@@ -1,85 +1,75 @@
...
 
-	async initialize(pbjs: Pbjs): Promise<void> {
-		if (!this.isEnabled()) {
-			utils.logger(logGroup, 'disabled');
-			return;
-		}
+	private async init() {
+		return pbjsFactory.init().then((pbjs: Pbjs) => {
+			communicationService.on(eventsRepository.AD_ENGINE_CONSENT_READY, () => {
+				if (!this.isEnabled()) {
```
szymon-nawrocki, line 30: This is redundant since we call `init` only if the IntentIQ is enabled

src/platforms/shared/utils/instant-config.setup.ts
```diff
@@ -35,11 +35,10 @@ export class InstantConfigSetup implements DiProcess {
 		if (instantConfig.get('icPrebid')) {
 			context.set('bidders.prebid.libraryUrl', instantConfig.get('icPrebidVersion'));
 
-			pbjsFactory.init().then(() => {
-				if (instantConfig.get('icPrebidIntentIQ')) {
-					return intentIQ.preloadScript();
-				}
-			});
+			pbjsFactory.init();
```
szymon-nawrocki, line 38:
```suggestion
```

szymon-nawrocki: This is not needed anymore, pbjs object is now created in IntentIQ.init() method

domjanik: But it will be created only after IIQ is being loaded and launched. It might delay whole PBJS. I think it's better to leave it at it is

PR #2140, opened 2023-09-12
ADEN-13541 - delay pipeline, waiting for onload event when set in icbm

adi-wesolowski-fandom:


src/core/pipeline/imps/process-pipeline/compound-processes/when-page-loaded-process.ts
```diff
@@ -0,0 +1,61 @@
+import { Container, Injectable } from '@wikia/dependency-injection';
+import { MetricReporter } from '../../../../../platforms/shared';
+import { InstantConfigService } from '../../../../services';
+import { ProcessPipeline } from '../process-pipeline';
+import { CompoundProcess, CompoundProcessStep, ProcessStepUnion } from '../process-pipeline-types';
+
+const WAIT_FOR_PAGE_LOAD_TIMEOUT = 2000;
```
mfarul, line 7: ideally it should be ICBM variable, can also be used to start testing, e.g. no value - don't wait

src/core/pipeline/imps/process-pipeline/compound-processes/when-page-loaded-process.ts
```diff
@@ -0,0 +1,61 @@
...
+			return pipeline.add(...payload).execute();
+		}
+
+		if (document.readyState === 'complete') {
+			this.addExecutePipeline(pipeline, payload);
+		}
+
+		this.runSafeTimeout(pipeline, payload);
+		window.onload = () => {
+			this.addExecutePipeline(pipeline, payload);
```
mfarul, line 30: kill timeout here, even with fired variable let's free up some browser resources

mfarul (changes requested):

src/core/pipeline/imps/process-pipeline/compound-processes/when-page-loaded-process.ts
adi-wesolowski-fandom: I think, its good to leave some kind of default value, in case someone miss any domain/page, or geo would be skipped? But it's certainly no problem, I will add reading icbm variable for timeout as well.

src/core/pipeline/imps/process-pipeline/compound-processes/when-page-loaded-process.ts
adi-wesolowski-fandom: to be honest, I think it will cost us more (more bytes to send as we need more code for clearing) than letting it run without action (timeout is lightweight for browsers IMO). But ok, I added this clearing.

mfarul (dismissed):
Let's move on, more changes required

mfarul (approved):

src/core/pipeline/imps/process-pipeline/compound-processes/phases/dom-content-loaded-phase-process.ts
```diff
@@ -0,0 +1,78 @@
...
+	private killSafeTimeout(): void {
+		clearTimeout(this.timeoutId);
+	}
+
+	private addExecutePipeline(
+		pipeline: ProcessPipeline,
+		payload: ProcessStepUnion<T>[],
+	): Promise<void> | void {
+		if (this.fired) {
+			console.log('AEPERF fired');
```
adi-wesolowski-fandom:
```
console.log('[AEPERF] fired');
```

adi-wesolowski-fandom:
logic moved to "warp" branch

PR #2142, opened 2023-09-13
TACO-184 Only read from basset cookie, now set by IE

jbienkowski-fandom:
So IE and AE can use the same set of cached values.

[Other part](https://github.com/Wikia/identity-engine/pull/201)


src/core/services/instant-config-cache-storage.ts
```diff
@@ -38,7 +41,7 @@ export class InstantConfigCacheStorage implements InstantConfigCacheStorageServi
 	resetCache(): void {
 		const serializedCache = this.cookieStorage.getItem<string>(this.cacheKey) || '';
 		this.cacheStorage = deserializeCache(serializedCache);
-		communicationService.emit(eventsRepository.AD_ENGINE_INSTANT_CONFIG_CACHE_RESET);
```
domjanik, line 41: Why we removed that event ?

jbienkowski-fandom: good catch, I wanted to emit it from IE, but we can leave it here as well

domjanik (dismissed):

szymon-nawrocki (approved):

PR #2146, opened 2023-09-15
TACO-183 remove irrelevant taxonomy tags

jakubbanas:


src/ad-services/lotame/index.ts
```diff
@@ -7,9 +7,20 @@ interface LotameNamespace {
...
+	return result;
+}
+
 export function mapTaxonomyToLotameBehaviorTags(tags: TaxonomyTags = {}): string[] {
-	// bundles are internal tags and should not be sent to Lotame
-	const taxonomyTags = { ...tags, bundles: [] };
+	const taxonomyTags = removeIrrelevantTags(tags);
 
 	return Object.values(taxonomyTags)
 		.map((tag, i) => tag.map((value) => Object.keys(taxonomyTags)[i] + ': ' + value))
```
jbienkowski-fandom, line 26: 1: we could add filter before the map instead of `removeIrrelevantTags` function
2: flatMap instead of map + flat?

jbienkowski-fandom (approved):

domjanik (approved):

PR #2152, opened 2023-09-20
ADEN-13549 delay events after document load completed

preludek:


src/platforms/shared/tracking/load-times-tracker.ts
```diff
@@ -77,18 +86,28 @@ export class LoadTimesTracker {
...
-		);
+		const event = {
+			event_name: eventName,
+			browser_ts: timestamp,
+			load_time: timestamp - this.getStartTime(),
+			tz_offset: this.getTimezoneOffset(),
+			country: utils.geoService.getCountryCode() || '',
+		};
+
+		if (document.readyState !== 'complete') {
```
mfarul, line 107: can we extract a Promise here with a timeout as well ?

preludek: I've sculpted something. Is that something you were thinking of?

src/platforms/shared/tracking/batch-processor.ts
```diff
@@ -0,0 +1,32 @@
...
+
+			return resultArray;
+		}, []);
+	}
+
+	dispatchEventsWithTimeout(sendRequest): void {
+		const batchTimer = setInterval(() => {
+			if (this.batchedTasks.length > 0) {
+				const batchToSend = this.batchedTasks.shift();
+				batchToSend.forEach(({ url, params, type }) => sendRequest(url, params, type));
```
mfarul, line 26: not clear where the task interface is, it should be rather generic with task, task processor (callback) and generic structure for tasks

mfarul (changes requested):

src/platforms/shared/setup/base-context.setup.ts
```diff
@@ -115,6 +115,7 @@ export class BaseContextSetup implements DiProcess {
 			'options.video.comscoreJwpTracking',
 			this.instantConfig.get('icComscoreJwpTracking'),
 		);
+		context.set('options.delayEvents', this.instantConfig.get('icDelayEvents'));
```
adi-wesolowski-fandom, line 118: why not set any default?

src/platforms/shared/tracking/batch-processor.ts
```diff
@@ -0,0 +1,32 @@
...
+
+			resultArray[chunkIndex].push(item);
+
+			return resultArray;
+		}, []);
+	}
+
+	dispatchEventsWithTimeout(sendRequest): void {
+		const batchTimer = setInterval(() => {
+			if (this.batchedTasks.length > 0) {
```
adi-wesolowski-fandom, line 24: quick exit should be more readable, hm?
```suggestion
			if (this.batchedTasks.length <= 0) { clear; return; }
```

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -17,6 +19,24 @@ export type DataWarehouseParams = TrackingParams & TimeBasedParams;
...
+	private eventsArray = [];
+
+	constructor() {
+		this.adEngineStageSetup = new AdEngineStageSetup();
+		this.init();
+	}
+
+	init() {
+		if (context.get('options.delayEvents.enabled')) {
+			document.addEventListener('readystatechange', () => {
```
adi-wesolowski-fandom, line 32: this wont trigger if we would jump after event occur. 
My suggestion: extract code below into function like "dispatchWhenReady" and also check if document isnt complete already, before we set "addEventListener" - if not - set it, if it already done, just run "displathWhenReady" function

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -141,6 +169,22 @@ export class DataWarehouseTracker {
 	/**
 	 * Send the get request
 	 */
+
+	private handleDwEvent(url: string, params: DataWarehouseParams, type = 'Event'): void {
```
adi-wesolowski-fandom, line 173: with suggestion above, this could be simplified

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -141,6 +169,22 @@ export class DataWarehouseTracker {
...
+		if (context.get('options.delayEvents.enabled')) {
```
adi-wesolowski-fandom, line 174: quicker exit would be more appropriate

src/platforms/shared/tracking/data-warehouse.ts
```diff
@@ -141,6 +169,22 @@ export class DataWarehouseTracker {
...
+			this.adEngineStageSetup
+				.afterDocumentCompleted()
+				.then(() => {
+					this.sendRequest(url, params, type);
```
adi-wesolowski-fandom, line 178: 4 level of nesting :(

adi-wesolowski-fandom (changes requested):

src/platforms/shared/setup/base-context.setup.ts
preludek: I just didn't see the point to be honest. This is behind a icVar and has to be enabled anyway, why not just set it there? In case we need to change the numbers quickly we will have to do it through ICBM as well so 🤷

src/platforms/shared/tracking/batch-processor.ts
preludek: 👍

src/platforms/shared/tracking/data-warehouse.ts
preludek: After this event occured we don't need this listener anyway and the events are being sent "the old way". Logic behind it is in `handleDwEvent` method

src/platforms/shared/tracking/data-warehouse.ts
preludek: I mean I could change the order and check if `!context.get('options.delayEvents.enabled')` first and then sendRequest and return but I think this way it is is more readable and straightforward

src/platforms/shared/tracking/batch-processor.ts
preludek: Done

src/platforms/shared/setup/base-context.setup.ts
adi-wesolowski-fandom: because there is always a slight chance, that icVar wont be set in some countries, if we eg. forget to set XX as fallback :) 
But ok, if you dont see a point, no problem

adi-wesolowski-fandom (approved):

mfarul (approved):

PR #2154, opened 2023-09-21
ADEN-13593 Anyclip after first header

verhopara:
[JIRA ➡️ ADEN-13593](https://fandom.atlassian.net/browse/ADEN-13593)
- Changed anyclip anchor point to be rendered after first header.

**Note**:
Since UCP https://github.com/Wikia/unified-platform/blob/master/skins/FandomMobile/scripts/CollapsibleSections.ts#L14 contains the following bug (in my opinion) with collapsible elements. And appends hide/show logic to the `nextSibling` instead of element inside `aria-controls` attribute we can't just add anyclip block `after` header. The workaround which I use is to add it inside content section controlled by header. As side effect hide section = video close + if areas collapsed by default, we not gonna have this ad.

Other options are:
- fix UCP collapsible behaviour, but I believe we not want to do it since it could break a lot of things there
- place ad on top of header, which results in stacked ads

<details>
<summary>📷 Screenshots</summary>
image
image
</details>

### DIFF
--- a/src/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
+++ b/src/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
@@ -283,7 +283,7 @@ export class UcpMobileSlotsDefinitionRepository {
 
 		const slotName = 'incontent_player';
 		const controlledSectionId = document
-			.querySelector('.mw-parser-output > h2:nth-of-type(1)')
+			.querySelector('.mw-parser-output > h2:first-of-type')
 			?.getAttribute('aria-controls');
 
 		return {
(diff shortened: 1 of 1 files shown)


src/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-slots-definition-repository.ts
```diff
@@ -282,14 +282,17 @@ export class UcpMobileSlotsDefinitionRepository {
 		}
 
 		const slotName = 'incontent_player';
+		const controlledSectionId = document
+			.querySelector('.mw-parser-output > h2:nth-of-type(1)')
```
preludek, line 286: We can swap that last for `first-of-type`

preludek (changes requested):

preludek (approved):

PR #2159, opened 2023-09-25
Fix getServicesBaseURL util

jbienkowski-fandom:
Uses [document.currentScript](https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript) property. It will return URL of loader script that has been included in the page.

### DIFF
--- a/src/core/utils/service-base-url.ts
+++ b/src/core/utils/service-base-url.ts
@@ -1,5 +1,19 @@
 export function getServicesBaseURL() {
-	return document.currentScript && 'src' in document.currentScript
-		? new URL(document.currentScript.src).origin + '/'
-		: 'https://services.fandom.com/';
+	const fandomDomains = [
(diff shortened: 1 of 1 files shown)


src/core/utils/service-base-url.ts
```diff
@@ -1,5 +1,15 @@
 export function getServicesBaseURL() {
-	const fandomDomains = ['fandom.com', 'muthead.com', 'futhead.com'];
+	const fandomDomains = [
+		'fandom.com',
+		'tvguide.com',
+		'metacritic.com',
+		'gamespot.com',
+		'giantbomb.com',
+		'muthead.com',
+		'futhead.com',
```
preludek, line 9: futhead.com got closed a month ago so we can remove it

preludek (approved):

domjanik (approved):

PR #2182, opened 2023-10-05
ADEN-13629 More incontent ads (mobile)

lnawrotek:


src/platforms/shared/setup/base-context.setup.ts
```diff
@@ -121,12 +121,18 @@ export class BaseContextSetup implements DiProcess {
...
 
-		if (
-			this.instantConfig.get('icExperiments', []).includes('incontentHeaders') &&
-			!communityExcludedByTag
-		) {
+		const isMobile = context.get('state.isMobile');
+		const isInContentHeadersExperiment = this.instantConfig
+			.get('icExperiments', [])
+			.includes('incontentHeaders');
+		const isDesktopExperiment = !isMobile && !communityExcludedByTag;
```
adi-wesolowski-fandom, line 132: `communityExcludedByTag` is duplicated in both variables and it is not related to desktop/mobile device. Could be moved to the variable above, or (better), put directly in the condition below, like so: 
```
if (isInContentHeadersExperiment && !communityExcludedByTag && (isDesktopExperiment || isMobileExperiment)) {
```

adi-wesolowski-fandom (approved):

adi-wesolowski-fandom (approved):

PR #2184, opened 2023-10-09
Adding support to automatic  changelog generation

houssem-yahiaoui:
In this PR I'm doing the following : 
1 - Adding support to creating a release with automatic change log creation.
2 - Send the automatic change to a slack channel for automatic discovery and centralization process.
3 - The list of labels is changeable so we can agree on a list of labels, I've added support to the one I found beneficial for this repo.
4 - This change log depends heavily on the PR being labelled, so we can also add later on an if check prior to the build start

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,122 @@
+name: Change Log & release generator
+
+on:
+  push:
+    tags:
+      - '*' # <= Action launched with every new tag.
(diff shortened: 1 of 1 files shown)


houssem-yahiaoui:
Prior to merging we need to add a new ENV variable which will be the slack channel incoming webhook link for the automatic notification :).

nandy-andy (dismissed):

PR #2185, opened 2023-10-09
ADEN-13006 - add new bidder: seedtag

adi-wesolowski-fandom:


src/ad-bidders/prebid/adapters/seedtag.ts
```diff
@@ -0,0 +1,38 @@
...
+					sizes,
+				},
+			},
+			bids: [
+				{
+					bidder: this.bidderName,
+					params: {
+						publisherId,
+						adUnitId,
+						placement: placement ?? 'inBanner',
```
preludek, line 32: For consideration, this can be probably set as default 'inBanner' in line 17 and here just pass it as placement.

src/ad-bidders/prebid/prebid-models.ts
```diff
@@ -36,6 +36,7 @@ export interface PrebidAdSlotConfig {
 	parameters?: object;
 	placement?: string;
 	placementId?: string | number;
+	publisherId?: string;
```
preludek, line 39: Order 😈

src/platforms/ucp-desktop/bidders/prebid/seedtag.ts
```diff
@@ -0,0 +1,55 @@
+export function getSeedtagContext(): object {
+	return {
```
preludek, line 2: `publisherId` is the same for all slots, wonder if it can be passed besides the slots key like for example in `getPubmaticContext()`

preludek (approved):
Looks good, some minor comments for consideration

src/platforms/ucp-desktop/bidders/prebid/seedtag.ts
```diff
@@ -0,0 +1,55 @@
...
+			bottom_leaderboard: {
+				sizes: [
+					[728, 90],
+					[970, 250],
+				],
+				publisherId: '7908-0833-01',
+				adUnitId: '29832715',
+				placement: 'inScreen',
+			},
+			floor_adhesion: {
```
preludek, line 44: Those sizes looks like from mobile, also I don't think we run floor_adhesion on desktop except some UAP variant. 🤔

src/ad-bidders/prebid/prebid-models.ts
adi-wesolowski-fandom: 66?

preludek (commented):
There's also a video slot in the ss and we are missing it here.

src/platforms/ucp-desktop/bidders/prebid/seedtag.ts
adi-wesolowski-fandom: agree, but in spreadsheet was connected as desktop, and we also have that slot in desktop context setup: src/platforms/ucp-desktop/setup/context/slots/ucp-desktop-slots-context.setup.ts (however with desktop/TLB size).

I will remove it from here, just for now.

src/platforms/ucp-desktop/bidders/prebid/seedtag.ts
adi-wesolowski-fandom: yes, it is the same within one page. For now. 
It would be great, to give possibility to put in eg alongside "enabled" flag. But right here, right now, there is no support for it, as far as I can see. But it is good idea, to add such feature. There is another bidder, with same issue.
For now, I think its good to keep it here: to make it clear and readable. Putting everything in this global "context" variable isnt good way in my opinion, especially here, make another path to passing arguments. It may be confusing, making such exception here, for this particular bidder.

src/ad-bidders/prebid/adapters/seedtag.ts
```diff
@@ -0,0 +1,73 @@
...
+				},
+			],
+		};
+	}
+
+	getVideoConfig(code, publisherId, adUnitId): PrebidAdUnit {
+		return {
+			code,
+			mediaTypes: {
+				video: {
```
preludek, line 48: I'd leave only required parameters here

src/platforms/ucp-desktop/bidders/prebid/seedtag.ts
```diff
@@ -0,0 +1,51 @@
...
+				placement: 'inBanner',
+			},
+			incontent_leaderboard: {
+				sizes: [[728, 90]],
+				publisherId: '7908-0833-01',
+				adUnitId: '29832784',
+				placement: 'inBanner',
+			},
+			incontent_boxad: {
+				sizes: [[300, 50]],
```
preludek, line 30:
```suggestion
				sizes: [[300, 250]],
```

src/ad-bidders/prebid/adapters/seedtag.ts
```diff
@@ -0,0 +1,73 @@
...
+					sizes,
+				},
+			},
+			bids: [
+				{
+					bidder: this.bidderName,
+					params: {
+						publisherId,
+						adUnitId,
+						placement: placement,
```
preludek, line 37:
```suggestion
						placement,
```

preludek (changes requested):
Some minor potential improvements + a size change for one slot

src/ad-bidders/prebid/adapters/seedtag.ts
adi-wesolowski-fandom: I used same as for pubmatic which was marked as supported in seedtag docs, to keep consistency, like eg. max time duration of the ad.

preludek (approved):

src/ad-bidders/prebid/adapters/seedtag.ts
```diff
@@ -0,0 +1,73 @@
...
+			],
+		};
+	}
+
+	getVideoConfig(code, publisherId, adUnitId): PrebidAdUnit {
+		return {
+			code,
+			mediaTypes: {
+				video: {
+					playerSize: [640, 480],
```
adi-wesolowski-fandom, line 49:
```suggestion
					playerSize: sizes[0] ?? [640, 480],
```

PR #2186, opened 2023-10-09
TACO-75 Add experian pixel

bckatarzyna:


src/ad-services/experian/index.ts
```diff
@@ -0,0 +1,29 @@
...
+			utils.logger(logGroup, 'disabled');
+			return;
+		}
+		this.insertExperianPixel();
+	}
+
+	private getExperianPixelUrl(): string {
+		const ppid = targetingService.get('ppid');
+
+		return `https://pixel.tapad.com/idsync/ex/receive
```
domjanik, line 20: Maybe we could move it to separate const like we did in liveRampPixel ?

src/ad-services/experian/index.ts
```diff
@@ -0,0 +1,29 @@
...
+		?partner_id=${this.PARTNER_ID}&partner_device_id=${ppid}`;
+	}
+
+	insertExperianPixel(): void {
```
domjanik, line 24: I'm wondering if we could create a separate method to load pixels. Currently, we have 2 separate pixels: LiveRampPixel and ExperianPixel. Creating common method to load pixels allows us to re-use that part of code in the future and ensures similar logic. On the other hand - when we'll resume our work on partner migration, Experian will be one of the first and most important partners to be moved out of the AdEng and it would be useless, therefore I'm leaving this up to you, if we should create such method.

domjanik (approved):

src/ad-services/experian/index.ts
bckatarzyna: ✅

domjanik (approved):

PR #2189, opened 2023-10-11
ADEN-13543 - move partners at the end, right after event: TLB rendered

adi-wesolowski-fandom:


src/platforms/shared/setup/partners.setup.ts
```diff
@@ -0,0 +1,73 @@
...
+			AdSlotEvent.SLOT_RENDERED_EVENT,
+			() => {
+				utils.logger('ad-stack-partners', 'starting');
+				this.pipelineExecute();
+			},
+			this.firstCallSlotName,
+		);
+	}
+
+	pipelineExecute(): void {
```
lnawrotek, line 51: I think it should be private method

src/platforms/shared/setup/partners.setup.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	execute(): void {
+		utils.logger('ad-stack-partners', 'waiting ...');
+		setTimeout(() => {
+			this.pipelineExecute();
+		}, this.safeTimeout);
+		communicationService.onSlotEvent(
+			AdSlotEvent.SLOT_RENDERED_EVENT,
+			() => {
+				utils.logger('ad-stack-partners', 'starting');
```
lnawrotek, line 44: Please add a new variable `logGroup` and replace all `ad-stack-partners`

src/platforms/shared/setup/partners.setup.ts
```diff
@@ -0,0 +1,73 @@
...
+	LiveRampPixel,
+	Lotame,
+	Nielsen,
+	PartnerPipeline,
+	utils,
+} from '@wikia/ad-engine';
+import { Injectable } from '@wikia/dependency-injection';
+
+@Injectable()
+export class PartnersSetup implements DiProcess {
```
lnawrotek, line 19:
```suggestion
export class UcpPartnersSetup implements DiProcess {
```

This file is in "shared" directory, but this class won't be used on N&R platforms because these sites don't load all partners

lnawrotek:
should we change N&R sites too?

src/platforms/shared/setup/partners.setup.ts
adi-wesolowski-fandom: ok

src/platforms/shared/setup/partners.setup.ts
adi-wesolowski-fandom: sure

src/platforms/shared/setup/partners.setup.ts
adi-wesolowski-fandom: ok, thx

jakubbanas (approved):

adi-wesolowski-fandom:
> should we change N&R sites too?

not at the moment. Its for fandom only, for testing is it good approach

PR #2190, opened 2023-10-12
TACO-212 add iiq start/done events

jakubbanas:


src/platforms/shared/tracking/load-times-tracker.ts
```diff
@@ -19,6 +19,8 @@ const eventsToTrack = {
 	eyeota_failed: eventsRepository.EYEOTA_FAILED,
 	a9_without_consents: eventsRepository.A9_WITHOUT_CONSENTS,
 	intentiq_ppid_not_set_on_time: eventsRepository.INTENTIQ_PPID_NOT_SET_ON_TIME,
+	intentiq_start: eventsRepository.INTENTIQ_START,
+	intentiq_done: eventsRepository.INTENTIQ_DONE,
 	lotame_loaded: eventsRepository.LOTAME_LOADED,
```
domjanik, line 24: I think you should also add ID5 events there 🤔

domjanik (approved):

src/platforms/shared/tracking/load-times-tracker.ts
jakubbanas: Right, added, thanks!

domjanik (approved):

PR #2191, opened 2023-10-12
COTECH-713 | Add Optimizely experiments to targeting

latata:
we want to pass information about the active Optimizely experiment to GAM. As the experiments sometimes are implemented in different code bases we want to have the targeting set independently to avoid introducing unnecessary communication between the adengine and the experiment code just to inform that the experiment is active.

### DIFF
--- a/src/platforms/shared/context/targeting/ucp-targeting.setup.ts
+++ b/src/platforms/shared/context/targeting/ucp-targeting.setup.ts
@@ -9,6 +9,7 @@ import {
 	utils,
 } from '@wikia/ad-engine';
 import { Injectable } from '@wikia/dependency-injection';
+import { getOptimizelyActiveVariations } from '../../utils/optimizely';
 import { createFandomContext } from './targeting-strategies/factories/create-fandom-context';
 import { createOpenRtb2Context } from './targeting-strategies/factories/create-open-rtb2-context';
 import { createSelectedStrategy } from './targeting-strategies/factories/create-selected-strategy';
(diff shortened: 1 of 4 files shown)


nandy-andy (approved):

src/platforms/shared/context/targeting/ucp-targeting.setup.ts
```diff
@@ -54,6 +55,18 @@ export class UcpTargetingSetup implements DiProcess {
 			utils.targeting.getTargetingBundles(this.instantConfig.get('icTargetingBundles')),
 		);
 
+		const optimizelyVariations = getOptimizelyActiveVariations();
+
+		if (optimizelyVariations) {
+			const optimizelyTargeting = Object.keys(optimizelyVariations).map((experimentId) => {
+				return `${experimentId}_${optimizelyVariations[experimentId].id}`;
+			});
```
andrzejkupczyk, line 63: Have you taken into account the option of integrating this transformation into `getOptimizelyActiveVariations` or any other suitable composable util instead of extending `UcpTargetingSetup` responsibility?

latata: makes sense. done.

andrzejkupczyk (approved):

src/platforms/shared/context/targeting/ucp-targeting.setup.ts
andrzejkupczyk: Awesome, thanks!

PR #2194, opened 2023-10-16
Aden 13633 Remove search ads from Escape from Tarkov

lnawrotek:


src/ad-services/system1/index.ts
```diff
@@ -166,4 +160,26 @@ export class System1 extends BaseServiceSetup {
 
 		return blockedBotUserAgents.some((botUserAgent) => userAgent.includes(botUserAgent));
 	}
+
+	private isDisabled(): boolean {
```
Jorewin, line 164: Now this class has two methods `isEnabled` and `isDisabled` and their results aren't a negation of each other.
I think that:

- either a name change of the `isDisabled` method needs to happen or
- `isEnabled` should get overridden in this class and then it's negation should be used.

Jorewin (approved):

PR #2201, opened 2023-10-17
COTECH-787 Refresh bids after Anyclip ad displayed

nandy-andy:
This PR includes two main changes to our Anyclip integration:
1. on `adImpression` event from Anyclip we refresh bids and update `incontent_player` targeting,
3. we don't send `featured` bids anymore on pages without JWPlayer and we don't send `incontent_player` bids anymore on pages with JWPlayer.


src/ad-services/anyclip/anyclip-bids-refresher.ts
```diff
@@ -0,0 +1,31 @@
...
+		}
+	}
+
+	private onAdImpressionHandler() {
+		utils.logger(logGroup, 'AnyclipBidsRefresher: ad impression in Anyclip detected - refreshing');
+
+		const slotName = 'incontent_player';
+		const playerAdSlot = slotService.get(slotName);
+
+		playerAdSlot.emit(AdSlotEvent.VIDEO_AD_IMPRESSION); // removes video bids
```
dzawada, line 27: Is there a chance that emitting this event will create "false" impressions in our monitoring?

src/platforms/shared/bidders/filter-video-bids.ts
```diff
@@ -0,0 +1,23 @@
...
+	const hasFeaturedVideo = context.get('custom.hasFeaturedVideo');
+	const bidConfigSlotNames = Object.keys(bidderContext.slots);
+	const hasFeaturedBidConfig = bidderContext.slots && bidConfigSlotNames.includes('featured');
+	const hasIncontentPlayerBidConfig =
+		bidderContext.slots && bidConfigSlotNames.includes('incontent_player');
+	const newVideoSlotsConfig = {
+		...bidderContext.slots,
+	};
+
+	if (hasFeaturedVideo && hasIncontentPlayerBidConfig) {
```
dzawada, line 13: There are cases here where newVideoSlotsConfig will contain both 'featured' and 'incontent_player' configs. Is that expected?

nandy-andy: What cases do you have in mind? 🤔 I though the `if` and then `else if` made sure there wouldn't be such cases at the end 😊

dzawada: Ok, so let's consider following scenario:
hasFeaturedVideo = false
hasIncontentPlayerBidConfig = true
hasFeaturedBidConfig = false
(P.S. I have no idea if it is in fact possible)
Then you won't enter if due to hasFeaturedVideo being false and you also won't enter else if since hasFeaturedBidConfig is false, causing no entry in newVideoSlotsConfig to be removed.

nandy-andy: Got it. Well, we can definitely think about improving this piece of code especially if we introduce more slots/bids configuration in the future. But for current setup where most of the bid configs has `featured` and additionally some have also `incontent_player` and we want to keep only one. This if-else-if should be enough for now as the situation you described means: there is no JWP on the page (!hasFeaturedVideo) and there is only bid config for `incontent_player` which is fine -- we don't need to remove anything from the bid video configs 👍

spec/platforms/ucp-desktop/setup/prebid/ucp-desktop-prebid-config.setup.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { context } from '@wikia/core';
+import { UcpDesktopPrebidConfigSetup } from '@wikia/platforms/ucp-desktop/setup/context/prebid/ucp-desktop-prebid-config.setup';
+import { getVideoBiddersWithVideoSlots } from '../../../../helpers/get-video-bidders-with-video-slots';
```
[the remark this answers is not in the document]
nandy-andy: It should be safe. Not many spec files use stuff from the helpers directory but if they do they do it with a relative path ;)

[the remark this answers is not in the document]
andrzejkupczyk: Oh man, my bad. I must have overlooked that it's spec... 🤦🏻

andrzejkupczyk (approved):

dzawada (approved):

PR #2209, opened 2023-10-20
COTECH-800 Bring back Connatix player

nandy-andy:
https://fandom.atlassian.net/browse/COTECH-800


src/ad-services/connatix/player-injector.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export class PlayerInjector implements PlayerInjectorInterface {
+	get playerId(): string {
+		return context.get('services.connatix.playerId');
+	}
+
+	get renderId(): string {
+		return context.get('services.connatix.renderId');
+	}
```
andrzejkupczyk, line 16: I'm unsure about coupling `PlayerInjector` with the ad engine context. It seems like the `Connatix` setup service should be the sole class aware of the service configuration and responsible for providing it consistently. This service already passes the `cid` value during player initialization and inserts it into the DOM. I'm curious why the injector fetches other parameters independently 🤔

spec/ad-services/connatix/connatix.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+			(event: EventOptions, callback: (payload?: any) => void) => {
+				const payload = { isLoaded: true };
+				callback(payload);
+			},
+		);
+
+		sinon.assert.notCalled(playerInjectorSpy.insertPlayerContainer);
+	});
+
+	it('is called when disabled', async () => {
```
andrzejkupczyk, line 53: Please disregard this comment if I'm mistaken, but it appears that this test only covers the happy path, as there's a possibility that `services.connatix.enabled` may be undefined for mobile UCP 🤔

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	constructor(
+		protected instantConfig,
+		protected globalTimeout,
+		private playerInjector: PlayerInjectorInterface,
+	) {
+		super();
+
+		const cnxPlayer = new ConnatixPlayer();
```
andrzejkupczyk, line 25: Could we please avoid using acronyms like this to prevent any confusion? 🙏🏻

```suggestion
		const connatixPlayer = new ConnatixPlayer();
```

src/ad-services/connatix/index.ts
```diff
@@ -0,0 +1,44 @@
...
+			utils.logger(logGroup, 'Connatix player is disabled');
+			return Promise.resolve();
+		}
+
+		communicationService.on(eventsRepository.AD_ENGINE_UAP_LOAD_STATUS, (action: UapLoadStatus) => {
+			if (!action.isLoaded) {
+				utils.logger(logGroup, 'No Fan Takeover loaded - injecting Connatix player');
+				initConnatixHeadScript(this.cid);
+				utils.logger(logGroup, 'Connatix head script is ready');
+				this.playerInjector.insertPlayerContainer(this.cid);
```
andrzejkupczyk, line 40: The `insertPlayerContainer` function expects `cid` to be a string, but `this.cid` can have any other value or be undefined since there's no validation. Should we consider adding an additional check to the `if (!this.isEnabled())` condition?

andrzejkupczyk (approved):

src/ad-services/connatix/player-injector.ts
nandy-andy: I believe since `PlayerInjector` is placed in the `src/ad-services/connatix` directory it's more like a `ConnatixPlayerInjector` but the author didn't like long names of classes, I guess. We could extract it to some shared directory but I don't recall any other players we've integrated with needed `cid`, `renderId` or any similar parameters like the Connatix one. Maybe we should just rename the class, to avoid confusion 🤔

andrzejkupczyk: Ah, I forgot that you're not the original author, sorry 'bout that. Let's just move on then. We can make further enhancements on the next iterations 👍🏻

nandy-andy: Well, I've started some of the changes but I'll leave the naming of `PlayerInjector` and `PlayerInjectorInterface` then 😊

andrzejkupczyk (approved):

PR #2214, opened 2023-10-24
TACO-205 - Changed getTCData method into addEventListener

domjanik:


src/core/wrappers/tcf.ts
```diff
@@ -1,13 +1,15 @@
 export class Tcf {
-	defaultVersion = 2;
+	defaultVersion = 0;
```
jbienkowski-fandom, line 2: why `0`?

jbienkowski-fandom (approved):

PR #2215, opened 2023-10-24
TACO-229 - Fixed problems with params not being cleared

domjanik:


src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -55,12 +55,17 @@ export class TrackingParametersSetup implements DiProcess {
 			: await this.getNewTrackingParameters();
 	}
 
-	async execute() {
+	async setTrackingParameters() {
```
jbienkowski-fandom, line 58: why async?

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -55,12 +55,17 @@ export class TrackingParametersSetup implements DiProcess {
...
+		console.log('DJ: setTrackingParameters');
 		const legacyEnabled = !this.instantConfig.get('icDisableLegacyTrackingParameters', false);
 		const trackingParameters = await this.getTrackingParameters(legacyEnabled);
 
 		context.set('wiki', { ...context.get('wiki'), ...trackingParameters });
+	}
 
+	async execute() {
+		console.log('DJ: execute');
```
szymon-nawrocki, line 67:
```suggestion
```

src/platforms/shared/setup/tracking-parameters.setup.ts
domjanik: 'cause we need to wait for getTrackingParameters, in which we might need to wait for getNewTrackingParameters ;)

szymon-nawrocki (approved):

jbienkowski-fandom (approved):

PR #2163, opened 2023-09-25
TACO-176 create amazon token from HEM

jakubbanas:
### JIRA
[TACO-176](https://fandom.atlassian.net/browse/TACO-176)

### Description

Implements integration with Apstag `window.apstag.rpa` method. This method collects user HEM (provided by MediaWiki or LiveConnect) and provides Amazon Token that can be later on used by Apstag and A9 bidder.

[TACO-176]: https://fandom.atlassian.net/browse/TACO-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2217, opened 2023-10-24
ADEN-13681 TACO-176 Create Amazon Token from HEM

jbienkowski-fandom:
### JIRA
[ADEN-13681](https://fandom.atlassian.net/browse/ADEN-13681)
[TACO-176](https://fandom.atlassian.net/browse/TACO-176)

### Description

Implements integration with Apstag `window.apstag.rpa` method. This method collects user HEM (provided by MediaWiki or LiveConnect) and provides Amazon Token that can be later on used by Apstag and A9 bidder.
This PR fixes previous problems introduced in #2163 which were caused by race condition... Before it could be a problem as well as we were not waiting for `apstag.init`, but in majority of cases it was working as expected. After some moving of code pieces the race condition was more easy to obtain.

[Fixed code diff vs previous implementation](https://github.com/Wikia/ad-engine/compare/v146.2.0...ADEN-13681-fix#diff-8e4e1901f435f31bd531736bed3ed6c16b4449e151d3330f823b99814294f637)

[ADEN-13681]: https://fandom.atlassian.net/browse/ADEN-13681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TACO-176]: https://fandom.atlassian.net/browse/TACO-176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ad-bidders/a9/index.ts
+++ b/src/ad-bidders/a9/index.ts
@@ -68,7 +68,6 @@ export class A9Provider extends BidderProvider {
 		this.slots = this.bidderConfig.slots;
 		this.slotsNames = Object.keys(this.slots);
 		this.bidsRefreshing = context.get('bidders.a9.bidsRefreshing') || {};
-		this.apstag.init();
 	}
 
 	getTargetingKeys(): string[] {
--- a/src/ad-bidders/a9/index.ts
+++ b/src/ad-bidders/a9/index.ts
@@ -68,13 +68,14 @@ export class A9Provider extends BidderProvider {
 		this.slots = this.bidderConfig.slots;
 		this.slotsNames = Object.keys(this.slots);
 		this.bidsRefreshing = context.get('bidders.a9.bidsRefreshing') || {};
+		this.apstag.init();
 	}
 
 	getTargetingKeys(): string[] {
 		return this.targetingKeys;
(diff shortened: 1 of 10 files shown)


src/ad-bidders/a9/index.ts
```diff
@@ -374,14 +326,12 @@ export class A9Provider extends BidderProvider {
...
-		if (this.usp.exists) {
-			signalData = await this.usp.getSignalData();
-		}
+	protected callBids(): void {
+		this.bids = {};
+		this.priceMap = {};
+		const a9Slots = this.getA9SlotsDefinitions(this.slotsNames);
 
-		this.init(signalData);
+		this.initIfNotLoaded().then(() => this.fetchBids(a9Slots));
```
jbienkowski-fandom, line 334: we cannot call `fetchBids` before `initIfNotLoaded` is resolved

domjanik (approved):

PR #2219, opened 2023-10-25
ADEN-13625: Fix the returned ad unit paths not being mapped to ad slot names

Jorewin:


src/ad-services/double-verify/index.ts
```diff
@@ -71,7 +77,12 @@ export class DoubleVerify extends BaseServiceSetup {
 
 	private addToSlotsTargeting(data: any, targetingKey: string): void {
 		if (typeof data === 'object') {
-			Object.entries(data).forEach(([slotName, value]) => {
+			Object.entries(data).forEach(([adUnitPath, value]) => {
+				const adUnit: AdUnit | undefined = this.adUnits.find(
```
lnawrotek, line 81:
```suggestion
				const adUnit: AdUnit | undefined = this.adUnits.find(adUnit => adUnit.adUnitPath === adUnitPath);
```

src/ad-services/double-verify/index.ts
```diff
@@ -71,7 +77,12 @@ export class DoubleVerify extends BaseServiceSetup {
...
+					({ adUnitPath: path }) => adUnitPath === path,
+				);
+				const slotName = adUnit?.slotName ?? '';
```
lnawrotek, line 84: I think we should skip it if we don't have a name

lnawrotek (approved):

PR #2224, opened 2023-10-31
ADEN-13661-v146-2-9 remove gpt from double verify dependency

lnawrotek:


lnawrotek:
### Don't merge dev to this branch!

PR #2228, opened 2023-11-09
MAIN-29031 | Avoid infinite recursion when resetting IC cache

mszabo-wikia:
Infinite recursion may occur via resetCache() ->
Interpreter::getValues() -> SamplingCache when setting IC cache values. As a workaround, add a recursion guard till we figure out a proper fix.


mszabo-wikia:
Likely flow: `InstantConfigCacheStorage::set` -> `InstantConfigCacheStorage::resetCache` -> `InstantConfigInterpreter::getValues` -> `SamplingCacheManager::apply` -> `InstantConfigCacheStorage::set`.

jakubbanas (approved):

latata (approved):

src/core/services/instant-config-cache-storage.ts
```diff
@@ -35,7 +35,11 @@ export class InstantConfigCacheStorage implements InstantConfigCacheStorageServi
 	private readonly cacheKey = 'basset';
 
 	private constructor() {
-		this.resetCache();
+		communicationService.on(eventsRepository.AD_ENGINE_CONSENT_READY, ({ gdprConsent }) => {
```
mszabo-wikia, line 38: I think it's worth adding a comment around here, or on resetCache(), to be mindful of potential recursion like the one we're eliminating here, for the benefit of our future selves.

jbienkowski-fandom:
Part 2: https://github.com/Wikia/identity-engine/pull/231

src/core/services/instant-config-cache-storage.ts
jbienkowski-fandom: The whole flow got refactored a bit more to mitigate a chance of recursive issue as now we have a separate PostQuecast event for that

mfarul (approved):

PR #2229, opened 2023-11-09
TACO-178 TACO-177 Update and Remove Amazon Token

jbienkowski-fandom:
## JIRA

https://fandom.atlassian.net/browse/TACO-178
https://fandom.atlassian.net/browse/TACO-177

## Description

In this PR a several important changes are being introduced:

1. First of all, new event `[AdEngine OptIn] update opt in` is being added. It is being emit only when consent has already been provided in 1 PV and it has been changed. Old event `[AdEngine OptIn] set opt in` is being sent as well for backwards compatibility.
Why new event? Because relying on `[AdEngine OptIn] set opt in` is more difficult then it looks... On Fandom, we receive 2 such events during stack initialization (1 from TrackingOptIn and 1 from IdentityEngine) while on other platforms we receive 1 event. On top of that, in most of the places we rely on the result of user consent for the whole PV. This is safe on Fandom, where consent change requires separate PVs, but on SPA platforms it is not a perfect solution and it could work better. These changes could help with that in the future. 
2. Implement missing `apstag.upa` method for updating user consent after token has been assigned. This requires some more data being stored in LocalStorage as according to Amazon documentation to properly update user consent we need to provide the HEM used for creating one. This one is being controlled by the same `icA9HEM` ICBM flag.
3. Implement missing `apstag.dpa` method for cleaning up Amazon Tokens. This method should be used when we decide to not use Amazon Tokens anymore and we want to clean all Amazon Tokens assigned to our users. This one is being controlled by the separate `icA9CleanupHEM` ICBM flag.

## TODOs
- [x] Unit tests
- [x] Small refactor of `sendHEM` method as there is a lot going on there atm


src/ad-bidders/wrappers/apstag.ts
```diff
@@ -63,17 +77,52 @@ export class Apstag {
...
+	public async sendHEM(
+		record: string,
+		consents?: GdprConsentPayload & CcpaSignalPayload,
+	): Promise<void> {
+		if (!record) {
+			utils.warner(logGroup, 'Trying to send HEM without source record');
+			return;
+		}
+
+		// TACO-177
```
domjanik, line 89: Do we want to leave that comment? If you want to know more context you can always use git blame

domjanik (approved):

domjanik (dismissed):

domjanik (approved):

domjanik (dismissed):

domjanik (approved):

domjanik (approved):

PR #2246, opened 2023-11-17
COTECH-819 Bring back Connatix to mobile and make it work like Anyclip

nandy-andy:
We bring back the Connatix player on FCP mobile behind `icConnatixPlayer` flag:
* the player behaves like Anyclip player (does not appear on pages with Fan Takeover, has disabled floating when floor_adhesion displays on a page),
* there is a bid refresher logic hooked with the `adImpression` event from the player,
* additionally the player is covered with a minimal internal tracking.


src/ad-services/connatix/index.ts
```diff
@@ -32,13 +50,57 @@ export class Connatix extends BaseServiceSetup {
...
+		utils.logger(logGroup, 'initialized', this.cid);
+
+		communicationService.on(
+			eventsRepository.AD_ENGINE_UAP_LOAD_STATUS,
+			this.loadOnUapStatus.bind(this),
+		);
+	}
+
+	private loadOnUapStatus({ isLoaded, adProduct }: UapLoadStatus) {
+		if (!isLoaded && adProduct !== 'ruap') {
```
andrzejkupczyk, line 62: Instead, I'd suggest reverse this condition and return early to enhance readability

andrzejkupczyk (approved):

src/ad-services/connatix/index.ts
nandy-andy: Yeah, let's leave it as-is so it matches the logic in Anyclip. However, it's fishy - I've reached out to Seba to find out the business use-case.

andrzejkupczyk: It was not about changing the logic. This is what I meant:

```suggestion
        if (isLoaded || adProduct === 'ruap') {
            utils.logger(logGroup, 'Connatix blocked because of Fan Takeover');
            return;
        }
        // rest of the code
```

andrzejkupczyk: Upon reversing the condition, it becomes apparent that the current log message lacks clarity.

PR #2250, opened 2023-11-21
TACO-274 Simplify ID5 control group logic

szymon-nawrocki:
### Description
Turns out that when ID5 `uid` is equal to 0, it means that the user is in the A/B testing Control Group.
With this information there is no more need for saving control group in local storage, so this logic can be removed

### DIFF
--- a/src/ad-bidders/prebid/id5/id5.ts
+++ b/src/ad-bidders/prebid/id5/id5.ts
@@ -1,5 +1,5 @@
 import { communicationService, eventsRepository } from '@ad-engine/communication';
-import { context, targetingService, UniversalStorage, utils } from '@ad-engine/core';
+import { context, targetingService, utils } from '@ad-engine/core';
 import { UserIdConfig } from '../index';
 
 interface Id5Config extends UserIdConfig {
@@ -16,12 +16,7 @@ const logGroup = 'Id5';
 
 class Id5 {
 	private partnerId = 1139;
-	private storage;
 	private id5GroupKey = 'id5_group';
-
-	constructor() {
-		this.storage = new UniversalStorage();
-	}
 
 	private isEnabled(): boolean {
 		return (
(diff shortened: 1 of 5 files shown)


domjanik (approved):

domjanik (approved):

src/ad-bidders/prebid/id5/id5.ts
```diff
@@ -73,52 +68,48 @@ class Id5 {
...
 	private setTargeting(key: string, value: string): void {
 		targetingService.set(key, value);
 		utils.logger(logGroup, 'set targeting', key, value);
 	}
+
+	public enableAnalytics(pbjs: Pbjs): void {
+		if (context.get('bidders.prebid.id5Analytics.enabled')) {
+			utils.logger(logGroup, 'enabling ID5 Analytics');
+
+			pbjs.que.push(() => {
```
jbienkowski-fandom, line 103: we could use `pbjsFactory` here:
```
pbjsFactory.init().then((pbjs) => pbjs.enableAnalytics({
					provider: 'id5Analytics',
					options: {
						partnerId: this.partnerId,
					},
				})
);
```

szymon-nawrocki: Good question, looking at the official Prebid documentation, for some of the Analytics adapters `enableAnalytics` is called directly on the `pbjs` object, for some `enableAnalytics` is pushed to que. (https://docs.prebid.org/overview/analytics.html)

But it looks like ID5 can be called directly on pbjs object, and it is done the same in the ID5 documentation: https://wiki.id5.io/identitycloud/retrieve-id5-ids/prebid-user-id-module/id5-prebid-analytics-module

szymon-nawrocki: When I was implementing ID5 Analytics I looked at ATS Analytics module which is pushed to `pbjs.que`: https://github.com/Wikia/ad-engine/blob/30de4d7c75d97558d4ab5fb52f5207c3ed83c723/src/ad-bidders/prebid/index.ts#L274

And I did the same for ID5.
But I can try to get rid of pushing it to the `pbjs.que` and call it directly with `pbjs.enableAnalytics({ provider: 'id5Analytics' })`

jbienkowski-fandom: It is a minor issue, you can give it a go or just ignore it ;) I assume we have more places like that... Maybe we should try and normalize them sometime...

szymon-nawrocki: Done ✅  Thanks for the suggestion

domjanik (approved):

PR #2258, opened 2023-11-23
Aden 13619 2

adi-wesolowski-fandom:


src/ad-bidders/prebid/liveramp-ats/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import { UserIdConfig } from '../index';
+
+const logGroup = 'LiveRamp';
+
+export abstract class LiveRampAtsTypes {
+	static PLACEMENT_ID = '2161';
+	static ENVELOPE_STORAGE_NAME = 'idl_env';
+}
+
+class LiveRampAts {
```
mfarul, line 11: It is rather RampID setup for prebid. ATS is build on it.

src/ad-services/ats/index.ts
```diff
@@ -0,0 +1,59 @@
...
+			utils.logger(logGroup, 'no hashes');
+			return;
+		}
+
+		const [md5, sha1, sha256] = userEmailHashes;
+		const launchpadScript = utils.scriptLoader.loadScript(this.launchpadScriptUrl);
+		const launchpadBundleScript = utils.scriptLoader.loadScript(this.launchpadBundleScriptUrl);
+
+		return Promise.all([launchpadScript, launchpadBundleScript])
+			.then(() => this.waitForAts())
```
mfarul, line 37: is this needed? then should be fired after the script is loaded & executed, so the variable should be there...

src/ad-bidders/prebid/index.ts
```diff
@@ -284,6 +295,23 @@ export class PrebidProvider extends BidderProvider {
 		}
 	}
 
+	private enableATSAnalytics(): void {
+		if (context.get('bidders.liveRampATSAnalytics.enabled')) {
```
mfarul, line 299: we should also check if liveRampId is enabled, as analytics probably will not work without prebid module

mfarul (approved):

src/ad-bidders/prebid/liveramp-ats/index.ts
preludek: renamed

src/ad-bidders/prebid/index.ts
preludek: done

src/ad-services/ats/index.ts
preludek: I removed it and scripts are being loaded so I guess it's redundant. Removing it

preludek (approved):

PR #2259, opened 2023-11-27
ADEN-12076 | disable ads for logged in users of gamepedia wikis

mklakus-fandom:
### Description
- using `?adengine_version=local` query param both on `fandomdesktop` and `fandommobile` skins when user is logged in on `gamepedia` wikis should result with displaying no ads, 
- nothing should be changed for not logged in users.


mfarul (changes requested):
I would rather unify FCP code to send no-ads app instead of checking this on AdEngine side.

PR #2267, opened 2023-12-05
ADEN-13712 | Measure all pipeline steps and partners

jakubbanas:


preludek (dismissed):
Overall looks good and like it's working 😄

During refinement I was thinking about emitting events directly from ScriptLoader or maybe a seperate class that would use script loader therefore we could just pass a string in all the partners instead of emiting events from those classes. What do you think about that solution?

jakubbanas:
> Overall looks good and like it's working 😄
> 
> During refinement I was thinking about emitting events directly from ScriptLoader or maybe a seperate class that would use script loader therefore we could just pass a string in all the partners instead of emiting events from those classes. What do you think about that solution?

Good idea, thanks!

preludek (approved):

src/platforms/shared/tracking/load-times-tracker.ts
```diff
@@ -83,6 +69,10 @@ export class LoadTimesTracker {
 				this.trackLoadTime('top_leaderboard_loaded', Date.now());
 			}
 		});
+
+		communicationService.on(eventsRepository.PARTNER_LOAD_STATUS, ({ status }) => {
+			this.trackLoadTime(status, Date.now());
+		});
```
mfarul, line 75:
```suggestion
		}, false);
```

mfarul (changes requested):

mfarul (approved):

PR #2269, opened 2023-12-05
ADEN-13438 Quick-fix for bottom ads being too close to each other

nandy-andy:
It changes the bottom ads on desktop so instead of displaying like this:
Screenshot 2023-12-05 at 17 11 32
They display with a spacing like this:
Screenshot 2023-12-05 at 17 17 10

That's a quick-fix before holidays ;-) A follow-up would be implementing it in Gamespot repository.


nandy-andy:
Moved to Gamespot: https://github.com/Wikia/gamespot/pull/4106

PR #2272, opened 2023-12-08
Inconsistent ad display on first page view for feature video

mfarul:


src/platforms/shared/setup/player.setup.ts
```diff
@@ -19,23 +21,33 @@ export class PlayerSetup extends BaseServiceSetup {
...
+	async call() {
 		const showAds = !context.get('options.wad.blocking');
-		const vastXml = utils.displayAndVideoAdsSyncContext.getVideoVastXml();
 
 		if (showAds) {
 			utils.logger(logGroup, 'JWP with ads controlled by AdEngine enabled');
+			const vastResponse: VastResponseData =
+				utils.displayAndVideoAdsSyncContext.isTaglessRequestEnabled()
+					? await this.vastTaglessRequest.getVast(
+							context.get('options.video.vastRequestTimeout') | 500,
```
nandy-andy, line 38: I'm afraid we need one more `|` before `500`

nandy-andy (approved):

dzawada (dismissed):

spec/platforms/shared/setup/player.setup.spec.ts
```diff
@@ -79,7 +93,13 @@ describe('PlayerSetup', () => {
...
 
 		dispatch = global.sandbox.spy(communicationService, 'dispatch');
-		const playerSetup = new PlayerSetup(instantConfigStub, null, new Optimizely(), jwpManagerStub);
+		const playerSetup = new PlayerSetup(
+			instantConfigStub,
+			null,
+			new Optimizely(),
+			jwpManagerStub,
+			global.sandbox.createStubInstance(VastTaglessRequest),
+		);
```
andrzejkupczyk, line 102: To be honest, it might be easy to lose track of what's covered by the test, especially regarding the target. We're initializing PlayerSetup, Optimizely, and creating a VastTaglessRequest sub-instance within the "act" phase. But in reality, we're testing the behavior related to calling the playerSetup, aren't we?

It's not a blocker, though. Just something to consider for enhancing readability in the future 👍🏻

andrzejkupczyk (dismissed):

spec/platforms/shared/setup/player.setup.spec.ts
mfarul: simplifed

andrzejkupczyk (approved):

PR #2273, opened 2023-12-11
COTECH-926, COTECH-921 | Add new video event for Connatix player

nandy-andy:


andrzejkupczyk (dismissed):

dzawada (approved):

spec/platforms/shared/setup/player.setup.spec.ts
```diff
@@ -7,6 +7,8 @@ import { expect } from 'chai';
 import { SinonSpy } from 'sinon';
 
 describe('PlayerSetup', () => {
+	const MOCKED_VAST_AD_UNIT = '/5441/test/vast/ad/unit';
```
andrzejkupczyk, line 10: It's not a mock of any kind; it's simply a test value 😉 A mock is an object (a type of test double) that mimics the behavior of real objects in a controlled way. I'd recommend choosing another name to prevent any confusion.

andrzejkupczyk (approved):

spec/platforms/shared/setup/player.setup.spec.ts
andrzejkupczyk: I had to scroll back up to this line to find out that this const holds a string literal, not a test double.

PR #2276, opened 2023-12-14
COTECH-832 | upgrade to always get VAST 4

mfarul:


src/core/utils/tagless-request-url-builder.ts
```diff
@@ -80,7 +80,7 @@ export function buildVastUrl(
 	options: Partial<VastOptions> = {},
 ): string {
 	const params: string[] = [
-		'output=vast',
+		'output=xml_vast4',
```
nandy-andy, line 83: https://support.google.com/admanager/answer/7358411?hl=en

nandy-andy (dismissed):

src/core/utils/tagless-request-url-builder.ts
mfarul: yep, so without it we allow VAST 2,3,4 - as we have VAST 2 as minimal version enabled in GAM and I doubt if JW is handling it....

mklakus-fandom (approved):

PR #2285, opened 2024-01-11
COTECH-961 Bring back JWPlayer strategy rules

andrzejkupczyk:
No need to worry about the Codecov Report since the already existing Player Setup tests are false positives anyway. I've gone ahead and created a [follow-up ticket](https://fandom.atlassian.net/browse/COTECH-975?atlOrigin=eyJpIjoiNzZlNDQ4ZDQ3NjBmNGY2OGE2ZGI3OWUxNmI4NmFjM2EiLCJwIjoiaiJ9) to describe the problem and work on addressing it.

Ticket ref: https://fandom.atlassian.net/browse/COTECH-961 @Wikia/cotech


.prettierignore
```diff
@@ -7,6 +7,7 @@ lib/**/*
 node_modules/**/*
 spec/build/**/*
 
+src/ad-services/connatix/head-script.ts
```
latata, line 10: Does it mean prettier is not checked in the PR checks?

latata (approved):
LGTM

src/platforms/shared/setup/player.setup.ts
```diff
@@ -52,21 +65,37 @@ export class PlayerSetup extends BaseServiceSetup {
...
-		if (showAds) {
+		if (showAds && !strategyRulesEnabled) {
 			communicationService.dispatch(
 				jwpSetup({
 					showAds,
 					autoplayDisabled: false,
 					vastXml: vastResponse?.xml,
 				}),
 			);
+		} else if (strategyRulesEnabled) {
```
dzawada, line 85: Shouldn't it be showAds && strategyRulesEnabled as well?

.prettierignore
andrzejkupczyk: I think so, yeah. When I ran `yarn prettier` locally, this file got prettified.

src/platforms/shared/setup/player.setup.ts
andrzejkupczyk: These are the conditions as they were before when the strategy rules were enabled. I thought the tests were ensuring safety, but since it seems the tests are giving false positives, I'll double-check this logic.

dzawada: with current implementation if ad block is turned on we will still move to the strategy rules part of code with the flag set to false. I'm not sure how strategy rules embed will process it.

andrzejkupczyk: Your statement is correct, but it applies only when the strategy rules are enabled. This means that, regardless of the 'showAds' value, we set up JWP with a VAST URL. I just reverted the previous logic. I'm not entirely sure if it was correct before we disabled strategy rules, but I assumed it was tested.

dzawada: Yeah, I think we can pass it to testing and confirm if everything works as expected in there.

dzawada (approved):

src/platforms/shared/setup/player.setup.ts
latata: I think the idea behind the strategy rules was that they take control over ads completely. So the possible ad block detection should happen on their side as well. So I think it's fine to leave it as is but @nandy-andy could probably have more insights or we can ask Jase.

andrzejkupczyk: I'm currently on the hunt for the ticket related to strategy rules and checking out what acceptance criteria were listed.

nandy-andy (approved):

dzawada (approved):

nandy-andy (approved):

latata (approved):

PR #2289, opened 2024-01-15
Aden 13781

adi-wesolowski-fandom:


src/platforms/ucp-desktop/utils/fmr-rotator.ts
```diff
@@ -36,6 +36,12 @@ export class FmrRotator {
 		private config: FmrRotatorConfig,
 	) {}
 
+	static isApplicable(slotName: string): boolean {
```
preludek, line 39: This seems out of place. I'd move that logic to `src/platforms/ucp-desktop/setup/dynamic-slots/ucp-desktop-slots-definition-repository.ts` and create a method like `isFmrApplicable`.

src/platforms/ucp-desktop/utils/basic-rotator.ts
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	private removeScrollListener(): void {
+		if (this.rotatorListener) {
+			scrollListener.removeCallback(this.rotatorListener);
+			this.rotatorListener = null;
+		}
+	}
+
+	private isItTheLastOne(): boolean {
```
preludek, line 109: nit pick: `isLastSlot` or maybe maybe `finishedRotating`? 🤔

preludek (commented):
Left something to consider. Otherwise looks good 👍

src/platforms/ucp-desktop/utils/fmr-rotator.ts
adi-wesolowski-fandom: it can be moved, ok. 

I just put it there, to keep logic for fmr tightly, and not to spread it over our lib, so it would be easier to clear "fmr" logic all together.

src/platforms/ucp-desktop/utils/basic-rotator.ts
adi-wesolowski-fandom: ok

preludek (approved):

PR #2290, opened 2024-01-16
COTECH-993

David-Iacovone:
Adding support for IAS tracking decoupled from specific player types.


src/platforms/shared/setup/video-player.setup.ts
```diff
@@ -0,0 +1,23 @@
+import {BaseServiceSetup, communicationService, context, eventsRepository, utils} from "@wikia/ad-engine";
+// eslint-disable-next-line no-restricted-imports
```
David-Iacovone, line 2: It's temporary since we agreed to not touch the current logic. Worth to add this to our ad-engine clean up ticket

src/platforms/shared/setup/video-player.setup.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const logGroup = 'player-setup';
+
+@Injectable()
+export class VideoPlayerSetup extends BaseServiceSetup {
+
+    async execute(): Promise<void> {
+        communicationService.on(eventsRepository.VIDEO_PLAYER_RENDERED, () => {
+            this.loadIasTrackerIfEnabled();
+        })
```
latata, line 14: Couldn't we just add it to `player.setup.ts` file? Also, I propose to make these changes based on branch `COTECH-961` as connatix/strategy rules changes are there.

David-Iacovone: I do not think using player.setup.ts is a good idea. It holds a lot of logic not related to this. I would rather clean up jw-player from player.setup than decouple the whole solution from it. I can open a PR towards COTECH-961 branch

latata: Player setup is the place where we currently listen to all video-related events so I think we should keep this one there as well. 
I think we could clean it up after the experiment.

David-Iacovone: I can move there if have a strong opinion about it. However, I would prefer to keep separate due to future benefits. Up to you

src/communication/event-types.ts
```diff
@@ -342,6 +342,11 @@ export const eventsRepository = {
 		name: 'Video player tracking',
 		payload: props<{ eventInfo: object }>(),
 	},
+	VIDEO_PLAYER_RENDERED: {
+		category: '[Video]',
+		name: 'Player rendered',
+		payload: props<{ videoEvent: object }>(),
```
andrzejkupczyk, line 348: I'm thinking we might also need to update the payload sent from the Video Player. Could you confirm, please?

https://github.com/Wikia/video-player/blob/main/lib/connatix/initializeConnatix.ts#L87?

andrzejkupczyk (approved):

src/platforms/shared/setup/video-player.setup.ts
latata: I think we might forget about it at some point so I'd either move it to player.setup.ts or create a ticket to make sure we don't forget we have two player setups and we should clean this up :)

src/communication/event-types.ts
David-Iacovone: Yea, didn't see that. I will update it

src/platforms/ucp-desktop/modes/ucp-desktop-ads.mode.ts
```diff
@@ -40,6 +40,7 @@ export class UcpDesktopAdsMode implements DiProcess {
 		private iasPublisherOptimization: IasPublisherOptimization,
 		private openWeb: OpenWeb,
 		private playerSetup: PlayerSetup,
+		private videoPlayerSetup: VideoPlayerSetup,
```
andrzejkupczyk, line 43: At first glance, it seems a bit odd to have both 'Player Setup' and 'Video Player Setup'. It almost feels as if 'Player Setup' isn't related to video, doesn’t it? 🤔

src/platforms/shared/setup/video-player.setup.ts
andrzejkupczyk: I have to say, I'm with @latata on this. His concerns, as outlined here [link to the comment](https://github.com/Wikia/ad-engine/pull/2290/files#r1455237923), resonate with mine. It might be a matter of naming, but something about it feels a bit unintuitive to me 🤔

David-Iacovone: Okay

src/platforms/shared/setup/player.setup.ts
```diff
@@ -45,6 +47,12 @@ export class PlayerSetup extends BaseServiceSetup {
 		this.jwpManager = jwpManager ? jwpManager : new JWPlayerManager();
 	}
 
+	async execute(): Promise<void> {
```
latata, line 50: shouldn't we put the code within the `call` method instead of overriding `execute`? 🤔

nandy-andy: Good question. It looks like a mess now :D I'd put it to `call()` since we've been using it in this class since the beginning. However, the `BaseServiceSetup` indeed implements `execute()`, so maybe it's worth to replace `call()` with `execute()` 🤔 @domjanik do you have any insights/guidelines? 😊

David-Iacovone: well it has been tested and passed so I am not sure now

nandy-andy: We can always clean it up in a follow-up ;)

David-Iacovone: Alright, I will create the follow up then. We can also add it to ad-engine cleanup 🤔

andrzejkupczyk (approved):

PR #2306, opened 2024-01-23
COTECH-667 Use JWP RTD module with the strategy rules

nandy-andy:
In order to improve CPM in the JWP strategy rules experiment JWP asked us to enable their real-time data module in Prebid.js. It turned out our prebid.js is already built with the module: https://github.com/Wikia/Prebid.js/blob/master/modules.json#L14 Therefore it seems we only needed to pass the additional data in the Prebid config and in the ad unit. I changed it only for IndexExchange as I was informed JWP works closely with them and MediaGrid.


@types/pbjs/index.d.ts
```diff
@@ -23,6 +23,12 @@ interface PrebidAdUnit {
 	ortb2Imp?: {
 		ext: {
 			gpid: string;
+			data?: {
```
nandy-andy, line 26: Source: https://docs.prebid.org/dev-docs/modules/jwplayerRtdProvider.html#implementation-for-publishers

src/ad-bidders/prebid/adapters/index-exchange.ts
```diff
@@ -108,4 +108,23 @@ export class IndexExchange extends PrebidAdapter {
...
+
+		if (!strategyRulesEnabled) {
+			return ortbData;
+		}
+
+		return {
+			...ortbData,
+			data: {
+				jwTargeting: {
+					playerID: '#featured-video__player-container',
```
nandy-andy, line 124: Calling `window.jwplayer().id` would be ideal but it's too soon at this point, so for now I hardcoded the parent ID 🤷‍♂️

spec/platforms/shared/setup/player.setup.spec.ts
```diff
@@ -109,7 +107,7 @@ describe('PlayerSetup', () => {
 			showAds: true,
 			autoplayDisabled: false,
 			videoAdUnitPath: MOCKED_VAST_AD_UNIT,
-			targetingParams: 'experiment_groups=strategy_rules_undefined&pos=featured&rv=1',
+			targetingParams: 'pos=featured&rv=1',
```
nandy-andy, line 110: Since the Optimizely logic got extracted from `playerSetup` the targeting does not include it when test is executed this way. That's not the most important targeting key-val, so I just removed it. However, if there are reasons to put it back I can either modify the test a bit or add one more testing the integration 🤔

src/ad-bidders/prebid/adapters/index-exchange.ts
latata: ID of the element that we render the player in is `featured-video__player`. Also, I think it should not begin with `#`

src/ad-bidders/prebid/adapters/index-exchange.ts
```diff
@@ -108,4 +108,23 @@ export class IndexExchange extends PrebidAdapter {
...
+		const strategyRulesEnabled = context.get('options.video.enableStrategyRules');
+		const initialMediaId = context.get('options.video.jwplayer.initialMediaId');
+
+		if (!strategyRulesEnabled) {
+			return ortbData;
+		}
+
+		return {
+			...ortbData,
+			data: {
```
latata, line 122: I think `ext` is missing here
From the doc:
```
ortb2Imp: {
     ext: {
       data: {
         jwTargeting: {
           // Note: the following Ids are placeholders and should be replaced with your Ids.
           playerID: 'abcd',
           mediaID: '1234'
         }
       }
     }
   }
```

src/ad-bidders/prebid/adapters/index-exchange.ts
nandy-andy: Good point, I'll remove `#` or the whole parameter for now as it's optional 🤷‍♂️

spec/ad-bidders/prebid/adapters/index-exchange.spec.ts
```diff
@@ -135,4 +151,23 @@ describe('IndexExchange bidder adapter', () => {
...
+			enabled: true,
+			slots: {
+				featured: {
+					siteId: '112233',
+				},
+			},
+		});
+		context.set('slots.featured.isVideo', true);
+
+		expect(indexExchange.prepareAdUnits()).to.deep.equal([
```
andrzejkupczyk, line 169: I just realized that we're actually focusing on the results of the `IndexExchange.prepareAdUnits()` function, not testing the `IndexExchange` initialization. Could we please restructure the test to enhance its readability? 🙏🏻

dzawada (dismissed):

spec/platforms/shared/setup/player.setup.spec.ts
andrzejkupczyk: Absolutely, you did the right thing. I remember including it specifically to make sure the tests would pass.

andrzejkupczyk (approved):

spec/ad-bidders/prebid/adapters/index-exchange.spec.ts
nandy-andy: I'm not sure if I understand. You're right that we mostly focus on testing `prepareAdUnit()` of the IX adapter 👌 However, I think the test is structured just fine ;) With the `described()` we describe the whole module that we're testing (IndexExchange adapter) and with the `it()` we describe the specific part of the adapter we test. At the end we have a readable output:
```
  IndexExchange bidder adapter
    ✔ can be enabled
    ✔ prepareAdUnits returns data in correct shape
    ✔ prepareAdUnits for video returns data in correct shape
    ✔ prepareAdUnits for video returns data in correct shape when JWP RTD module is enabled
```
It's not too long and open to add more unit tests not only for `prepareAdUnits()` ;)

The only improvement I can think of is adding one more `describe()` for prepareAdUnit(). Is this what you're suggesting? :) Something like:
```js
describe('IndexExchange bidder adapter', () => {
  describe('can be enabled/disabled', () => {
    it('enabled', () => {});
  });
  describe('prepareAdUnits', () => {
    it('returns data in correct shape', () => {});
    it('returns data in correct shape for video', () => {});
    it('returns data in correct shape when JWP RTD module is enabled for video', () => {});
  });
});
```

andrzejkupczyk: Apologies for any confusion. We've discussed this a few times already, so I was under the impression that we were aligned on this 😬 What I meant is the structure of this particular test:

```ts
it('prepareAdUnits for video returns data in correct shape when JWP RTD module is enabled', () => {
		context.set('options.video.enableStrategyRules', true); // we use JWP RTD when strategy rules are enabled
		context.set('options.video.jwplayer.initialMediaId', MOCKED_INITIAL_MEDIA_ID);
		context.set('slots.featured.isVideo', true);
		const indexExchange = new IndexExchange({...});

		const result = indexExchange.prepareAdUnits();

		expect(result).to.deep.equal([...]);
})
```

nandy-andy: We did! :D Given/when/then or arrange/act/assert 👌 Will do :)

andrzejkupczyk (approved):

PR #2314, opened 2024-01-29
COTECH-993 moved event listener to call() function

David-Iacovone:


nandy-andy (dismissed):

src/communication/event-types.ts
```diff
@@ -361,11 +355,6 @@ export const eventsRepository = {
 		name: 'Video player tracking',
 		payload: props<{ eventInfo: object }>(),
 	},
-	VIDEO_PLAYER_RENDERED: {
```
latata, line 364: It should probably not be removed, right?

David-Iacovone: Hmm wired, it's added not removed on local

David-Iacovone (commented):
I will re open this since github bugged out here.

PR #2324, opened 2024-02-05
TACO-410 Add DSA setup to prebid js

bckatarzyna:


src/ad-bidders/prebid/index.ts
```diff
@@ -413,6 +414,25 @@ export class PrebidProvider extends BidderProvider {
...
 	}
 
+	private configureDSA(): object {
+		if (context.get('options.dsa.enabled')) {
+			return {
+				ortb2: {
+					regs: {
+						ext: {
+							dsa: {
+								required: 1,
```
jbienkowski-fandom, line 424:
```suggestion
								dsarequired: 1,
```
DSA field names changed in the meantime 🙈

src/ad-bidders/prebid/index.ts
```diff
@@ -413,6 +414,25 @@ export class PrebidProvider extends BidderProvider {
...
 
+	private configureDSA(): object {
+		if (context.get('options.dsa.enabled')) {
+			return {
+				ortb2: {
+					regs: {
+						ext: {
+							dsa: {
+								dsarequired: 1,
+								pubrender: 2,
```
jbienkowski-fandom, line 425: pubrender for now should be 0 or 1 I suppose

> Flag to indicate if the publisher will render the DSA Transparency info. This will signal if the publisher is able to and intends to render an icon or other appropriate user-facing symbol and display the DSA transparency info to the end user.
> 0 = Publisher can't render
> 1 = Publisher could render depending on adrender
> 2 = Publisher will render

src/ad-bidders/prebid/index.ts
```diff
@@ -413,6 +414,25 @@ export class PrebidProvider extends BidderProvider {
...
+								datatopub: 2,
```
jbienkowski-fandom, line 426: I wonder, if we provide `datatopub: 2` then shouldn't we also include `transparency` here? And if we do so, what should we put there?
```
"transparency": [
  {
    "domain": "fandom.com",
    "dsaparams": [1] // based on TC string?
  }
]
```

src/ad-bidders/prebid/index.ts
jbienkowski-fandom: Confirmed on daily meeting, we want to set 2 here.

jbienkowski-fandom (approved):

PR #2329, opened 2024-02-06
COTECH-1035 Enable no incontent player experiment via Pathfinder

nandy-andy:


src/core/video/no-incontent-player-experiment.ts
```diff
@@ -1,27 +1,17 @@
...
-		EXPERIMENT_ID: '27342680309',
-		NO_INCONTENT_PLAYER_VARIATION_ID: '27359980218',
-	},
-	// https://app.optimizely.com/v2/projects/17632281258/experiments/27295370448/variations
-	DESKTOP: {
-		EXPERIMENT_ID: '27295370448',
-		NO_INCONTENT_PLAYER_VARIATION_ID: '27266830455',
-	},
-};
+const NO_INCONTENT_PLAYER_EXPERIMENT = defineExperiment({
```
andrzejkupczyk, line 5: It's not a blocker, but perhaps adopting the camelCase naming convention would be more appropriate?

src/core/video/no-incontent-player-experiment.ts
```diff
@@ -1,27 +1,17 @@
...
+	name: 'no-incontent-player-experiment',
+	buckets: ['A', 'B', 'C', 'D', 'E', 'F', 'G'], // 10.93%
```
andrzejkupczyk, line 7: I noticed that you initially mentioned using `ABCDEF`. I just want to double-check that ☝🏻 is correct.

andrzejkupczyk: I've updated the configuration to `ABCD` following Jase's suggestion.

src/core/video/no-incontent-player-experiment.ts
```diff
@@ -1,27 +1,17 @@
...
-	},
-	// https://app.optimizely.com/v2/projects/17632281258/experiments/27295370448/variations
-	DESKTOP: {
-		EXPERIMENT_ID: '27295370448',
-		NO_INCONTENT_PLAYER_VARIATION_ID: '27266830455',
-	},
-};
+const NO_INCONTENT_PLAYER_EXPERIMENT = defineExperiment({
+	name: 'no-incontent-player-experiment',
+	buckets: ['A', 'B', 'C', 'D'], // 6.25%
```
latata, line 7: according to https://docs.google.com/spreadsheets/d/1IxYa8v2_bl_Juquppvs4dtkjzz8TpHJZn1dXywyfBDA/edit#gid=0 we should use small letters `abcd`

src/platforms/shared/setup/base-context.setup.ts
```diff
@@ -115,6 +115,10 @@ export class BaseContextSetup implements DiProcess {
 			'options.video.comscoreJwpTracking',
 			this.instantConfig.get('icComscoreJwpTracking'),
 		);
+		context.set(
+			'options.video.experiments.disabled.no_incontent_player',
+			this.instantConfig.get('icDisableNoIncontentPlayerExperiment'),
```
latata, line 120: minor: I'd avoid negated variables as they might be confusing in usage

src/core/video/no-incontent-player-experiment.ts
andrzejkupczyk: https://github.com/Wikia/pathfinder-lite?tab=readme-ov-file#default-buckets-rules
> Groups a-d are reserved for control and can never be assigned.

latata: I wouldn't believe this readme ;) it is outdated probably (I assume it's from the times before we started using Optimizely) - the spreadsheet is fresh and this is our source of truth ;)

latata: Also I've just realised that the spreadsheet covers only Desktop experiment. 🤔

latata: Aaaah... there is another tab in the spreadsheet with mobile experiments :D

andrzejkupczyk: So... Should it be `abcd` for desktop, and `CDEabc` for mobile?

latata: My guess is `abcd` for desktop no player variant, and `4567` for control desktop group 
`CDE` for mobile no player variant and `abc` for mobile control group (I assume there is a mistake and `abc` was supposed to be a control group)

nandy-andy:
> I wonder if we need a tracking event - similar to what they did here: https://github.com/Wikia/unified-platform/pull/16477/files#diff-0b8ea5c28b6a47a7561e9507b017ba493f2ef611981a00ce9a02834b8f64c17cR3 🤔

I'm not sure why it's needed if we have the information about group in the beacon. It will also make AdEngine package bigger as we'd need to add [browser-tracking-metrics](https://github.com/Wikia/browser-tracking-metrics/) to dependencies if we want to have it unified. I'd avoid it for now but we need to double-check with DA if beacon is good enough.

.npmrc
```diff
@@ -1,2 +1,10 @@
 loglevel="warn"
 save-exact=true
+@wikia:registry=https://artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/
+@fandom:registry=https://artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/
+//artifactory.wikia-inc.com:443/artifactory/api/npm/wikia-npm/:_authToken=${ARTIFACTORY_NPM_TOKEN}
+//artifactory.wikia-inc.com:443/artifactory/api/npm/wikia-npm/:always-auth=true
+//artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/:_authToken=${ARTIFACTORY_NPM_TOKEN}
+//artifactory.wikia-inc.com/artifactory/api/npm/wikia-npm/:always-auth=true
```
latata, line 8: This is not needed anymore, right? 🤔

src/core/video/no-incontent-player-experiment.ts
```diff
@@ -1,27 +1,65 @@
+import defineExperiment from '@fandom/pathfinder-lite/experiments/defineExperiment';
```
latata, line 1: `@wikia/pathfinder-lite`

.npmrc
andrzejkupczyk: Indeed

dzawada (dismissed):

dzawada (approved):

latata (approved):

PR #2331, opened 2024-02-07
TACO-406 - Added new LiveIntent module

domjanik:
In this PR I've added a new module to PreBid alongside with its configuration. Configuration was prepared based on that:
https://docs.prebid.org/dev-docs/modules/userid-submodules/liveintent.html
We use a 3 special IDs that were given to us by Nef


src/ad-bidders/prebid/index.ts
```diff
@@ -229,6 +230,21 @@ export class PrebidProvider extends BidderProvider {
...
+		this.configureLiveIntentConnectedId();
+	}
+
+	private configureLiveIntentConnectedId(): void {
+		const liveIntentConnectedIdConfig = connectedId.getConfig();
+		if (liveIntentConnectedIdConfig) {
+			this.prebidConfig.userSync.userIds.push(liveIntentConnectedIdConfig);
+			targetingService.set('li-module-enabled', ['on']);
+			communicationService.emit(eventsRepository.PARTNER_LOAD_STATUS, {
+				status: 'liveintent_connectid_started',
```
bckatarzyna, line 242: Thank you for keeping consistency!

bckatarzyna (approved):

src/ad-bidders/prebid/index.ts
domjanik: 🙇

PR #2337, opened 2024-02-09
COTECH-1035 Fix tracking in connatix

latata:


nandy-andy (approved):

src/ad-services/connatix/index.ts
```diff
@@ -51,6 +51,7 @@ export class Connatix extends BaseServiceSetup {
 		const connatixPlayer = new ConnatixPlayer();
 		this.playerInjector = playerInjector ? playerInjector : new PlayerInjector(connatixPlayer);
 		this.tracker = tracker ? tracker : new ConnatixTracker();
+		this.dwTracker = dwTracker ? dwTracker : new DataWarehouseTracker();
```
andrzejkupczyk, line 54: I'm considering the idea of renaming the `tracker` property to something more descriptive, like `cnxTracker`, to enhance clarity. Would it be feasible?

andrzejkupczyk (approved):

src/ad-services/connatix/index.ts
nandy-andy: It's putting data to the our data-warehouse, so I'd leave it. Additionally, I'd love to have it removed from the `index.ts` files as it's a quick hack. We have classes like `AnyclipTracker` or `ConnatixTracker`, so if it's not for a quick experiment I'd think of adopting them to send the events 🤔

andrzejkupczyk: Ah, gotcha. Thanks for the clarification.

PR #2338, opened 2024-02-12
ADEN-13883 - upgrade to time delta measurement

adi-wesolowski-fandom:


lnawrotek (approved):

src/core/utils/logger.ts
```diff
@@ -13,7 +13,7 @@ export function getTimeOrigin() {
 }
 
 export function getTimeDelta(): number {
-	return Date.now() - getTimeOrigin();
+	return window.performance?.now() ?? 0;
```
mfarul, line 16:
```suggestion
	return window.performance?.now() ?? Date().now - getTimeOrigin();
```

PR #2345, opened 2024-02-13
TACO-427 Add cmp tracking key val

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-427

### Description
In this PR I've added `cmp` targeting key value that is sent along GAM requests.
I've also added sending information about CMP to `adengadinfo` data table. Because it is temporary, I've added it to `rollout_tracking` column.


@types/window/index.d.ts
```diff
@@ -67,4 +67,5 @@ interface Window {
 			getVariationMap: () => Record<string, { id: string; name: string }>;
 		};
 	};
+	OneTrust?: any;
```
domjanik, line 70: Maybe we could type this object a bit? At least include the most important methods/values here?

src/ad-services/identity-setup/index.ts
```diff
@@ -68,4 +71,8 @@ export class IdentitySetup implements DiProcess {
 			}
 		});
 	}
+
+	private getCmp(): '1t' | 'toi' {
```
domjanik, line 75: Maybe we could extract '1t' | 'toi' to separate type / enum? Like:
type CmpTypes = '1t' | 'toi' 
or 
enum CmpTypes {
  OneTrust = '1t',
  TrackingOptIn = 'toi'
}

szymon-nawrocki: Great idea, changed ✅

src/ad-tracking/compilers/slot-tracking-compiler.ts
```diff
@@ -23,6 +23,10 @@ function checkOptOutSale(): string {
 	return '';
 }
 
+function getCmp(): '1t' | 'toi' {
```
domjanik, line 26: You can re-use CMP_TYPE from identity-setup in here to preserve consistency.

domjanik (changes requested):

domjanik (dismissed):

domjanik (approved):

PR #2349, opened 2024-02-16
ADEN-13905 - wait for pathfinder feature

adi-wesolowski-fandom:


src/platforms/shared/utils/pathfinder.ts
```diff
@@ -0,0 +1,27 @@
...
+	const makeCall = () => {
+		if (called) {
+			return;
+		}
+
+		called = true;
+		callback();
+	};
+
+	setTimeout(() => {
```
lnawrotek, line 18: Do we need "setTimeout" if AE waits for the event?

src/platforms/shared/utils/pathfinder.ts
```diff
@@ -0,0 +1,27 @@
...
+		makeCall();
+	}, 500);
+
+	window.addEventListener('PathfinderModulesLoaded', () => {
```
lnawrotek, line 22: nitpick: You can add `{once: true}`

adi-wesolowski-fandom: yes we can (but this wont remove flag for "already called" :))

src/platforms/shared/utils/pathfinder.ts
adi-wesolowski-fandom: in case there will be no pathfinder at site at all - we dont want to wait for it forever :)

lnawrotek (dismissed):

lnawrotek (approved):

PR #2354, opened 2024-02-19
COTECH-1084 Fix JS error in the console connected to incontent video

nandy-andy:
With the changes in this PR we stop injecting incontent video on pages with featured video for Connatix. This way there is no JS error in the console: `Cannot read properties of null (reading 'appendChild')`.

The `custom.hasIncontentPlayer` flag is being set in the AdEngine when there is no featured video: https://github.com/Wikia/ad-engine/blob/bacb10abe892e49b7fc4f5bcfbb72c999af02fa2/src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts#L56-L62

### DIFF
--- a/src/ad-services/connatix/index.ts
+++ b/src/ad-services/connatix/index.ts
@@ -79,10 +79,12 @@ export class Connatix extends BaseServiceSetup {
 
 		utils.logger(logGroup, 'initialized', this.cid);
 
-		communicationService.on(
-			eventsRepository.AD_ENGINE_UAP_LOAD_STATUS,
-			this.loadOnUapStatus.bind(this),
-		);
+		if (context.get('custom.hasIncontentPlayer')) {
+			communicationService.on(
+				eventsRepository.AD_ENGINE_UAP_LOAD_STATUS,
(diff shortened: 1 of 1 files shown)


dzawada (approved):

src/ad-services/connatix/index.ts
```diff
@@ -79,10 +79,12 @@ export class Connatix extends BaseServiceSetup {
 
 		utils.logger(logGroup, 'initialized', this.cid);
 
-		communicationService.on(
-			eventsRepository.AD_ENGINE_UAP_LOAD_STATUS,
-			this.loadOnUapStatus.bind(this),
-		);
+		if (context.get('custom.hasIncontentPlayer')) {
```
latata, line 82: can't we include it in `isEnabled()`?

nandy-andy: Unfortuantely, no :( If it'd be enough it'd early return in line 60 :/ So, it seems `isEnabled()` answers only the question if the player has been enabled in `instant-config` not if it's "allowed" to be incjected 🙈

PR #2358, opened 2024-02-20
COTECH-1073

David-Iacovone:
A temporary solution allowing us to report properly. This will be changed after the connatix experiment is over.

Ticket: https://fandom.atlassian.net/browse/COTECH-1073


spec/platforms/shared/setup/player.setup.spec.ts
```diff
@@ -106,6 +106,7 @@ describe('PlayerSetup', () => {
 		const expectedDispatchArg = {
 			showAds: true,
 			autoplayDisabled: false,
+			player: 'cnx',
 			videoAdUnitPath: MOCKED_VAST_AD_UNIT,
 			targetingParams: 'pos=featured&rv=1',
```
nandy-andy, line 111: We need the `player=cnx` in this place 😊 So, in GAM reports a dimension can be created based on the key-val.

src/core/models/ad-slot.ts
```diff
@@ -182,7 +182,16 @@ export class AdSlot {
 		return this.adUnit;
 	}
 
-	getVideoAdUnit(): string {
+	getVideoAdUnit(player?: string): string {
```
nandy-andy, line 185: I wonder if we really need to modify the old model for a temporary change ;) Maybe it's worth moving the logic to `initConnatixPlayer()` or `PlayerSetup` which we'll clean up after the migration 🤔

David-Iacovone: I can move it to PlayerSetup if you think that's a better place

spec/platforms/shared/setup/player.setup.spec.ts
David-Iacovone: Can you elaborate a bit more ? 🤔

nandy-andy: In the ticket we wrote we want the `player=cnx` in the key-vals. We want it there, so the RevOrg team can run a report based on the player key-val. It also gives opportunity to set up test ad campaign targeted specifically to Connatix. So, it's the `pos=featured&rv=1&player=cnx` not a new param in the object passed with the event.

src/platforms/shared/setup/player.setup.ts
```diff
@@ -150,9 +150,21 @@ export class PlayerSetup extends BaseServiceSetup {
...
+			videoAdUnitPath: this.modifyAdUnitPath(adSlot),
 			targetingParams: utils.getCustomParameters(adSlot, {}, false),
+			player: 'cnx',
 			vastXml: vastResponse?.xml,
 		});
 	}
+
+	private static modifyAdUnitPath(adSlot: AdSlot): string {
+		let ad = adSlot.getVideoAdUnit();
+		const searchString = 'VIDEO/';
```
David-Iacovone, line 162: Not sure if this string will always have the VIDEO in it. Need to check that

src/ad-services/connatix/player-injector.ts
```diff
@@ -8,7 +8,10 @@ export interface PlayerInjectorInterface {
 
 export class PlayerInjector implements PlayerInjectorInterface {
 	get playerId(): string {
-		return context.get('services.connatix.playerId');
+		const queryParams = new URLSearchParams(window.location.search);
+		const overriddenPlayerId = queryParams.get('cnxPlayerId');
```
nandy-andy, line 12: I wondered if we should have to query-string params for "primary" and "secondary" player but since they don't co-exist it should be fine now, shouldn't it? 🤔

latata: I think we can leave it as is now. We could adjust later when we move the incontent player code to `video-player` repo

src/platforms/shared/setup/player.setup.ts
```diff
@@ -150,9 +150,27 @@ export class PlayerSetup extends BaseServiceSetup {
...
+				{
+					player: 'cnx',
+				},
+				false,
+			),
 			vastXml: vastResponse?.xml,
 		});
 	}
+
+	private static modifyAdUnitPath(adSlot: AdSlot): string {
```
latata, line 165: This method seems to be a bit hacky, but I'm fine with leaving it as is if we agree to make it cleaner at the connatix clean-up phase ;)

latata, line 165: please add unit tests for it

src/platforms/shared/setup/player.setup.ts
```diff
@@ -150,9 +150,27 @@ export class PlayerSetup extends BaseServiceSetup {
...
+		let ad = adSlot.getVideoAdUnit();
+		const searchString = 'VIDEO/';
+		const index = ad.indexOf(searchString);
+		if (index != -1) {
```
latata, line 169: minor: `!==`

spec/platforms/shared/setup/player.setup.spec.ts
```diff
@@ -6,7 +6,7 @@ import { expect } from 'chai';
 import { SinonSpy, SinonStubbedInstance } from 'sinon';
 
 describe('PlayerSetup', () => {
-	const MOCKED_VAST_AD_UNIT = '/5441/test/vast/ad/unit';
+	const MOCKED_VAST_AD_UNIT = '/5441/cnx-test/vast/ad/unit';
```
latata, line 9: Not sure how it works as the `modifyAdUnitPath` looks for `VIDEO/` but I don't see it here... 🤔

src/platforms/shared/setup/player.setup.ts
David-Iacovone: Yeah, I know it's hacky. However, it is small and easy to remove. We agreed that this will be only temporary

src/platforms/shared/setup/player.setup.ts
David-Iacovone: will fix in a sec

spec/platforms/shared/setup/player.setup.spec.ts
David-Iacovone: Worth to mention that this test is connatix specific

latata: ok can we add then a test that will test what `modifyAdUnitPath` does?

David-Iacovone: I didn't add it because it was supposed to be temporary. I can add it if you want it

David-Iacovone: I tried to write a meaningful test but to be honest it's private method. This makes it a bit tricky to test directly this method. Perhaps we can leave it as is. I can try to do some hacks to test but I do not feel like there is a particular need for this. Unless you think it's absolutely necessary. What do you think @latata ?

latata (approved):

andrzejkupczyk (approved):

PR #2361, opened 2024-02-21
Remove N&R cookie based targeting setup

jbienkowski-fandom:
## Description
There is indirect lodash dependency that was missing from package.json... It can be replaced with simple custom functions.


src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
```diff
@@ -235,13 +262,20 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
...
+				subses: result.subses,
+			};
+			const bbDailySession: DailySessionData = {
+				ttag: result.ttag,
+				ftag: result.ftag,
+				pv: result.pv,
+			};
 
-			if (!isMatch(bbCookies[0], bbBrowserSession)) {
+			if (!browserSessionDataMatch(browserSessionData, bbBrowserSession)) {
```
adi-wesolowski-fandom, line 275: `isBrowserSessionDataMatch`?

src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
```diff
@@ -235,13 +262,20 @@ export class NewsAndRatingsTargetingSetup implements DiProcess {
...
 				cookieAdapter.setItem(browserSessionCookie, JSON.stringify(bbBrowserSession));
 			}
-			if (!isMatch(bbCookies[1], bbDailySession)) {
+			if (!dailySessionDataMatch(dailySessionData, bbDailySession)) {
```
adi-wesolowski-fandom, line 278: `isDailySessionDataMatch`?

adi-wesolowski-fandom (approved):

src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
jbienkowski-fandom: tbh I do not like `isMatch` name from lodash at all... if I were to change the name of these functions I would go towards `browserSessionDataMatches` so when you read it it is like a sentence: "if not(browserSessionDataMatches) then set cookie"

adi-wesolowski-fandom: Ok. As for me, on other hand, I like to have functions used in conditions and returning bool, to start with "is". To create sentence (in positive, which is 'by default' IMO) like "if (it) is Xxxx". 

But if you'd like to leave as is, its ok.

src/platforms/news-and-ratings/shared/setup/context/targeting/news-and-ratings-targeting.setup.ts
```diff
@@ -9,9 +9,32 @@ import {
 	targetingService,
 	utils,
 } from '@wikia/ad-engine';
-import isMatch from 'lodash/isMatch.js';
 import { CookieBasedTargetingParams, TargetingParams } from './interfaces/targeting-params';
 
+type BrowserSessionData = Pick<CookieBasedTargetingParams, 'session' | 'subses'>;
```
mfarul, line 14: @szymon-nawrocki I don't think we use those cookies. They were required during migration but now we don't use them.

szymon-nawrocki: @mfarul Exactly.
This logic was transferred to ad-engine for backward compatibility.
I am 90% sure that the method `getCookieBasedTargeting` can be removed.

I'd just ask Seba if they still use `ttag` and `ftag` keys for targeting, if not I'd remove it all together from the ad-engine.

jbienkowski-fandom: I guess they do not as `ftag` is always empty string... (just my short testing observation)

szymon-nawrocki (approved):

PR #2363, opened 2024-02-21
Aden-13832 ATF vs BTF slots auctions split  

lnawrotek:


src/ad-bidders/bidder-provider.ts
```diff
@@ -54,6 +54,14 @@ export abstract class BidderProvider {
 		utils.logger(this.logGroup, 'respond');
 	}
 
+	protected onNoSlotsError(): void {
```
preludek, line 57: In both places this method is used there's already a utils logger that there are no slots to fetch bids. We can either remove it from here or where this function is called

src/platforms/shared/setup/bid-auction-split-experiment-setup.ts
preludek, line 1: We used to have a separate path for experiments. `../experiments` i think 🤔 I suggest moving it either way so it's easier to find and remove later on

preludek (dismissed):

preludek (dismissed):

src/ad-bidders/bidder-provider.ts
```diff
@@ -54,6 +54,12 @@ export abstract class BidderProvider {
...
```
adi-wesolowski-fandom, line 57: nitpick: "on" suggest some kind of listener. Maybe "markNoSlots"?

src/ad-bidders/prebid/prebid-data-tracking.ts
```diff
@@ -0,0 +1,117 @@
...
+			(response: { adId: string; bid: PrebidBidResponse; doc: Document | null }) =>
+				intentIQ.reportPrebidWin(response.bid),
+		);
+
+		utils.logger(logGroup, 'Bids tracking configured');
+	}
+
+	private enableATSAnalytics(): void {
+		if (
+			context.get('bidders.liveRampATSAnalytics.enabled') &&
```
adi-wesolowski-fandom, line 85: nitpick: quick return would be more clear to read. And those conditions can be merged into one function "isLiveRampEnabled()"

src/platforms/shared/experiments/bid-auction-split-experiment-setup.ts
```diff
@@ -0,0 +1,85 @@
...
+
+	call(): void {
+		this.setupExperimentVariants();
+		this.activeExperimentVariant = getExperiment(this.experimentVariants);
+
+		if (this.isExperimentEnabled()) {
+			addExperimentGroupToTargeting(this.activeExperimentVariant.name);
+
+			if (this.isControlVariant()) {
+				utils.logger(logGroup, 'Experiment but control group', this.activeExperimentVariant.name);
```
adi-wesolowski-fandom, line 35: nitpick: here also, deep nesting. With quick return of the first condition, there will be one indent less

src/platforms/shared/experiments/bid-auction-split-experiment-setup.ts
```diff
@@ -0,0 +1,85 @@
...
+			} else {
+				utils.logger(logGroup, 'Experiment - active group', this.activeExperimentVariant.name);
+				context.set('custom.bidAuctionSplitEnabled', true);
+			}
+		}
+	}
+
+	private setupExperimentVariants() {
+		if (this.isMobile) {
+			this.experimentVariants = [
```
adi-wesolowski-fandom, line 45: nitpick: we got it as class property, so maybe extract these two condition result blocks, into separate function? This also would reduce nesting :)

adi-wesolowski-fandom (approved):

src/ad-bidders/bidder-provider.ts
lnawrotek: It's true, but I named this function so that the name would be similar to the existing "onBidResponse"

adi-wesolowski-fandom (approved):

adi-wesolowski-fandom (approved):

adi-wesolowski-fandom (dismissed):

preludek (approved):

preludek (dismissed):

Jorewin (approved):

PR #2365, opened 2024-02-23
COTECH-1050 Change player setup for Connatix so the tentpole is more stable

nandy-andy:
There are three reasons why tentpole wasn't working as expected:
* we had a bug in listening to video events -- we listened for the first one and ignored the rest,
* Connatix especially on slower network connection plays first content video and then pre-roll ad (so resolving the display&video sync after play event messed up the sync)
* Connatix sometimes ignore the response from GAM and doesn't display the video ad (I thought it was because the tagless request but I think it was a timeout set up in the player for preroll).


src/platforms/shared/setup/player.setup.ts
```diff
@@ -21,6 +21,10 @@ import { Injectable } from '@wikia/dependency-injection';
 import { iasVideoTracker } from '../../../ad-products/video/porvata/plugins/ias/ias-video-tracker';
 
 const logGroup = 'player-setup';
+let firstVideoAdImpressionEmitted = false;
+const wasPlayEmittedBeforeAd = (eventName: string) => {
+	return eventName === 'play' && !firstVideoAdImpressionEmitted;
+};
```
latata, line 27: wouldn't it be better to keep it within `resolveVideoDisplaySyncBasedOnPlayerEvent` function? 🤔

latata (approved):

src/platforms/shared/setup/player.setup.ts
nandy-andy: Maybe 🤔 I extracted it here because of readability reasons but probably from memory usage reasons it's better to leave it in the `resolveVideoDisplaySyncBasedOnPlayerEvent()`. Did you have yet another reason in mind? :)

src/platforms/shared/setup/player.setup.ts
```diff
@@ -118,26 +122,51 @@ export class PlayerSetup extends BaseServiceSetup {
...
+			);
+			adSlot.setStatus(AdSlotStatus.STATUS_SUCCESS);
+			adSlot.emit(AdSlotEvent.VIDEO_AD_IMPRESSION);
+		} else if (wasPlayEmittedBeforeAd(eventName)) {
+			PlayerSetup.registerTimeoutForVideoDisplayTakeoverSync(
+				() => firstVideoAdImpressionEmitted === true,
+			);
+		} else if (['adError', 'playError'].includes(eventName)) {
+			videoDisplayTakeoverSynchronizer.resolve();
+		}
```
andrzejkupczyk, line 161: Would the video display synchronization be handled correctly if the event name differs from 'adImpression', 'adError', or 'playError', and if the 'play' event has not been emitted before the first video ad impression? I'm wondering whether these three conditions comprehensively cover all potential scenarios. Or, would it make sense to add an 'else' clause solely for debugging purposes to ensure we're not missing anything?

andrzejkupczyk (approved):

src/platforms/shared/setup/player.setup.ts
```diff
@@ -21,6 +21,10 @@ import { Injectable } from '@wikia/dependency-injection';
 import { iasVideoTracker } from '../../../ad-products/video/porvata/plugins/ias/ias-video-tracker';
 
 const logGroup = 'player-setup';
+let firstVideoAdImpressionEmitted = false;
```
andrzejkupczyk, line 24: I propose simplifying it to `videoAdImpressionEmitted` because we can't necessarily say it's the first video ad impression since it can be called multiple times, right?

src/platforms/shared/setup/player.setup.ts
andrzejkupczyk: Discussed f2f 👍🏻

andrzejkupczyk (approved):

PR #2366, opened 2024-02-23
Aden 13830

p-forfa:


src/platforms/shared/setup/player.setup.ts
```diff
@@ -35,9 +34,6 @@ export class PlayerSetup extends BaseServiceSetup {
 	}
 
 	async call() {
-		communicationService.on(eventsRepository.VIDEO_PLAYER_RENDERED, () => {
-			this.loadIasTrackerIfEnabled();
```
latata, line 39: I'm pretty sure this IAS thing is needed as we added it recently. cc: @nandy-andy

nandy-andy: Good point 👍 The data IAS tracks for sure is needed for targeting direct ads. @p-forfa is removing as many things connected to direct ads as possible AFAIK. However, it's worth double-checking if we don't want to keep tracking video ads plays with IAS even for the "no direct ads bundle" (an indirect ads to track may appear, right?). So, the volumen of data passed to IAS doesn't change.

spec/ad-bidders/prebid/adapters/criteo.spec.ts
```diff
@@ -1,66 +0,0 @@
-import { Criteo } from '@wikia/ad-bidders/prebid/adapters';
```
nandy-andy, line 1: Prebid and A9 are our bidding partners and I always thought about it as indirect ads. Are you removing the tests just because of low value of it? Or removing some bidders as well as the direct ads related stuff?

src/platforms/shared/setup/player.setup.ts
preludek: We have a spreadsheet with partners needed for direct/ indirect and IAS is apparently not needed for indirect - https://docs.google.com/spreadsheets/d/12TGr-D5s2I5jSaSOzIYCXc9WlxUpiBjtiK7Eilboijs/edit#gid=0

preludek (changes requested):

src/platforms/shared/setup/player.setup.ts
p-forfa: Yeah, I was going off of that spreadsheet when removing IAS. Also, the IAS plugin was under Porvata, which I've removed since it's part of UAP.

If we need to keep this for Ads to track properly, then I'll revert it. @nandy-andy / @latata, since the IAS tracker function is in the Porvata directory, should this be moved into a separate or shared directory?

spec/ad-bidders/prebid/adapters/criteo.spec.ts
p-forfa: I was following the list of prebid bidders that SHOULD stay, as outlined in [the tracking ticket](https://fandom.atlassian.net/browse/ADEN-13830). Here's the list of prebid bidders that should've stayed:

Index
Magnite/Rubicon
Triplelift
GumGum
AppnNeus/Xandr
Kargo
Pubmatic

src/ad-services/index.ts
```diff
@@ -1,25 +1,16 @@
...
-export * from './double-verify';
 export * from './duration-media';
 export * from './experian';
-export * from './ias-publisher-optimization';
 export * from './identity-setup';
 export * from './live-connect';
 export * from './liveramp-pixel';
 export * from './lotame';
 export * from './nielsen';
-export * from './open-web';
```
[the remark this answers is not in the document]
p-forfa: Ah, you're right. I removed this based on two things:

[Slack thread](https://fandom.slack.com/archives/C06HNLDJAF8/p1709022018769589?thread_ts=1708990173.431769&cid=C06HNLDJAF8) and the part in the ticket that stated: `NOTE: This assumes we have removed [Duration](https://fandom.atlassian.net/browse/ADEN-13806) and OpenWeb.'`

src/ad-bidders/prebid/adapters-registry.ts
```diff
@@ -3,28 +3,14 @@ import { hasCorrectBidGroup } from '../bidder-helper';
 import {
 	Appnexus,
 	AppnexusAst,
-	Criteo,
```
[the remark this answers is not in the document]
p-forfa: For the prebid adapters, I followed what the ticket stated. Source - [Previous PR comment](https://github.com/Wikia/ad-engine/pull/2366#discussion_r1514659533)

For TestBidder, Wikia, and WikiaVideo I followed [this](https://fandom.slack.com/archives/C06HNLDJAF8/p1709023441081479?thread_ts=1708989580.519929&cid=C06HNLDJAF8).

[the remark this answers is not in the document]
p-forfa: I've also removed a lot of the specs/tests related to TestBidder, Wikia and WikiaVideo. That made the unit test suites pass, but that's expected :). I'm not sure now if it's the right approach to remove this now.

src/platforms/shared/dynamic-slots/activate-floor-adhesion-on-uap.ts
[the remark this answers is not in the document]
p-forfa: Removed in f7fa00edf5e40009b38f91e2a050a54741426f14

src/platforms/shared/setup/player.setup.ts
p-forfa: Ok, so I've brought back the loadIasTrackerIfEnabled function. This was all part of the Porvata plugin directory. I've stripped out all the Porvata code that was unrelated to this function, and left only this function. Hopefully that's enough to get this working, unless there's some magic going on behind the scenes of the ias-video-tracker, where the init() command gets called. Let me know if the code I've brought back in 2062323f5dc8472bc15fcf7273b9663ad4874683 is enough to have this work. I'm honestly not sure how to test this either.

@nandy-andy , @latata

nandy-andy: Sorry, I wasn't aware of the spreadsheet. My understanding was that IAS tracks all our video ads in selected countries and based on that it added additional key-values connected to viewability and fraud detection. I think if the business marked IAS as not needed for indirect then you can go ahead and remove it as everyone is probably aware of what consequences. Probably the indirect bundle will only run in countries where we didn't use the IAS features 🤞

p-forfa: I've asked Lisandro to verify if that'll be 100% needed for the indirect bundle.

preludek (approved):

PR #2368, opened 2024-02-26
COTECH-1087 Pass v1 key-value to GAM when JWP:SR or Cnx players are on the page

nandy-andy:


src/platforms/shared/setup/player.setup.ts
```diff
@@ -145,11 +145,19 @@ export class PlayerSetup extends BaseServiceSetup {
 		adSlot: AdSlot,
 		vastResponse?: VastResponseData,
 	) {
+		const mediaId = window?.mw?.config?.get('wgArticleFeaturedVideo')?.mapping?.connatixMediaId;
```
latata, line 148: This is actually tricky and it won't work as expected as we use playlist now (mediaIds were only for article-targeted videos which is deprecated). I think there is no easy way to know the mediaId of the first video from connatix playlist (I assume this is what we would need there). We can discuss it tomorrow.

nandy-andy:
Decided as "Won't Do" for now. Closing the PR.

PR #2370, opened 2024-02-27
COTECH-1007 | no video experiment

latata:
https://fandom.atlassian.net/browse/COTECH-1007


andrzejkupczyk (approved):

@types/window/index.d.ts
```diff
@@ -17,6 +17,7 @@ interface Window {
 	beacon_id?: string;
 	captify_kw_query_12974?: string;
 	canPlayVideo?: any;
+	fandomIsVideoPossible?: () => boolean;
```
dzawada, line 20: The name could be better, I think isVideo would be enough or isVideoAvailable, not a blocker for sure, just FYI.

dzawada (approved):

@types/window/index.d.ts
latata: as it's a global variable it's always better to have more specific name to avoid conflicts and confusion

PR #2373, opened 2024-02-27
COTECH-1101 Add data privacy consents to VAST URL when in "tagless" mode

nandy-andy:
In countries with video+display sync direct ad product we build VAST URL and send request to GAM as soon as possible. But it lacked `gdpr_consent` parameter. In this PR I'm adding the param as well as the `us_privacy` one. I'm also removing unused function and cleaning up the unit test.

Why do we need to add the parameters by ourselves? Because we send this request by ourselves without GPT or IMA libraries which reach out to IAB APIs for the params:
> Passing the TC string to tags: GPT, GPT Passbacks, AdSense, and Ad Exchange Tags automatically communicate with the IAB CMP to forward the TC string to Ad Manager without publisher configuration. IMA SDK and Mobile Ads SDK automatically obtain, parse, and respect the TC string from within local storage.

Source: https://support.google.com/admanager/answer/9805023


dzawada (approved):

src/core/utils/tagless-request-url-builder.ts
```diff
@@ -141,43 +140,11 @@ export function buildVastUrl(
...
 
-	if (over18) {
-		params.push(`over_18=${over18}`);
+		const { type, consentString } = trackingOptIn.getConsentData();
+		type === 'gdpr' && consentString ? params.push(`gdpr_consent=${consentString}`) : undefined;
+		type === 'ccpa' && consentString ? params.push(`us_privacy=${consentString}`) : undefined;
 	}
 
-	if (ppid) {
-		params.push(`ppid=${ppid}`);
```
domjanik, line 162: Why PPIDs are being removed from here ? 🤔

src/core/utils/tagless-request-url-builder.ts
```diff
@@ -141,43 +140,11 @@ export function buildVastUrl(
...
-			`t=${encodeURIComponent(
-				Object.keys(options.targeting)
-					.filter((key: string) => options.targeting[key])
-					.map((key: string) => `${key}=${options.targeting[key]}`)
-					.join('&'),
-			)}`,
-		);
-	}
-
-	params.push(`rdp=${trackingOptIn.isOptOutSale() ? 1 : 0}`);
```
domjanik, line 180: AFAIK - rdp ( param telling us about user consent for data selling, CCPA ) and npa ( param telling us about user consent for gdpr ) params are later used. Maybe it would be good to leave them in those requests? Also - we're currently missing NPA param in these requests

src/core/utils/tagless-request-url-builder.ts
```diff
@@ -141,43 +140,11 @@ export function buildVastUrl(
...
+		type === 'gdpr' && consentString ? params.push(`gdpr_consent=${consentString}`) : undefined;
+		type === 'ccpa' && consentString ? params.push(`us_privacy=${consentString}`) : undefined;
 	}
 
-	if (ppid) {
-		params.push(`ppid=${ppid}`);
-	}
-
-	if (isCoppaSubject()) {
-		params.push('tfcd=1');
```
domjanik, line 166: Are you sure, that it is not used by @sseyda/Nef for reports?

domjanik (changes requested):

src/core/utils/tagless-request-url-builder.ts
nandy-andy: I removed whole `buildTaglessRequestUrl()` because the only place it's been used was a unit test and GitHub presents the change in a weird way. So it's all still there when building a VAST URL: https://github.com/Wikia/ad-engine/blob/2b2c673d9761508df3088bdb2798b197834e84de/src/core/utils/tagless-request-url-builder.ts#L104-L110

src/core/utils/tagless-request-url-builder.ts
nandy-andy: It's still there :)

src/core/utils/tagless-request-url-builder.ts
nandy-andy: I can add `npa` 👌 

Although, are we sure it's `npa` query-string param not `npa` key-val in `custom_param` query-string? 🤔

nandy-andy: I've added `npa` query-string param.

domjanik (dismissed):

src/ad-products/video/jwplayer/helpers/vast-tagless-request.ts
```diff
@@ -35,12 +37,39 @@ export class VastTaglessRequest {
...
 		}
+
 		const biddersTargeting = await this.bidders.getBidParameters(slotName);
-		return utils.buildVastUrl(aspectRatio, slotName, {
+		const defaultVastOptions: Partial<utils.VastOptions> = {
 			vpos: position,
 			targeting: biddersTargeting,
-			isTagless: true,
-		});
+			isTagless: isTagless,
```
latata, line 45: mnor: 
could be just 
```
isTagless,
```

instead of 
```
isTagless: isTagless,
```

latata (approved):

PR #2374, opened 2024-02-28
Prevent exception when mw and fandomContext.tracking is not available

latata:


src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -9,11 +9,16 @@ export class TrackingParametersSetup implements DiProcess {
 	constructor(private instantConfig: InstantConfigService) {}
 
 	private getPvUniqueId() {
-		return (
+		const pvUID =
 			getMediaWikiVariable('pvUID') || // UCP
 			window.pvUID || // F2
-			window.fandomContext.tracking.pvUID // N+R
-		);
+			window.fandomContext?.tracking?.pvUID; // N+R
```
adi-wesolowski-fandom, line 15: lets add `|| 0`, temporary, as we may be able to find in DWH reports, how many of such cases are, where pvUID is missing.

adi-wesolowski-fandom (changes requested):

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -9,11 +9,17 @@ export class TrackingParametersSetup implements DiProcess {
...
+		const pvUID =
 			getMediaWikiVariable('pvUID') || // UCP
 			window.pvUID || // F2
-			window.fandomContext.tracking.pvUID // N+R
-		);
+			window.fandomContext?.tracking?.pvUID || // N+R
+			0;
+
+		if (!pvUID) {
+			console.error('pvUID is not set');
```
mklakus-fandom, line 19: Can we `console.error` more precise message?

E.g.
```ts
console.error('pvUID is not set correctly');
```
I think the error message is misleading,because when `pvUID` is equal to 0 we would get the error message and also we `return pvUID` which I suppose is intended for this case.

mklakus-fandom (approved):

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -9,11 +9,17 @@ export class TrackingParametersSetup implements DiProcess {
 	constructor(private instantConfig: InstantConfigService) {}
 
 	private getPvUniqueId() {
```
jbienkowski-fandom, line 11: `getPvUniqueId` is only used by `getLegacyTrackingParameters`, on most of the platforms `icDisableLegacyTrackingParameters` is set to `true` and we are not relying on these values (we use `window.fandomContext`). I have set it consistently for all platforms and I believe the best solution would be to delete `getLegacyTrackingParameters` method all together.
Please see https://github.com/Wikia/ad-engine/pull/2381

PR #2394, opened 2024-03-12
ADEN refactor tracking

preludek:


src/ad-tracking/tracking-urls.ts
```diff
@@ -1,4 +1,6 @@
-import { Dictionary } from '@wikia/ad-engine';
+interface Dictionary<T = any> {
+	[key: string]: T;
+}
```
jbienkowski-fandom, line 3: interesting idea (implemented by @adi-wesolowski-fandom on another branch) was to split one big `trackingUrls` into separate exported objects such as:

```javascript
export const AD_ENG_LOAD_TIMES = {
		name: 'AdEngLoadTimes',
		icbmName: 'loadTimes',
		url: 'https://beacon.wikia-services.com/__track/special/adengloadtimes',
		allowed: {
			sampling: true,
			aggregation: true,
			aggregationLimit: 5,
		},
	};
```
it has few advantages, you import only the url which you need (one small object instead of big object with reference) and helps with webpack tree shaking
this PR is already quite big so I would suggest doing it as a separate follow-up task

jbienkowski-fandom (approved):

preludek:
those changes were merged to ADEN-13916 noads optimization

PR #2397, opened 2024-03-13
Aden 13793 ver keyval

mfarul:


lnawrotek (approved):

src/platforms/shared/setup/tracking-parameters.setup.ts
```diff
@@ -29,6 +35,10 @@ export class TrackingParametersSetup implements DiProcess {
 					: false,
 			);
 		});
+
+		if (window.ads?.version?.ae) {
```
jbienkowski-fandom, line 39: in `log-version.ts` we are also setting `window.ads.adEngineVersion` based on `package.json`, should we get rid of that? btw, aeLoader based version is better, because it will get `prod/BRANCH` version as well

jbienkowski-fandom (approved):

jbienkowski-fandom (approved):

src/platforms/shared/setup/tracking-parameters.setup.ts
mfarul: removed

jbienkowski-fandom (approved):

PR #2412, opened 2024-03-20
TACO-458 - Fixed slot targeting

domjanik:
Currently, we’ve a problem with retrieving slot targeting. When a user enters a page on mobile, we completely ignore the targeting set for mobile_top_leaderboard. That’s why some values, like testgroups are never available in our slots. In this PR I’ve added code responsible for merging slot-specific targeting with default slot targeting ( mobile_top_leaderboard + top_leaderboard ) so all values would be available.


src/core/models/ad-slot.ts
```diff
@@ -309,7 +316,7 @@ export class AdSlot {
 				result[key] = value;
 			}
 		});
-
+		utils.logger('DJ', result);
```
adi-wesolowski-fandom, line 319: ?

src/core/models/ad-slot.ts
```diff
@@ -91,7 +91,9 @@ export class AdSlot {
 	viewed: Promise<void> = null;
 
 	constructor(ad: AdStackPayload) {
+		utils.logger('DJ:ad', ad);
```
adi-wesolowski-fandom, line 94: ?

src/core/models/ad-slot.ts
```diff
@@ -91,7 +91,9 @@ export class AdSlot {
...
 		this.config = context.get(`slots.${ad.id}`) || {};
+		utils.logger('DJ:config', this.config);
```
adi-wesolowski-fandom, line 96: ?

jbienkowski-fandom (approved):
LGTM

PR #2415, opened 2024-03-21
ADEN-13939 - resize TLB placeholders on refresh #2414

mklakus-fandom:
QE-217 test update

### DIFF
--- a/src/platforms/shared/templates/helpers/uap-dom-manager.ts
+++ b/src/platforms/shared/templates/helpers/uap-dom-manager.ts
@@ -1,9 +1,11 @@
 import {
 	AdSlot,
 	communicationService,
+	context,
 	eventsRepository,
 	TEMPLATE,
 	UapParams,
+	universalAdPackage,
 } from '@wikia/ad-engine';
 import { Inject, Injectable } from '@wikia/dependency-injection';
 import { PAGE } from '../configs/uap-dom-elements';
@@ -52,7 +54,16 @@ export class UapDomManager {
 		this.setSlotHeight(`${this.reader.getSlotHeightImpact()}px`);
 	}
 
+	private addTransitionProperty() {
+		const transition = `height 300ms ${universalAdPackage.CSS_TIMING_EASE_IN_CUBIC}`;
+
+		const topAdsContainerElement: HTMLElement = document.querySelector('.top-ads-container');
+		this.manipulator.element(this.adSlot.getElement()).setProperty('transition', transition);
+		this.manipulator.element(topAdsContainerElement).setProperty('transition', transition);
+	}
+
 	private setSlotHeight(height: string): void {
+		this.addTransitionProperty();
 		this.manipulator.element(this.adSlot.getElement()).setProperty('height', height);
 	}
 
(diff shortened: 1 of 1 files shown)


src/platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -63,11 +74,29 @@ export class UapDomManager {
 	}
 
 	setPlaceholderHeightResolved(): void {
-		this.setPlaceholderHeight(`${this.reader.getSlotHeightResolved()}px`);
+		const slotRefresherConfig = context.get('slotConfig.slotRefresher.sizes') || {};
+		if (this.adSlot.getSlotName() in slotRefresherConfig) {
+			const iframe = this.adSlot.getIframe();
+			const placeholderHeight = `${Number(iframe.height) + 39}px`;
```
lnawrotek, line 80: 39 is a magic number :)

src/platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -63,11 +74,29 @@ export class UapDomManager {
 	}
 
 	setPlaceholderHeightResolved(): void {
-		this.setPlaceholderHeight(`${this.reader.getSlotHeightResolved()}px`);
+		const slotRefresherConfig = context.get('slotConfig.slotRefresher.sizes') || {};
```
lnawrotek, line 77: This code is duplicated (lines 77-86 and 90-99). Please add a new method

src/platforms/shared/templates/helpers/uap-dom-manager.ts
mklakus-fandom: Yeah, but I didn't find anything related to the: 90px tlb has a 129px placeholder etc...

src/platforms/shared/templates/helpers/uap-dom-manager.ts
mklakus-fandom: moved the part inside of the if statement to a separate method in [945c761](https://github.com/Wikia/ad-engine/pull/2415/commits/945c76117218ec0cf208de8acab1bf91ad8e0211) as the else is different for these

src/platforms/shared/templates/helpers/uap-dom-manager.ts
mklakus-fandom: what do you think about doing something like this:

```css
padding: 18px 0 0 0;
```

src/platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -62,12 +73,27 @@ export class UapDomManager {
 			.setProperty('clip', this.reader.getSlotHeightClipping());
 	}
 
+	resizePlaceholderToIframe(height: number) {
+		const PLACEHOLDER_HEIGHT_OVERHEAD = 39;
```
lnawrotek, line 77: nitpick: it can be `static PLACEHOLDER_HEIGHT_OVERHEAD = 39`

example https://github.com/Wikia/ad-engine/blob/b706e4af5036ce1a455afdaaf5eb7ea178a61f0b/src/core/models/ad-slot.ts#L65

PR #2420, opened 2024-03-26
COTECH-1025 Track eligible event for connatix

andrzejkupczyk:
https://fandom.atlassian.net/browse/COTECH-1025


src/ad-services/connatix/connatix-tracker.ts
```diff
@@ -44,6 +44,11 @@ export class ConnatixTracker implements VideoTracker {
 		this.track(eventName, VideoEventProvider.getEventData(this.getVideoData(eventName)));
 	}
 
+	trackEligible() {
+		const eventName = 'eligible';
```
latata, line 48: redundant var ;)

latata (approved):

src/ad-services/connatix/connatix-tracker.ts
andrzejkupczyk: Objectively speaking, that is not true :)

dzawada (dismissed):

src/ad-services/connatix/connatix-tracker.ts
```diff
@@ -7,7 +7,7 @@ const logGroup = 'connatix';
 export class ConnatixTracker implements VideoTracker {
 	private playerApi: ConnatixPlayerApi;
 
-	private trackingEvents = {
+	private playerEvents = {
```
nandy-andy, line 10: If we want to rename the property then I think both `trackingEvents` and `playerEvents` are equally wrong :D It's rather `playerEventsToTrackingEventsMap` 😆 The keys are names of the events emitted by the player whereas the values are the events/values we store in our DWH.

andrzejkupczyk: The context is connatix TRACKER, so adding "to tracking" seems redundant. I can update the name if needed. What's you final verdict? :)

nandy-andy: We've discussed it on slack. It was faster this way :)

nandy-andy (approved):

PR #2434, opened 2024-04-04
Revert "Revert "ADEN-13967 | Add DTCO inc1 tags""

vforge:
Reverts Wikia/ad-engine#2431

https://github.com/Wikia/unified-platform/pull/16996


p-forfa (approved):
Revert of revert LGTM. The one thing that I'm not sure about is v146.31.0 being released to the [N&R platform only](https://fandom.slack.com/archives/C9MDQHA6Q/p1712245989781869).

These changes would make it to the fandomdesktop and fandommobile apps if we were to release this.

PR #2436, opened 2024-04-04
ADEN-14016

mklakus-fandom:
https://fandom.atlassian.net/browse/ADEN-14016


lnawrotek (approved):

src/platforms/shared/templates/helpers/uap-dom-manager.ts
```diff
@@ -61,6 +61,9 @@ export class UapDomManager {
 		const transition = `height 300ms ${universalAdPackage.CSS_TIMING_EASE_IN_CUBIC}`;
```
jbienkowski-fandom, line 61: (turbo optimization) this could be moved after the guard clause

jbienkowski-fandom (approved):
LGTM

PR #2440, opened 2024-04-08
COTECH-1063 Disable in-content player on home pages

andrzejkupczyk:


src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
```diff
@@ -144,4 +140,16 @@ export class CommonTags implements TargetingProvider<Partial<SlotTargeting>> {
...
+	private shouldDisplayIncontentPlayer(pageType: string, videoStatus: VideoStatus): boolean {
+		if (pageType === 'home') {
+			return false;
+		}
+		return (
+			!!videoStatus.hasVideoOnPage &&
+			// COTECH-1007: No video experiment - function set in UCP
+			!window.fandomIsVideoPossible?.() &&
+			!!document.querySelector(context.get('templates.incontentAnchorSelector'))
+		);
```
latata, line 153:
```suggestion
		return (
			pageType !== 'home' &&
			!!videoStatus.hasVideoOnPage &&
			// COTECH-1007: No video experiment - function set in UCP
			!window.fandomIsVideoPossible?.() &&
			!!document.querySelector(context.get('templates.incontentAnchorSelector'))
		);
```

latata (dismissed):

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
andrzejkupczyk: I wrote it like that for the sake of readability, but I think it does not improve anything 😉

latata (approved):

nandy-andy (dismissed):

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
```diff
@@ -144,4 +140,18 @@ export class CommonTags implements TargetingProvider<Partial<SlotTargeting>> {
...
 	}
+
+	private shouldDisplayFeaturedVideo(videoStatus: VideoStatus): boolean {
+		return !!videoStatus.hasVideoOnPage;
+	}
+
+	private shouldDisplayIncontentPlayer(videoStatus: VideoStatus, pageType: string): boolean {
+		return (
+			pageType !== 'home' &&
+			!this.shouldDisplayFeaturedVideo(videoStatus) &&
```
latata, line 151: Wouldn't just `!videoStatus.hasVideoOnPage` be more readable to you? For me it would be ;)

latata (approved):

src/platforms/shared/context/targeting/targeting-strategies/providers/common-tags.ts
andrzejkupczyk: Previously, the functionality relied on the presence of a featured video (regardless of how it was determined), so depending on `hasVideoOnPage` could potentially result in unforeseen outcomes if the logic were to change in the future. That being said, I assume that even if it's less readable, it's more prone to bugs in the feature, wdyt?

andrzejkupczyk: Based on our discussion on Slack and your approval of the PR, I will keep it unchanged.

PR #2454, opened 2024-04-11
Update README.md

psmagiel:


abador:
This will cause errors because the branch name has `()`

PR #2460, opened 2024-04-15
COTECH-1164 Quick, temporary fix for floor_adhesion with JWP:SR

nandy-andy:
This is temporary fix that emits additional event on loading JWP with strategy rules. The event triggers logic that requests ads for floor_adhesion slot once the floating player is closed: https://github.com/Wikia/ad-engine/blob/dev/src/platforms/ucp-mobile/setup/dynamic-slots/ucp-mobile-dynamic-slots.setup.ts#L176-L207

Ideally we'd emit `VIDEO_EVENT` from the JWP with the strategy rules and emit the additional event in AdEngine only we receive it. This is something I'll work on in a follow-up PRs.


latata (approved):

nandy-andy:
We're not happy with this solution. Closing.

PR #2461, opened 2024-04-16
COTECH-1164 Replace ad impression event with video rendered

nandy-andy:
Still not perfect but a bit better solution to have `floor_adhesion` slot after floating video player is closed.


andrzejkupczyk (dismissed):

latata (dismissed):

src/platforms/shared/setup/player.setup.ts
```diff
@@ -97,6 +97,17 @@ export class PlayerSetup extends BaseServiceSetup {
 		}
 	}
 
+	private static getPrimaryVideoAdSlot() {
+		const videoAdSlotName = 'featured';
```
andrzejkupczyk, line 101: I recommend placing it above the class definition or declaring it as a class property to avoid repetition here and on line 121.

andrzejkupczyk (approved):

latata (approved):

andrzejkupczyk (approved):

PR #2463, opened 2024-04-17
ADEN-13845 | ADEN-14052 :: Control Group for Stripped down Indirect Package

p-forfa:
Added in a control group using pathfinder-lite, to create a control sample set against the Indirect AdEng package which is being served through ICBM as a completely different AdEng package.

This will allow us to compare data using key-val pairs against a smaller overall sampleset.


src/platforms/ucp-mobile/setup/experiments/ucp-mobile-indirect-package-experiment-control.ts
```diff
@@ -0,0 +1,54 @@
...
+		this.pathfinderExperimentControl,
+	);
+
+	private configure() {
+		if (!this.isEnabledExperiment() || !this.pathfinderActiveExperimentControl?.name) {
+			utils.logger(logGroup, 'Experiment was disabled');
+			return;
+		}
+
+		addExperimentGroupToTargeting('ucp-mobile-indirect-package-experiment-control');
```
PGalicia, line 46: Small thing: might be best to just use `pathfinderExperimentControlName` instead of this string for a cleaner code

p-forfa: Nice catch. I changed it in the desktop, but forgot about it here.

p-forfa: Fixed in 23eec365e850975cfa533b5a6226f421a9e3175a

PGalicia (approved):

src/platforms/ucp-mobile/setup/experiments/ucp-mobile-indirect-package-experiment-control.ts
PGalicia: Thank you!

adi-wesolowski-fandom (approved):

PR #2469, opened 2024-04-22
ADEN-13878: Show floor adhesion unit in FV pages

PGalicia:


src/platforms/ucp-desktop/setup/dynamic-slots/activate-floor-adhesion-on-fv-pages.ts
```diff
@@ -0,0 +1,29 @@
+import {
+	AdSlotEvent,
+	communicationService,
+	eventsRepository,
+	universalAdPackage,
+} from '@wikia/ad-engine';
+
+export function activateFloorAdhesionOnFVPages(callback: (payload) => void) {
+	// Trigger callback after the scrollable video has been closed and top leaderboard has been resolved
+	communicationService.on(eventsRepository.VIDEO_CLOSE_BUTTON_CLICKED, () => {
```
p-forfa, line 10: This event gets triggered for both JWP and Connatix?

PGalicia: Yup it works for both!

p-forfa (approved):
LGTM.

There is some small issue in the GitHub test and linting job, but it looks like strict typescript check complaints after they were turned on a few weeks ago.

p-forfa (approved):

PR #2470, opened 2024-04-23
TACO-531 Split labrador tracking into dev and prod

bckatarzyna:


src/platforms/shared/tracking/labrador-tracker.ts
```diff
@@ -14,13 +14,19 @@ export class LabradorTracker {
 	 */
 	track(value: string): void {
 		const now = new Date();
+		// eliminates dev records from fact_adenglabradorinfo_events table
+		const trackingUrl =
+			window?.ads?.context?.env === 'dev'
+				? trackingUrls.AD_ENG_DEV_LABRADOR_INFO
```
jbienkowski-fandom, line 20: maybe instead of multiplying whole configuration for dev events, let's just add `_dev` suffix to the URL?
```suggestion
				? trackingUrls.AD_ENG_LABRADOR_INFO + '_dev'
```

domjanik: I think that Kasia's solution is a bit cleaner and might be easier to use and maintain in the future 🤔 Also - it will be easier to change config for existing events while maintaining the other config for dev env ( for example turn on sampling for prod events only )

domjanik (approved):

jbienkowski-fandom (dismissed):
LGTM

domjanik (approved):

PR #2473, opened 2024-04-30
Fix for "Cannot read properties of null (reading 'classList')" Sentry issue

latata:
* removing `@ts-strict-ignore` comments from two files
* adding a null-check for slot element

Sentry issue: https://fandom-lf.sentry.io/issues/5043089866/?environment=preview&project=4506854504923136&query=is:unresolved+adengine:True+!external:True&sort=freq&statsPeriod=7d&stream_index=0


jbienkowski-fandom (approved):
LGTM

PR #2475, opened 2024-04-30
Fix for Sentry issue - defineSlot may return null

latata:
According to https://developers.google.com/publisher-tag/reference#googletag.defineSlot `defineSlot` may return `null`

https://fandom-lf.sentry.io/issues/5044489531/?environment=preview&project=4506854504923136&query=is:unresolved+adengine:True+!external:True&sort=freq&statsPeriod=7d&stream_index=1


src/core/providers/gpt-provider.ts
```diff
@@ -285,7 +285,7 @@ export class GptProvider implements Provider {
 
 		return window.googletag
 			.defineSlot(adSlot.getAdUnit(), adSlot.getDefaultSizes(), adSlot.getSlotName())
-			.defineSizeMapping(sizeMap.build());
+			?.defineSizeMapping(sizeMap.build());
```
jbienkowski-fandom, line 288: this method will now return `googletag.Slot | undefined`

jbienkowski-fandom (approved):
LGTM

PR #2476, opened 2024-04-30
ADEN-14010 :: Add FandomTaxonomy consumer

vforge:
- Add consumer for https://github.com/Wikia/unified-platform/pull/17211
- Replace current DTCO hack logic with proper one


src/core/services/fandom-taxonomy-service.ts
```diff
@@ -0,0 +1,105 @@
...
+	public static applyToTargetingService() {
+		const fandomTaxonomy = new FandomTaxonomy();
+
+		// Regular Key Values are just added to the targeting service
+		targetingService.extend({
+			...targetingService.dump(),
+			...fandomTaxonomy.getKeyValues(),
+		});
+
+		// Tags are added to the targeting service as well, but should be merged with the existing values (if possible)
```
szymon-nawrocki, line 26: Could you tell me why the taxonomy tags are added to new `fandomTaxonomy` object?
We already have taxonomy tags added to `window.fandomContext.page.tags` (page level tags) and `window.fandomContext.site.tags` (site level tags) and these are already 'existing' values that you are mentioning in this comment.

I am not the biggest fan of the fact that we have taxonomy tags in 2 separate objects and two of these objects are used by ad-engine for getting the same thing - taxonomy tags

vforge: @szymon-nawrocki This new taxonomy will be injected into the page via the JavaScrtipt script tag. The intention is not to pollute any existing object and not have conflict with it.

As for the long-term plans for current tags, IIRC the ad-tag-manager-based tags will the replaced by DTCO tags, but I am not sure what will happen to the rest of them. This is the current requirement from Eric to have it work that way.

vforge: For some of the new DTCO values we're creating new KVs (code above), but for `fran` tag specifically, we were asked to merge it with the existing one. The current code is my best attempt at doing that in a clear way, but I am open to any other suggestions.

vforge: I am sure the long-term intention is to replace the current page/wiki tags with DTCO taxonomy, but for now, we need to have some overlap.

szymon-nawrocki: okay, thanks for the explanation

szymon-nawrocki (approved):

PR #2477, opened 2024-05-06
COTECH-1177 | Fix BaseServiceSetup timeouts

latata:
When BaseServiceSetup reaches the timeout it should not only `setInitialized` but also call `this.call` method.
The changes introduced here will ensure that the `call` method is called even on timeout.


spec/core/pipeline/imps/partner-pipeline/partner-pipeline.spec.ts
```diff
@@ -46,14 +53,39 @@ describe('PartnerPipeline', () => {
 				spy(2);
 			});
 
-		await progress(40);
+		await progress(41);
```
andrzejkupczyk, line 56: I find the current setup a bit challenging to understand initially. Perhaps we could introduce some clearly named constants? Here's a suggestion:

```js
const SHORT_WAIT_MS = 20;
const MEDIUM_WAIT_MS = 40;
const LONG_WAIT_MS = 1000;
const TIMEOUT_SHORT_MS = 10;
```

Instead of progressing by an unclear increment of `41`, we could break it down into two clear steps:

```js
await progress(MEDIUM_WAIT_MS);
await progress(1);
```

andrzejkupczyk (approved):

PR #2484, opened 2024-05-09
ADEN-14022: Add display priority for the Ad Choices screen

Jorewin:


src/platforms/shared/setup/ad-engine-stack.setup.ts
```diff
@@ -1,14 +1,45 @@
...
 
 		adEngineInstance = new AdEngine();
 		return adEngineInstance.init();
 	}
+
+	private setUpAdChoicesDisplayPriority(adSlot: AdSlot): void {
+		const iFrame = adSlot.getIframe();
+		const iFrameHasBody = !!iFrame?.contentDocument?.body;
+
+		if (iFrameHasBody) {
```
mklakus-fandom, line 29: nitpick: I think we can do an `early return` there

Jorewin: Done!

PR #2485, opened 2024-05-09
TACO-511 Add Price Floors Mechanism

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-511

### Description
Add Price Floors mechanism for handling Prebid Price Floor Module.
It should work in a way that that the configuration will be set in ICBM variable `icPrebidPriceFloorsConfig` and it will have the configuration with price per bidder. It will look like this:
```
{
    "appnexus": 0.1,
    "rubicon": 0.05,
    "indexExchange": 0.15,
    "ozone": 0.20,
    "triplelift": 1.10,
    "medianet": 2.22
}
```

This configuration should be then set in pbjs config `floors.data.values`.

The keys are mapped to look like: "banner|<bidder_name>: <bidder_price".
The banner mediaType is added so it will be applied only to banner mediaType and not affect video bids for example.

Link do Price Floors module documentation: https://docs.prebid.org/dev-docs/modules/floors.html


src/ad-bidders/prebid/price-floors-mechanism.ts
```diff
@@ -0,0 +1,101 @@
...
+	isEnabled(): boolean {
+		return context.get('bidders.prebid.priceFloors.enabled');
+	}
+
+	bidderType(bidRequest: BidRequest) {
+		return bidRequest.bidder;
+	}
+
+	mapConfigKeysToBannerMediaType(config: BidderPriceFloors) {
+		const mediaType = 'banner';
```
bckatarzyna, line 89: Do we want to use it also for video?

src/ad-bidders/prebid/price-floors-mechanism.ts
```diff
@@ -0,0 +1,101 @@
...
+		const mappedConfig = {};
+
+		for (const [key, value] of Object.entries(config)) {
+			const combinedKey = mediaType + '|' + key; // TODO: change name
```
bckatarzyna, line 93: TODO left - was it supposed to be done in this ticket?
If not, you can add the code of the ticket that is going to change it (TACO-XXX) to let the others know when it should be done

src/ad-bidders/prebid/price-floors-mechanism.ts
```diff
@@ -0,0 +1,101 @@
...
+		utils.logger(logGroup, 'enabled');
+
+		const config = context.get('bidders.prebid.priceFloors.config');
+
+		if (!config || Object.keys(config).length === 0) {
+			utils.logger(logGroup, 'no config provided');
+			return;
+		}
+
+		if (typeof config !== 'object') {
```
bckatarzyna, line 50: 1. What will happen if we have a valid object but with a typo in bidder's name?
2. What will happen if we have a valid object but with invalid price (e.g. different type)?

src/ad-bidders/prebid/price-floors-mechanism.ts
szymon-nawrocki: As discussed on yesterday Daily for now we want to use it only for banners.
This is the main reason I added this mediaType variable, to make sure that this change won't affect video bids.
Setting price floors for video bids will be considered in the future

src/ad-bidders/prebid/price-floors-mechanism.ts
szymon-nawrocki: `combinedKey` is the final name :) Sorry for not removing the TODO

src/ad-bidders/prebid/price-floors-mechanism.ts
szymon-nawrocki: 1. With typo in a bidder name like for example `banner|openxD` the price floor won't be applied to openx
2. When price is invalid and it is for example string '666' then Prebid does not add it to the `config.floors.data.values`

image

src/ad-bidders/prebid/price-floors-mechanism.ts
szymon-nawrocki: But thanks for pointing that out, I've added a comment in code for the clarification

src/ad-bidders/prebid/price-floors-mechanism.ts
bckatarzyna: Ok, so they're not critical issues, thank you for the answer!

src/ad-bidders/prebid/price-floors-mechanism.ts
bckatarzyna: I wasn't present yesterday, thank you for the explanation 🙏

bckatarzyna (approved):

bckatarzyna (dismissed):

bckatarzyna (approved):

PR #2490, opened 2024-05-12
TACO-566 - Optimized slot refreshing 

domjanik:
In this PR I’ve slightly changed the behavior of the DSA Handler. Previously it was adding a new listener for each slot every time a slot has refreshed. It caused multiple problems with both - web performance and synchronization of all our handlers. Also - I’ve added new loggers so we can track whole DSA process.


src/ad-bidders/prebid/dsa/handler.ts
```diff
@@ -6,20 +6,48 @@ import { DsaData } from '../utils/dsa';
 type DsaRenderProps = DsaData;
 
 export class DsaHandler {
+	private static isRegistered = false;
```
jbienkowski-fandom, line 9: `const dsaHandler = new DsaHandler()'` acts as a singleton, therefore it is not necessary to introduce the static methods and properties

src/ad-bidders/prebid/dsa/handler.ts
```diff
@@ -6,20 +6,48 @@ import { DsaData } from '../utils/dsa';
...
+					utils.logger('DSA', 'Slot does not exist');
 					return;
 				}
 
 				if (!slot.getElement()) {
+					utils.logger('DSA', 'Slot ' + slot.getSlotName() + ' element does not exist');
+					return;
+				}
+
+				if (!behalf || !paid) {
```
jbienkowski-fandom, line 41: According to the OpenRTB spec, both Behalf and Paid must be present, when I was implementing TACO-461 I think there were some cases in which paid was not available, but I have double check with DWH and both fields are present in the bidder response

src/ad-bidders/prebid/dsa/handler.ts
```diff
@@ -6,20 +6,48 @@ import { DsaData } from '../utils/dsa';
...
+					utils.logger('DSA', 'For slot ' + slot.getSlotName() + ' behalf or paid is missing');
 					return;
+				} else {
```
jbienkowski-fandom, line 44: this else is not necessary here

src/ad-bidders/prebid/prebid-data-tracking.ts
```diff
@@ -78,12 +78,15 @@ class PrebidDataTracking {
 			});
 		};
 		const handleDsa = (response: PrebidBidResponse, document: Document | null) => {
-			if (dsaDataAvailable(response)) {
```
jbienkowski-fandom, line 81: why get rid of this if check?

jbienkowski-fandom (changes requested):

jbienkowski-fandom (approved):

jbienkowski-fandom (approved):

PR #2494, opened 2024-05-16
ADEN-14060: Add sticky top box ad experiment

PGalicia:
Ticket: https://fandom.atlassian.net/browse/ADEN-14086


src/core/video/pathfinder-lite/experiments/isExperimentEnabled.ts
```diff
@@ -0,0 +1,14 @@
+import { InstantConfigService } from '../../../services';
+import { Experiment } from '../types';
+
+type RunningExperimentType = Experiment | null | undefined;
+
+export default function isExperimentEnabled(
```
p-forfa, line 6: This'll be useful for the future. This logic was probably duplicated multiple times.

p-forfa (approved):
LGTM.

I also followed the testing instructions, and verified that the top box ad sticks on buckets R and S. I've also verified that tracking events send out the experiment key-val when bucketed into the active and control groups.

PR #2499, opened 2024-05-17
TACO-594 Prebid - Fix ad units with bids=undefined

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-594

### Description
With the current Prebid code from time to time it happens that `top_leaderboard` and `floor_adhesion` ad units have field `bids=undefined`.

This breaks the Prebid code, because Prebid expects bids field to be array, since they are iterating on it: [Link](https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.js#L358)
When Prebid gets `bids=undefined` in Prebid auction, the code breaks and in result the price floor module does not work in the particular auction. [Link](https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.js#L243-L244)

> This error happening causes the auction to not set floors for it at all. Never sets the _floorDataForAuction so the getFloor function exits early each time

**In this PR I am trying to fix it. The fix I applied does not add ad units that have `bids` field `undefined` to ad units setup.**


jbienkowski-fandom (approved):

src/ad-bidders/prebid/adapters-registry.ts
```diff
@@ -100,8 +100,12 @@ class AdaptersRegistry {
 				const adapterAdUnits = adapter.prepareAdUnits();
 
 				adapterAdUnits.forEach((adUnit) => {
+					const hasCorrectBidsField =
```
domjanik, line 103: Nit pick - maybe we could extract it to a separate function similary to hasCorrectBidGroup ?

domjanik (approved):

src/ad-bidders/prebid/adapters-registry.ts
szymon-nawrocki: Good idea. Added ✅  Also used this opportunity to add unit tests for it as well  :)

domjanik (dismissed):

src/ad-bidders/bidder-helper.ts
```diff
@@ -37,6 +37,10 @@ export function hasCorrectBidGroup(code: string, group: string, a9?: boolean): b
 	return slotBidGroup === group;
 }
 
+export function hasCorrectBidsField(adUnitBids: PrebidBid[]) {
```
jbienkowski-fandom, line 40:
```suggestion
export function hasCorrectBidsField(adUnitBids: PrebidBid[] | undefined) {
```

domjanik (approved):

spec/ad-bidders/bidder-helper.spec.ts
```diff
@@ -116,5 +117,25 @@ describe('bidder-helper', () => {
 
 			expect(result).to.deep.be.false;
 		});
+
+		describe('hasCorrectBidsField', () => {
+			it('should return undefined if bids field is undefined', () => {
+				const result = hasCorrectBidsField(undefined as unknown as PrebidBid[]);
```
jbienkowski-fandom, line 123:
```suggestion
				const result = hasCorrectBidsField(undefined);
```
```suggestion
				const result = hasCorrectBidsField(undefined as unknown as PrebidBid[]);
```

jbienkowski-fandom (approved):

PR #2504, opened 2024-05-21
ADEN-13542 AE/IE partners logic moved behind document load but setup part done as fast as possible

preludek:


src/platforms/shared/utils/instant-config.setup.ts
```diff
@@ -32,6 +32,7 @@ export class InstantConfigSetup implements DiProcess {
 			instantConfigEndpoint,
 			instantConfigFallbackEndpoint,
 			instantConfigVariant: context.get('wiki.services_instantConfig_variant'),
+			...(window?.ads?.context?.warp && { lockDelay: 0 }),
```
jbienkowski-fandom, line 35:
```suggestion
			...(window?.ads?.context?.warp ? { lockDelay: 0 } : {}),
```

src/platforms/shared/setup/ad-engine-phases-setup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		adEnginePhases.firstAdCall = waitForEventPromise(
+			eventsRepository.AD_ENGINE_SLOT_LOADED,
+			context.get('options.phases.firstAdCallTimeout') ?? DEFAULT_AD_CALL_WAIT_TIMEOUT,
+		);
+	}
+
+	setupWarpContext(): void {
+		context.set('options.phases', this.instantConfig.get('icWarp', {}));
+		context.set('options.phases.enabled', !!this.instantConfig.get('icWarp'));
```
jbienkowski-fandom, line 55: `window?.ads?.context?.warp ?? false` let's use one var to control both setups, `icWarp` can provide timeout config

src/platforms/shared/utils/instant-config.setup.ts
preludek: We don't need ternary here, above should work as expected 🙈

src/platforms/shared/setup/ad-engine-phases-setup.ts
preludek: done

src/platforms/shared/setup/ad-engine-phases-setup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+		adEnginePhases.firstAdCall = waitForEventPromise(
+			eventsRepository.AD_ENGINE_SLOT_LOADED,
+			context.get('options.phases.firstAdCallTimeout') ?? DEFAULT_AD_CALL_WAIT_TIMEOUT,
+		);
+	}
+
+	setupWarpContext(): void {
+		context.set('options.phases', this.instantConfig.get('icWarp', {}));
+		context.set('options.phases.enabled', window?.ads?.context?.warp);
```
jbienkowski-fandom, line 55:
```suggestion
		context.set('options.phases.enabled', window?.ads?.context?.warp === true);
```
this way we will always deal with boolean (as JS handles logical expressions in a very weird way)

jbienkowski-fandom (dismissed):

@types/ads/index.d.ts
```diff
@@ -42,6 +42,7 @@ interface AdsContext {
 		main_entity_type: string;
 		main_vertical: string;
 	};
+[key: string]: unknown;
```
jbienkowski-fandom, line 45: where linter?

preludek: fixed

jbienkowski-fandom (approved):

PR #2507, opened 2024-05-23
ADEN-13787 move bt-force to bt-rec and remove duplicated code

preludek:


src/ad-tracking/trackers/bab-tracker.ts
```diff
@@ -3,19 +3,15 @@ import { DataWarehouseTracker } from '../data-warehouse';
...
-): void {
+export function trackBab(isBabDetected: boolean): void {
 	const dataWarehouseTracker = new DataWarehouseTracker();
 
 	dataWarehouseTracker.track({
 		category: 'ads-babdetector-detection',
 		action: 'impression',
 		label: isBabDetected ? 'Yes' : 'No',
-		value: detector,
+		value: 0,
```
mklakus-fandom, line 15: Just a question, is 0 enough here? 
I mean can we use one of the ` 'wad-runner' | 'wad-runner-bt' | 'bt'`?

mklakus-fandom (approved):

src/ad-tracking/trackers/bab-tracker.ts
preludek: We probably could but it also seems unnecessary 😅  before the A/B test it was 0 so I basically reverted the changes. After the test we don't need to differentiate the value. But yeah, we could even default it to 'bt'.

PR #2523, opened 2024-06-04
COTECH-1009 | `cnx-` prefix in adUnit for tagless vast request

latata:


src/platforms/shared/setup/player.setup.ts
```diff
@@ -98,10 +102,16 @@ export class PlayerSetup extends BaseServiceSetup {
 		}
 	}
 
-	private static getPrimaryVideoAdSlot() {
+	private static getPrimaryVideoAdSlot(isConnatix: boolean) {
+		console.log({ slotService: slotService.get(primaryVideoAdSlotName) });
```
andrzejkupczyk, line 106: Did you mean to use `utils.logger` instead?

src/platforms/shared/setup/player.setup.ts
```diff
@@ -116,11 +126,13 @@ export class PlayerSetup extends BaseServiceSetup {
 		}
 	}
 
-	private static initConnatixPlayer(showAds: boolean, vastResponse?: VastResponseData) {
+	private initConnatixPlayer(
+		showAds: boolean,
+		vastResponse: VastResponseData | undefined,
+		adSlot: AdSlot,
```
andrzejkupczyk, line 132: I don't understand why this change is being made. Can you please explain the reasoning behind it? Particularly: 
1) why is the `vastResponse` now required without clearly indicating that it's optional? If it is optional, why isn't it listed at the end of the parameter list just like it was before?
2) why is the `adSlot` required? this forces us to pass something even if `showAds=false` 🤔

src/platforms/shared/setup/player.setup.ts
```diff
@@ -116,11 +126,13 @@ export class PlayerSetup extends BaseServiceSetup {
 		}
 	}
 
-	private static initConnatixPlayer(showAds: boolean, vastResponse?: VastResponseData) {
+	private initConnatixPlayer(
```
andrzejkupczyk, line 129: Why was this method changed from static to non-static?

andrzejkupczyk (changes requested):

src/platforms/shared/setup/player.setup.ts
```diff
@@ -34,24 +36,26 @@ export class PlayerSetup extends BaseServiceSetup {
...
+		this.connatixFeaturedVideoEnabled =
+			this.instantConfig.get('icFeaturedVideoPlayer') === 'connatix';
 	}
 
 	async call() {
 		communicationService.on(eventsRepository.VIDEO_PLAYER_RENDERED, () => {
 			this.loadIasTrackerIfEnabled();
 		});
 		const showAds = !context.get('options.wad.blocking');
+		const adSlot = PlayerSetup.getPrimaryVideoAdSlot(this.connatixFeaturedVideoEnabled);
```
andrzejkupczyk, line 48: Is it necessary to call a private static method from the instance scope and pass its property as an argument? Wouldn't it be simpler to make the method non-static and access the `connatixFeaturedVideoEnabled` property directly from the instance?

src/platforms/shared/setup/player.setup.ts
latata: that was just a console log for debugging purposes while coding. I removed it

src/platforms/shared/setup/player.setup.ts
latata: while developing I wanted to use `this` but then I changed it and forgot to change it back to static. I will update it

src/platforms/shared/setup/player.setup.ts
latata: adslot is being created regardless the `showAds` flag. While it doesn't make much sense I'd leave it as is, as it was like that before my changes. 

I will change `vastResponse` to be the last argument. 👌

andrzejkupczyk (approved):
LGTM. Thanks for the updates!

